PR #1, opened 2021-02-10
benchmarks

gwbischof:


danielballan:
Relative imports are nice, especially in packages are likely to be reorganized or renamed in the future. I would be surprised if asv didn't support them. I can't think of a good reason why it couldn't, and know some projects that use asv (like dask) also use relative imports. I suspect bc22715 isn't needed and/or that something else is going wrong.

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,42 @@
...
+import uvicorn
+import time
+import subprocess
+
+from catalog_server.client import ClientCatalog
+from catalog_server.server import app
+from catalog_server.example_catalogs import catalog
+
+HOST = '0.0.0.0'
+PORT = '9040'
```
danielballan, line 13: It looks off to store this as a string. Can it be an int?

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,42 @@
...
+class TimeSuite:
+    """
+    An example benchmark that times the performance of various kinds
+    of iterating over dictionaries in Python.
+    """
+    def setup(self):
+        self.server_process = subprocess.Popen((f"uvicorn catalog_server.server:app"
+                                                f" --host {HOST} --port {PORT}").split())
+        time.sleep(5)
+        self.catalog = ClientCatalog.from_uri('http://' + HOST + ':' + PORT)
```
danielballan, line 24: This is a good use case for f-string (and it will naturally handle `PORT` being an int).

danielballan (commented):
One fiddly comment. Then I think this good to go.

PR #2, opened 2021-03-08
Fastapi simple security

gwbischof:


danielballan:
I don't think this clears the bar for "worth adding a dependency vs just maintaining it ourselves". It's [10 commits from one day in August](https://github.com/mrtolkien/fastapi_simple_security/commits/master) by a single contributor, not touched since. It's not really much more than [this blog post on which the current implementation in catalog-server in based](https://medium.com/data-rebels/fastapi-authentication-revisited-enabling-api-key-authentication-122dc5975680) plus SQLite for state. I would feel better about owning this part of the stack ourselves.

PR #4, opened 2021-03-18
add jwt auth

gwbischof:


gwbischof:
image

gwbischof:
image

In the first example get_public_jwt is used and we can see the "medium" catalog entry.
In the second example I hardcode token = "" and we the authentication error

tiled/server/jwt_auth.py
```diff
@@ -0,0 +1,123 @@
...
+from fastapi.security import OAuth2PasswordBearer, OAuth2PasswordRequestForm
+from jose import JWTError, jwt
+
+from passlib.context import CryptContext
+
+from pydantic import BaseModel
+
+# to get a string like this run:
+# openssl rand -hex 32
+SECRET_KEY = "09d25e094faa6ca2556c818166b7a9563b93f7099f6f0f4caa6cf63b88e8d3e7"
```
danielballan, line 14: Marking this to be sure we revisit it before merging. This should come from Settings and ultimately from the environment. It should also be a `;`-separated list of keys to support key rotation. The first entry will be used for encoding but any of them will be valid for deciding.

tiled/client/catalog.py
```diff
@@ -21,6 +23,26 @@
...
+    username = input("Username: ")
+    password = getpass.getpass()
+    form_data = {'grant_type': 'password',
+                 'username': username,
+                 'password': password}
+    response = httpx.post(uri + '/token', data=form_data)
+    return 'Bearer ' + json.loads(response.text)['access_token']
+
+
+def get_public_token(uri):
```
danielballan, line 36: Interesting idea. Placing something in the `Authorization` header that is actually not a secret but in fact a hard-coded public constant feels off to me. What would it look like to make a special case in the server where, if Authorization is not set, the user is treated as `SpecialUsers.public`?

Another consideration is that any string we use for username here could potentially collide with an actual username in whatever user management system we integrate with. (Perhaps some guy Paul Ublic has public@bnl.gov — who knows!) It’s unlikely but it is cleaner to make the public “user” a special sentinel that is “out of band” from the username namespace, as I have done in the sketch in `main`.

tiled/client/catalog.py
```diff
@@ -21,6 +23,26 @@
 from .utils import handle_error
 
 
+def get_token(uri):
+    username = input("Username: ")
```
danielballan, line 27: I like making this a stand-alone function. 👍

Prompts are rare in Python APIs; they are more of a CLI thing. They do not play well with batch processing or notebook workflows. I suggest we require the username as a normal function argument (it’s not a secret) and look for the password in an env var first, `TILED_PASSWORD`, and then fall back to prompting for the password only if that is unset.

tiled/client/catalog.py
```diff
@@ -21,6 +23,26 @@
...
+    password = getpass.getpass()
+    form_data = {'grant_type': 'password',
+                 'username': username,
+                 'password': password}
+    response = httpx.post(uri + '/token', data=form_data)
+    return 'Bearer ' + json.loads(response.text)['access_token']
```
danielballan, line 33: Minor thing: we might want to the token alone. Can Bearer be tacked on later, when it is placed in the header?

tiled/server/jwt_auth.py
```diff
@@ -0,0 +1,123 @@
...
+ALGORITHM = "HS256"
+ACCESS_TOKEN_EXPIRE_MINUTES = 30
+
+
+fake_users_db = {
+    "alice": "$2b$12$EixZaYVK1fsbw1ZfbX3OXePaWxn96p36WQoeG6Lruj3vjPGga31lW",
```
danielballan, line 20: We might as well factor this out into a constant and include a comment with its plaintext value.

tiled/client/catalog.py
danielballan: Oh, and `response.json()` would be more idiomatic here.

tiled/client/catalog.py
danielballan: I’ve been giving this more thought and looking around a bit, and I’d like to revise my recommendation. This function should accept a username and password directly. It’s annoying as a user to have to know a specific environment variable name or config location if I just want to get it working using

```py
f(username=os.environ["WHATEVER_I_WANT"], password= os.environ["AGAIN_WHATEVER_I_WANT"])`
```

That explicit spelling should be _option_ and the core implementation. It should come with a documented warning not to hard-code the variables...but we needn't make it _impossible_ to do so.

In later PRs, once we grow a config system, we can look to examples like https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html which provide a whole range of ways to do this which are convenient in different situations. We can add support for things like

```py
if password is None:
    password = os.getenv("TILED_PASSWORD")
```

and

```py
ClientCatalog.from_profile("CSX")  # finds a section named "CSX" in a config file containing URI, username, password, and options
```

tiled/server/jwt_auth.py
gwbischof: Can we make this into an issue and do this later?

tiled/client/catalog.py
gwbischof: If we pass username, and password as function arguments. Then people will leave their password laying around in python code in plain text.

gwbischof: also it seems weird to stick your plaintext password in an environment variable

danielballan: > If we pass username, and password as function arguments. Then people will leave their password laying around in python code in plain text.

I think this is a documentation issue. Higher layers, to be added _soon_ but now _now_ can provide indirect ways of injecting the password, include interactive prompting. But I really don't like the _lowest_ layer building in a prompt and assuming a TTY.

> also it seems weird to stick your plaintext password in an environment variable. Because it is not an API key, and it's a users password, this may be a password that they reuse in other places

That's a fair point, but many services including PyPI work this way. Using a token is better practice, but when we have username/password it's pretty standard to accept it via env var. Perhaps the best, most mainstream balance of convenience and security here is to support password-via-env-var but guide people toward OAuth2- or SAML-based Authenticators in the long run, where they'll be generating a code instead of handing us a password directly. Basically how Globus works from the CLI.

I pinged @dylanmcreynolds on Slack for a second opinion.

danielballan: @dylanmcreynolds is with @gwbischof that password-in-env-var is too dodgy. (As I refresh my memory..PyPI _does_ do this but [it is a known problem](https://glyph.twistedmatrix.com/2017/10/careful-with-that-pypi.html).)

I propose this, then, which is just a cosmetic refinement on what @gwbischof proposed:

```python
>>> from tiled.client.catalog import generate_token, Catalog
>>> URI = "http://..."
>>> token = generate_token(URI)
username: dallan
password: [hidden input]
>>> catalog = Catalog.from_URI(URL, token)
```

with an option to do the "text adventure" prompts in the shell:

```
$ tiled token generate http://
username:
passsword:
Token:
XXX
This token will expire in X minutes at <time>.
$ python
>>> client = Catalog.from_uri("...", token="...")
```

danielballan: (Fine with me if the CLI is a separate PR, btw. Would need a special header in the token response so that the client can know the expiration time without installing jwt and encryption libraries.)

danielballan:
Looks right but we should handle the "wrong password" case better.

```py
In [1]: from tiled.client.catalog import Catalog, generate_token

In [2]: token = generate_token("http://localhost:8000")
Username: alice
Password: 

In [3]: generate_token("http://localhost:8000")
Username: alice
Password: 
KeyError                                  Traceback (most recent call last)
<ipython-input-3-1965e68419aa> in <module>
----> 1 generate_token("http://localhost:8000")

~/Repos/bnl/tiled/tiled/client/catalog.py in generate_token(uri)
     33     form_data = {"grant_type": "password", "username": username, "password": password}
     34     response = httpx.post(uri + "/token", data=form_data)
---> 35     return response.json()["access_token"]
     36 
     37 

KeyError: 'access_token'

In [4]: catalog = Catalog.from_uri("http://localhost:8000")

In [5]: list(catalog)
Out[5]: ['medium']

In [6]: catalog = Catalog.from_uri("http://localhost:8000", token=token)

In [7]: list(catalog)
Out[7]: ['medium', 'large']
```

danielballan:
This function https://github.com/bluesky/tiled/blob/811b3f97701b896b9a48f0f30043d6e7a43738e0/tiled/client/utils.py#L86 can check for general errors.

```py
try:
    handle_error(response)
except err:
    if err.response.status_code == 401:  # Unauthorized (wrong username or password)
        # raise friendlier error
    raise  # Some other failure
```

PR #20, opened 2021-07-07
Fix unsigned int dtype character

SimonHeybrock:


danielballan:
Thanks @SimonHeybrock!

Confirmed with reference to https://numpy.org/devdocs/reference/arrays.interface.html#object.__array_interface__ that the current implementation is just plain wrong---likely a typo.

PR #22, opened 2021-07-08
Support adding custom formats via server config. Add how-to guide with XDI as example.

danielballan:
The export function in the documentation is currently just a sketch. Once that is worked out more realistically, then I think this is ready to go.


danielballan:
For convenience: https://github.com/XraySpectroscopy/XAS-Data-Interchange/blob/master/specification/spec.md

That's also linked in the docs already.

docs/source/how-to/custom-formats.md
```diff
@@ -0,0 +1,173 @@
+# Add Support for a Custom Format
+
+For a given "structure family" (e.g. `array`, `dataframe`) the Tiled server
+can provide the data---in whole or in part---in a variety of formats. 
+See {doc}`../tutorials/export` for a list of the formats supported out of the
+box.
+
+To teach Tiled to serve another format, first decide which structure family or
+families are appropriate for this format? For example, it make sense to export
```
dylanmcreynolds, line 9: maybe remove the question mark?

danielballan: Ah yes. Left over from a grammatical refactor.

docs/source/how-to/custom-formats.md
```diff
@@ -0,0 +1,176 @@
...
+    1. Use the standard [HTTP `Accept` Header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept).
+
+    ```
+    $ curl -H 'Accept: text/csv' http://localhost:8000/dataframe/full/example
+    ```
+
+    2. Place the media type in a `format` query parameter.
+
+    ```
+    $ curl -H http://localhost:8000/dataframe/full/example?format=text/csv
```
kleinhenz, line 67: Should be `curl -L ...`

danielballan: Good catch. Was a copy/paste error. Fixed.

danielballan:
@kleinhenz I reverted 918a768 and pushed an alternative implementation that achieves the same end in 75dbb99. See commit message in 75dbb99 for my reasoning.

I cobbled to together a very basic test in 613969a to verify it still works.

PR #25, opened 2021-07-09
pin httpx version

dylanmcreynolds:
httpx changed their api in 0.18...url.netloc used to provide a string, now it provides bytes. Tiled has calls that treat this as the newer bytes, so this PR pins httpx to 0.18.2 or greater.

https://github.com/bluesky/tiled/blob/fb7acf75373d86386206c6b1d922ff0ea1ceac30/tiled/client/utils.py#L101

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,37 @@
+# .git
+# .gitignore
+# node_modules
+# npm-debug.log
+# Dockerfile*
+# docker-compose*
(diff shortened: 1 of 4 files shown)


danielballan:
Nice! I have also encountered this but had not yet found time to track down the specific change in httpx that caused it. Nice!

attn @jklynch

PR #29, opened 2021-07-13
Added case to get shape information for macrostructure

jmaruland:
Some files don't contain shape information in the metadata. If this field is empty, the size of the data/array is read to build the list that contains this information.

### DIFF
--- a/tiled/readers/tiff.py
+++ b/tiled/readers/tiff.py
@@ -52,7 +52,7 @@ class TiffReader:
         if self._file.is_shaped:
             shape = tuple(self._file.shaped_metadata[0]["shape"])
         else:
-            shape=[len(self._file.asarray()), len(self._file.asarray()[0])]
+            shape = self._file.asarray().shape
         return ArrayMacroStructure(
             shape=shape,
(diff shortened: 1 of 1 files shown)


tiled/readers/tiff.py
```diff
@@ -49,7 +49,10 @@ def microstructure(self):
         return MachineDataType.from_numpy_dtype(self._file.series[0].dtype)
 
     def macrostructure(self):
-        shape = tuple(self._file.shaped_metadata[0]["shape"])
+        if self._file.is_shaped:
+            shape = tuple(self._file.shaped_metadata[0]["shape"])
+        else:
+            shape=[len(self._file.asarray()), len(self._file.asarray()[0])]
```
danielballan, line 55: Two notes on this line:

1. We need `shape` to be a `tuple`.
2. Does `self._file.asarray().shape` work? What is `type(self._file.asarray())`?

jmaruland: I confirmed that this function returns a NumPy array and proceed to add the proper command using `self._file.asarray().shape`

danielballan (approved):

Issue #31, opened 2021-07-13
Make XAS example more precise


danielballan:
We rushed in #22 with a half-baked (but sufficient) example because there is broad interest in registering custom formats. Now let's revisit that example.

The `metadata` was transliterated from the XAS ASCII format. It is a dict with dots in the keys:

https://github.com/bluesky/tiled/blob/3f0087adf03f850618e6812a615ac617c394acbc/tiled/examples/xas.py#L10-L33

Instead, it should be a nested dict like:

```py
{
    "Column": {
        "1": "energy eV",
        "2": "i0",
        ...
    },
    ...
}
```

The data types of the values should also be reviewed and potentially revised. In Tiled, `metadata` may contain `str`, `int`, `float`, and `datetime`.

kleinhenz:
closed by #34

PR #34, opened 2021-07-28
improve xdi example

### LINKED ISSUE #28: Support registering serializers for a spec
In #22 we add support for registering serializers (a.k.a. exporters) for a _structure_ like `"dataframe"`. We should also support registering serializers for a _spec_ like `"xdi"`.

This is actually a pretty simple change to execute, I think:

- [ ] Update the service configuration schema to allow `additionalProperties`.
- [ ] Update the server's response-builders to check the `serialization_registry` for each `spec` in `specs` in turn before falling back to the `structure`.

kleinhenz:
This improves the XDI example by showing how to ingest XDI data as well as how to serve it. Additionally this enables registering serializers for specs and makes use of this in the XDI example. Currently these serializers are only enabled for the `/dataframe/full` endpoint. If you are happy with this approach we discuss where else this should work. Closes #28, #31.


kleinhenz:
Note sure what the documentation failure is coming from. Work locally.

docs/source/how-to/custom-formats.md
```diff
@@ -13,18 +13,156 @@ assumptions about the array.)
...
+
+def read_xdi(path):
+    with open(path, "r") as f:
+        metadata = {}
+        fields = collections.defaultdict(dict)
+
+        line = f.readline()
+        m = re.match("# XDI/(\S*)\s*(\S*)?", line)
+        if not m:
+            raise AssertionError(f"not an XDI file, no XDI versioning information in first line\n{line}")
```
danielballan, line 87: This should be `ValueError`. CPython says assertions are for "This state should be impossible; the program has a bug," not for input validation.

docs/source/how-to/custom-formats.md
```diff
@@ -13,18 +13,156 @@ assumptions about the array.)
...
+
+        field_end_re = re.compile("#\s*/{3,}")
+        header_end_re = re.compile("#\s*-{3,}")
+
+        has_comments = False
+
+        # read header
+        for line in f:
+            if line[0] != "#":
+                raise AssertionError(f"reached invalid line in header\n{line}")
```
danielballan, line 100: Ditto

docs/source/how-to/custom-formats.md
```diff
@@ -13,18 +13,156 @@ assumptions about the array.)
...
+                break
+            elif re.match(header_end_re, line):
+                break
+
+            try:
+                key, val = line[1:].strip().split(":", 1)
+                val = val.strip()
+                namespace, tag = key.split(".")
+                #TODO coerce to lower case?
+            except:
```
danielballan, line 112: Avoid bare `except:` --- this would catch `KeyboardInterrupt` for example as well. I suggest:

```py
except ValueError as err:
    raise ValueError(f"error processing line\n{line}") from err
```

That will show the specific error at the bottom but also include the original error above as its direct cause.

docs/source/how-to/custom-formats.md
```diff
@@ -142,13 +285,14 @@ First consider the section
 
 ```yaml
 media_types:
-  dataframe:
-    application/x-xdi: exporter:serialize_xdi
+  xdi:
```
danielballan, line 288: We had talked about separate namespaces here, as in

```yaml
media_types:
    by_structure_family:
        ...
    by_spec:
        xdi:
```

Do we still want to go that way?

tiled/server/core.py
```diff
@@ -343,7 +348,7 @@ def construct_array_response(
 
 
 def construct_dataframe_response(
-    serialization_registry, df, metadata, request_headers, format=None
+    serialization_registry, df, metadata, request_headers, format=None, specs=[]
```
danielballan, line 351: Avoid mutable defaults. (See for example: https://docs.python-guide.org/writing/gotchas/)

tiled/server/core.py
```diff
@@ -367,16 +372,21 @@ def construct_dataframe_response(
...
     # that we support.
     for media_type in media_types:
         if media_type == "*/*":
             media_type = APACHE_ARROW_FILE_MIME_TYPE
-        if media_type in serialization_registry.media_types("dataframe"):
-            content = serialization_registry("dataframe", media_type, df, metadata)
-            return PatchedResponse(
-                content=content, media_type=media_type, headers={"ETag": etag}
-            )
+        for spec in specs:
```
danielballan, line 384: Can we do `spec in specs + ["dataframe"]` rather than mutating `specs` itself above? Mutating the user input could have confusing side-effects.

danielballan (commented):
I like the approach. I left some implementation suggestions.

docs/source/how-to/custom-formats.md
danielballan: Follow-up thought: this may require a fairly significant refactor. I'd be fine with using a single namespace for now and separating them in a future PR.

kleinhenz: makes sense to me. lets leave it for now

kleinhenz:
I think I have addressed your comments.

danielballan (approved):

PR #35, opened 2021-07-29
docker

kleinhenz:
This adds a basic Dockerfile for the tiled server component. I've checked that this works with the simple examples.


untzag:
If y'all add something like this we can install directly from GHCR :smile: 

https://github.com/bluesky/bluesky-webclient/blob/main/.github/workflows/publish-image.yml

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+FROM base as builder
+
+WORKDIR /install
+
+COPY . .
+
+RUN pip install --upgrade pip
+RUN pip install -r requirements-server.txt
```
danielballan, line 14: I think you can drop these lines. The pip selector below, `[…]`, is equivalent. It reads these files and installs the dependencies they list.

danielballan (commented):
Nice!

I’d like to take this for a spin before we merge, but it looks good. I left one minor suggestion.

danielballan:
Stepping back, I wonder if we want to make a container with a single-process app. If we are at the point of deploying a container, we likely also want gunicorn to horizontally scale.

kleinhenz:
This runs tiled through gunicorn now.

danielballan:
Great!

danielballan:
In a follow-up PR I think we need a how-to guide in the docs on deploying a horizontally scaled app.

docs/source/how-to/docker.md
```diff
@@ -0,0 +1,44 @@
...
+    args:
+      directory: "example_files"
+uvicorn:
+  host: 0.0.0.0
+  port: 8000
+authentication:
+  allow_anonymous_access: true
+```
+and serve it using the docker container
+```
```
dylanmcreynolds, line 36: Do we want to use the more concise `--volume` flag rather than `--mount`?


docker run --rm -p 8000:8000 \
  --volume "/deploy/example_files: $(pwd)/config.yml" \
  --volume "$(pwd)/config.yml:/deploy/config.yml" \
  --env TILED_CONFIG=/deploy/config.yml tiled

docs/source/how-to/docker.md
```diff
@@ -0,0 +1,44 @@
...
+docker run --rm -p 8000:8000 \
+  --mount type=bind,source="$(pwd)/example_files",target=/deploy/example_files \
+  --mount type=bind,source="$(pwd)/config.yml",target=/deploy/config.yml \
+  --env TILED_CONFIG=/deploy/config.yml tiled
+```
```
dylanmcreynolds, line 41: This example seems that it would launch tiled, and avoid gunicorn.

dylanmcreynolds (dismissed):
I really like this. I had always imagined using tiangolo's gunicorn image, but this is more maintainable in the long run (tiangalo's image has already fallen a little behind python versions.)

Can we update the documentation to reflect the change from tiled to gunicorn?

docs/source/how-to/docker.md
kleinhenz: The container is named tiled so this will run the container entrypoint which runs gunicorn. Maybe the container should be named something else to make it more clear?

dylanmcreynolds: Oh dangit...I misread that (thinking it was exec'ing the tiled not defining the container name). Tiled is the right name.

docs/source/how-to/docker.md
kleinhenz: The docker documentation encourages the more verbose form.
https://docs.docker.com/storage/bind-mounts/
I can change it though if you prefer.

dylanmcreynolds: so it does! their logic is funny "research has shown that it is easier to use"... anyway, no, don't change it.

kleinhenz:
@dylanmcreynolds What documentation do you want updated?

kleinhenz:
@danielballan I've added some docs showing how to run a simple example using the container

dylanmcreynolds:
> @dylanmcreynolds What documentation do you want updated?

Sorry, my initial misunderstanding was that the docker command as documented hadn't been updated to reflect the gunicorn change. I rescind my request. Thanks!

danielballan (approved):

danielballan:
Thanks, @kleinhenz!

PR #37, opened 2021-07-30
Remove generated files from version control.

danielballan:


danielballan:
Two of the `.md` files are generated at docs build time via https://github.com/bluesky/tiled/blob/7d967ea7932bdb0318b9169a261cc2ae0f89a643/docs/source/conf.py#L272-L291

These should never have been committed.

PR #46, opened 2021-08-02
Fix regressions

danielballan:
I discovered that #34 broke Tiled's `/array/...` routes because of the issue fixed in d08d6b4.

This was not caught by CI because _the workflows "passed" even when `pytest` failed_ due to the issue fixed in b379f2a.

The test in `tiled/_tests/test_custom_format.py` was also failing, and it needed to be rethought due to changes in the API of the module it was designed to test. I made some small changes to `tiled.examples.xdi` to facilitate that testing.


danielballan:
Confirmed that the tests _actually_ passed this time at https://github.com/bluesky/tiled/runs/3224440890#step:6:71

kleinhenz (approved):

Issue #48, opened 2021-08-04
Enhancements to the directory tree


danielballan:
- [x] By default, strip file extension from node name. That is, `a.tif` becomes `a`. Rationale: the file extension is not part of the "name". (Windows hides it by default.) Including it makes it confusing to do things like `http://.../thing.xlsx?format=csv`. This _does_ create the possibility of name collisions, but if you have `example.tiff` and `example.json` next to each other, they are likely related --- part of the same "data" --- and need a special Adapter to deal with. #49 
- [ ] Add a configurable function that, given a full path to a directory, returns a metadata dict. (The default will always return an empty dict.)
- [x] Add a configurable function that, given a full path to a directory, returns a mimetype or `None`. If it returns `None`, that means, "Crawl into this subdirectory." If it returns a mimetype, that means, "I recognize this as a {TIFF sequence, Zarr directory, TileDB directory} with internal structure than an Adapter shoudl manage. Do not crawl any further. We might want to make this a _list_ of functions to make it easy to extend default behavior. https://github.com/bluesky/tiled/pull/62

Developed in a phone conversation with @tacaswell

J-avery32:
> Add a configurable function that, given a full path to a directory, returns a metadata dict.

Currently is there no way to add metadata for a node without re-writing the adapter to add this in?

danielballan:
Correct. I think what you want here is similar to what we added in #259 — a way to _extend_ the existing behavior without having to rewrite a bunch of stuff.

danielballan:
I think the feature you need is `subdirectory_handler`. Does each directory map to a single (3-dimensional) array, comprised off all the TIFFs stacked up, with a dictionary of metadata attached? If so, it should be fit your use case as is. See

https://github.com/bluesky/tiled/blob/34ede8a48bf3c9c901acc6ae7db3f12fbe2d0f2b/tiled/adapters/tiff.py#L178

and the (rough) documentation

https://blueskyproject.io/tiled/explanations/scaling-down.html

If the structure you need is more complex than that, we should talk about how to extend `subdirectory_handler` to enable it.

danielballan:
@J-avery32 Does `subdirectory_handler` address your use case?

J-avery32:
[...]

PR #49, opened 2021-08-04
Strip suffixes from filenames.

danielballan:
This changes the default behavior of `tiled serve directory ...` to name its nodes after just the file _name_ excluding its suffix(es). That is, if the file is `a.tif` the URL will be `http://.../a`, not `http://.../a.tif`.

At first I thought we couldn't/shouldn't do this because it could potentially result in name collisions, but after thinking it over for awhile and chatting with @tacaswell I've come a fairly strong opinion that it is the right thing to do.

Philosophically the file extension is not part of the "name". Windows hides it by default in their UI, for example. It might be thought of as metadata tacked alongside the name using an in-band mechanism.

When the file extension is included in the node name, it makes it a little confusing to request things like `http://.../thing.xlsx?format=csv` and I think makes it unclear for people who are just encountering Tiled what exactly Tiled is _doing_ here.

If the event of name collisions, a warning will be issued, and the last one we see will "win". If you have `example.tif` and `example.json` next to each other, they are likely related --- part of the same "data" --- and need a special `Adapter` to deal with. Or, you can always override our default and set

```yaml
key_from_filename: tiled.trees.files:identity  # Use full filename as key.
```

to revert to the old behavior, or put some custom function there with some more complex scheme if you have a special use case.

Addresses one part of https://github.com/bluesky/tiled/issues/48


danielballan:
Just for the record...

Another, maybe stronger reason to do this, is that the premise of tiled is I can change the format from `.csv` to `.tsv` (to give a simplistic example) that the client should not feel that change. Exposing the file extension leaks storage details to the client and thus undercuts what Tiled is trying to achieve.

tacaswell (approved):

kleinhenz (approved):

danielballan:
To summarize Slack discussion on this PR:

* We want to expose flag `--keep-ext` / `--strip-ext` to make it easy to discover and invoke this option.
* We are split on what the best default is.

danielballan:
Just had a thought: if there is a key collision we should drop _all_ the colliding entries rather than serving one arbitrarily.

That’s much better than the current implementation in the PR serves whichever one we saw most recently / was updated most recently. If the item is _missing_ the user will investigate rather than assuming all is well and then running into confusing behavior downstream when an array suddenly transmutes into a dataframe or something.

danielballan:
Implementing this will take a bit but I think it's worth doing before we merge.

tiled/trees/files.py
```diff
@@ -16,10 +16,61 @@
...
+
+    >>> strip_suffixes("a.tif")
+    "a"
+
+    >>> strip_suffixes("thing.tar.gz")
+    "thing"
+    """
+    path = Path(filename)
+    # You would think there would be a method for this, but there is not.
+    filename_without_suffixes = str(path)[: -sum([len(s) for s in path.suffixes])]
```
tacaswell, line 36:
```suggestion
    if len(path.suffixes):
        return str(path)[: -sum([len(s) for s in path.suffixes])]
    else:
        return filename
```

which prevents

```
In [5]: path = Path('abc')

In [6]: str(path)[: -sum([len(s) for s in path.suffixes])]
Out[6]: ''


```

tiled/trees/files.py
```diff
@@ -378,21 +486,20 @@ def _process_changes(
...
-                if greedy:
-                    index[parent_parts][key] = reader_factory()
-                else:
-                    index[parent_parts][key] = reader_factory
             except NoReaderAvailable:
                 # Ignore this file in the future.
                 # We already know that we do not know how to find a Reader
                 # for this filename.
                 ignore.add(path)
+            if greedy:
```
tacaswell, line 498: This needs to be protected or you will get a name error?

tiled/trees/files.py
```diff
@@ -345,30 +406,77 @@ def _process_changes(
...
+                            readers_by_mimetype,
+                            mimetypes_by_file_ext,
+                            path,
+                        )
+                    except NoReaderAvailable:
+                        # Ignore this file in the future.
+                        # We already know that we do not know how to find a Reader
+                        # for this filename.
+                        ignore.add(path)
+                    if greedy:
```
tacaswell, line 446: This also needs to be protected.

PR #56, opened 2021-08-13
fix tree issue and add test stub

dylanmcreynolds:
With a pretty simple configuration for serving a directory tree:

```  
config = {
        "trees": [
            {
                "tree": "tiled.trees.files:Tree.from_directory",
                "path": "/data",
            },
        ],

    }
```

I get an error:
```
line 99, in construct_serve_tree_kwargs
    tree = obj(**item["args"])
  File "/opt/venv/lib/python3.8/site-packages/tiled/trees/files.py", line 199, in from_directory
    return cls(
  File "/opt/venv/lib/python3.8/site-packages/tiled/trees/files.py", line 234, in __init__
    self._manual_trigger = manual_trigger
AttributeError: 'Tree' object has no attribute '_manual_trigger'
```

This appears to be because a recent commit added a member to files:Tree, `_manual_trigger`, but did not add that to `_slots_`

This was a relatively simple fix.

I added a test that fails before this fail and passes after it. This adds a _tests/data directory...it would be nice to use the examples_files, but we can't depend on that being on a machine, right?

### Edit
This was an edit from the origin description...removing a complaint about test failing.


danielballan:
Thanks. Tests on this were overdue.

danielballan:
> it would be nice to use the examples_files

We can make a fixture to generate that (in a temporary location) and clean it up afterward.

PR #57, opened 2021-08-16
Fetch variables in batch to avoid URI size limit.

danielballan:
The Tiled client for xarray.Datasets can request each column individually, as a separate HTTP request. For wide tables with many scalar columns this is very slow because it involves many small requests. Therefore, the client has a _wide table optimization_ that requests all the scalar columns in a single request, encoded as an Apache Arrow file.

The list of the scalar columns (or "variables", in xarray jargon) can be long. In the wild, I have seen this list get so long that it trips nginx's limit on URI length. This PR batches the requests, aiming to keep the URI below ~2000 characters. See code comments for where that number came from.

I tested it interactively by reducing `INIT_BUDGET` from 2000 to 20 and 200 and watching server longs. Update: And added proper unit tests.


tiled/client/xarray.py
```diff
@@ -369,16 +370,42 @@ def register(self, name, data_array):
...
+            # we can bump into the URI size limit commonly imposed by
+            # HTTP stacks (e.g. nginx). The HTTP spec does not define a limit,
+            # but a common setting is 4K or 8K (for all the headers together).
+            # As another reference point, Internet Explorer imposes a
+            # 2048-character limit on URLs.
+            variables = []
+            dataframes = []
+            budget = INIT_BUDGET = 2000  # number of characters
+            budget -= len(self.link)
+            CONSTANT = len("&variable=")
```
dylanmcreynolds, line 383: this could be moved outside the method to not get redefined each call for a micro-small performance improvement :)

dylanmcreynolds (changes requested):
Could this be done more simply by always fetching one variable per request? More messages would be bad for some cases and good for others.

At the very least, I think you want to add pandas to `requirements-xarray.txt`

Finally, I don't now the tiled test suite, but I don't see where this is tested.  This slightly complicated mechanism for getting xarrays with several branches of code is beyond my comfort level for not having a unit test.

tiled/client/xarray.py
dylanmcreynolds: Actually, what if you made this injectable somehow so that a unit test could add a smaller value, and a future PR could make it configurable, if desired?

tiled/client/xarray.py
```diff
@@ -2,6 +2,7 @@
 import builtins
 import itertools
 
+import pandas
```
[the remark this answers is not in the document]
danielballan: [To recap a Zoom converstaion] Xarray depends on it anyway, but you are right that we should explicitly list it now since we import it directly. That way, we are future-proofed against xarray possibly dropping it as a required dependency.

tiled/client/xarray.py
danielballan: Done in 3551d0b

dylanmcreynolds (approved):

PR #59, opened 2021-08-18
docker improvements

### LINKED ISSUE #55: Ensure a consistent signing key is used inside the container
Something I missed when reviewing #35:

If secret key(s) are not explicitly set _somewhere_ then each process managed by gunicorn will generate its own independently

https://github.com/bluesky/tiled/blob/97d158ef1de745ebba7cb510fb8e2d5999d13f87/tiled/server/settings.py#L35

and therefore JWTs signed by one worker will be rejected by other workers.

The fix is to ensure that the secret keys are set either in the user-provided config file or via the environment.

Off the top of my head, this probably means we need a short script to serve as the Docker entrypoint. It needs to invoke Tiled’s config parser to check whether secrets have been specified there or via the environment. If neither, it must either fail and insist that the user provide a secret with a clear error message—or automatically generate one and place it somewhere outside the container where it can be found when the container restarts. The latter sounds fraught so I would suggest the former, at least as a starting point.

https://blueskyproject.io/tiled/reference/authentication.html#set-and-rotate-the-signing-key

kleinhenz:
This makes a few improvements to the docker configuration. 

* adds a script to parse the config/environment to make sure that either the server is running as public or else the secret keys are explicitly set to make sure that all gunicorn workers have the same keys (closes #55) 
* updates the github action to push an image when a new tag version tag is pushed 
* moves the docker related files into their own subdirectory
* pins the dependencies to specific versions. This make the container more reproducible and also makes it so that the pip install step doesn't have to run every time the image to built as long as the requirements file has not changed.


danielballan (approved):
I like all the changes.

The configuration-checking logic wasn't quite correct. A public deployment still needs secret keys if it has an Authenticator configured because users _may_ authenticate, and it can affect which entries they are allowed to see.

I addressed this myself because it seemed faster to pitch in than to explain. :- ) I also made the error messages more prescriptive. Can you can give that a close look @kleinhenz and let me know if you see anything off?

danielballan (commented):
Attempting to actually run this I ran into an issue that I think forces us to restructure a little.

Previously, when `Dockerfile` was at the repo root, the tiled source tree could be copied in like

```
COPY . .
```

But now we would need something like `COPY .. .`, which is not allowed, for good reason.

```
Step 11/19 : COPY ../ .
COPY failed: forbidden path outside the build context: ../ ()
```

I think our only option is to move `Dockerfile` back to the repository root. We can continue to keep the support files organized under `docker/`. I have done so in ccfc82b6a4fd9dff29ed156139ea8dea755c20d5. Feel free to blow away that commit if you see a better way.

dylanmcreynolds:
> Previously, when `Dockerfile` was at the repo root, the tiled source tree could be copied in like

I have at least one project that stores Dockerfile in a subdirectory.  Generally, the general strategy to run from the directory that the Dockerfile is in but pass it the higher directory, so that the Dockerfile sees the top level even if it's located in a child. Something like:

`docker build -t <tag> -f Dockerfile ../../`

Could be accomplished in the build file by changing the build file from:

```
 with:
          context: .
          file: docker/Dockerfile
```

to 

```
 with:
          context: ./docker
          file: Dockerfile
```

.github/workflows/testing.yml
```diff
@@ -40,3 +40,9 @@ jobs:
         set -vxeuo pipefail
         coverage run -m pytest -v
         coverage report
+
+    - name: Test docker entrypoint configuration checker
+      shell: bash -l {0}
+      run: |
+        set -vxeuo pipefail
+        ./docker/test_check_config.sh
```
dylanmcreynolds, line 48: Holy mole this is cool.

kleinhenz:
You can also run from the project root directory and pass it the location of the Dockerfile with the `-f` flag: `docker build -t tiled-dev -f docker/Dockerfile .` This is what the current version of the workflow does. This seems like the easiest thing to me so I would revert ccfc82b6a4fd9dff29ed156139ea8dea755c20d5.

dylanmcreynolds (commented):
I love these changes.

I'd like for us to think a little bit about pinning the dependencies for docker but not for other builds. I definitely see the point of pinning them in docker for consistency. It does leave the possibility that someone commits something that depends on a newer version of a library and breaks the docker build...this situation will be invisible to our unit test suite and the developer, and will be kind of hard to track down.

kleinhenz:
Yeah its a tradeoff against the opposite problem where a new version breaks your stuff unexpectedly (which I think we have run into at least once already). My thought was that the docker container will be used for deployments so it should take the more conservative position. Updating the requirements file then becomes part of the release process.

danielballan:
OK, @kleinhenz feel free to revert  ccfc82b. Do we need to do anything special in the GH Action workflow to pass that configuration into the `build-push-action`?

I feel strongly that `requirements-*.txt` is the wrong place to pin. Libraries shouldn't pin _down_ ("version less than ...") just to be safe. They should only do so if a dependency has a specific known issue they are trying to avoid --- and even then, it's better to ban specific broken releases (`!= ...`) rather than pin down (`< ...`). Otherwise, second- and third-order dependencies very quickly run into irreconcilable requirements as other libraries they need may be pinning up.

At the same time, _applications_ definitely should provide a list of known-good versions of their dependencies, and I think we should treat our Docker image as an application. Could we add running the tests _in the container_ to the CI? Could we configure dependabot to keep the pins up to date?

untzag:
The published image is immutable, right? Deployments that pull from your repository will get exactly the same container every time (which they may then build on top of if desired). So what's the concern about dependency pinning? Are you trying to ensure that the container for an old release can rebuilt at a later date?

(asking purely out of ignorance, I'm trying to learn best practices in container workflows)

kleinhenz:
> Do we need to do anything special in the GH Action workflow to pass that configuration into the build-push-action?

We need to have
```
        with:
          context: .
          file: docker/Dockerfile
```
which is already done. Hopefully this will work when it actually runs.

dylanmcreynolds:
@untzag Imagine this scenario. Tiled depends on library `foo`. While I'm developing, I use a feature introduced in `foo==2.1`, (which I got because I just setup an environment.) Works great on my dev machines and passes unit tests. But nobody thought to update docker's requirements.txt.

Eventually, someone in production upgrades their docker container  to a new image build recently. Docker image was pinned to `foo==2.0`. 

I like both the ideas of keeping the pins up to date and running tests against the container. I don't know much about Dependabot. I love the idea of using it for vulnerabilities but I think we'd still be in the same situation if it was upgrading dependencies automatically without the result going through testing.

danielballan:
@kleinhenz Ah, I see, I had missed that. Thanks.

Would it be easy to add a second workflow, one that runs on every PR and commit push, that builds and then runs the tests, as in:

```
docker run --cidfile=cidfile -dit <the container we just built> /bin/bash
docker exec `cat cidfile` pip install -r requirements-dev.txt
docker exec `cat cidfile` pytest
```

Then, when dependabot opens a PR increasing the version of a given pin, we can tell from the CI whether that upgrade would break anything before we merge it.

danielballan:
In my experience, as long as the dependabot PRs run tests, it's as great way to get early notification that a new release of some dependency has broken something. Of course, most updates are benign, and the biggest drawback is that dependabot generates a _lot_ of PR traffic for projects with many dependencies. We could also try it and see if we find the volume overwhelming.

kleinhenz:
I added a test stage to the dockerfile so that the tests get run in the container during the build. The test dependencies aren't included in the final image so it is still reasonably slim. This can be incorporated in a separate workflow to run on pull requests like @danielballan suggested.

dylanmcreynolds (approved):

danielballan (approved):

PR #60, opened 2021-08-19
add workflow for testing container

kleinhenz:
This adds a new workflow for running the unit tests inside the docker image. As discussed before this will help us to discover problems with the pinned dependencies for the container. It works by building the test target defined in the Dockerfile. It is a separate workflow from the regular unit tests because the python version matrix doesn't make sense here.


dylanmcreynolds (commented):
Awesome.

danielballan (approved):

PR #65, opened 2021-08-28
Server-side caching, Part 1

danielballan:
## We need both an internal, in-process cache and a shared (e.g. Redis) cache

Reading some surveys of caching, such as [Microsoft's](https://docs.microsoft.com/en-us/azure/architecture/best-practices/caching), I understand there are two kinds of cache:

1. Private cache: held in the address space of a single worker process and accessed directly by the code that runs in that process. The main disadvantage is that each worker has to keep their own, so there may be redundancy (the same data taking up memory in N different workers' caches) and one must manage the possibility that workers' caches can get out of sync. The main advantage is that we can put runtime objects here without serializing them so it is extremely cheap to store and retrieve items.
2. Shared cache: an external service, such as Redis, available to all workers. The main advantage is that it is shared by all workers. The main disadvantage is the objects must be serialized and transferred over the (local) network, so accessing and storing objects is more expensive.

I believe that Tiled needs both of these kinds of cache.

(1) will be used by Tiled Adapters to make situation-specific optimizations that require caching various situation-specific objects, including unserialized objects like DataFrames and unserializable objects like file handles and h5py objects. Likewise, Adapters can employ situation-specific rules to invalidate (expire) cache entries when the underlying data changes. This cache will help Tiled efficiently serve distinct requests that access the same underlying resource (e.g. file, database query) but request different parts (e.g. columns, chunks) or ask for the response to be serialized into different formats.

For example, it can address the issue of slow disk access we are observing on NERSC Spin Community File System,  especially if we can afford to warm up the cache in advance — open and read all the files up front. (For AIMM's use case of curated XAS data that is on the scale of kilobytes, this is feasible.)

Separately, (2) can be used to efficiently serve repeat requests, such as when a user re-executes a Jupyter notebook or a script. An external (out-of-process) cache like Redis, shared across the deployment of workers, will cache Response bodies keyed on (hashed) Requests. In this case, since we have no access to situation-specific details, cache invalidation will be time-based (TTL). Clients can use standard HTTP headers to circumvent the cache.

Note that this cache does _not_ solve the slow disk access problem because it is not practical to warm up a cache with every possible request/response, especially when we consider requests for subsets of columns, multiplied across the range of supported formats.

This PR implements (1). A future PR will implement (2).

## This PR: an in-process cache

With this PR, the Tiled server now provides a global singleton cache, settable and gettable via `tiled.server.set_internal_cache(cache)` and `tiled.server.get_internal_cache()` respectively. The cache MAY be used by Adapter constructors such as `DataFrameAdapter.read_csv` to stash objects for future reuse. The cache manages eviction rules and ensures that the total memory usage stays below a configurable limit. As an implementation detail, it uses [dask/cachey](https://github.com/dask/cachey), a nice simple Cache implementation that takes into account the cost of each item and how often it has been used recently, with exponential decay.

The Adapter code does not engage with the details of how caching works: it doesn't need to create a cache or configure its rules; it just asks for the cache and uses it, as shown below. Switching to a different cache implementation with different rules won't change the Adapter code.

Cache invalidation can be managed at this layer or one layer up, for example by observing the file mtime of a file on disk.

```py
    @classmethod
    def read_csv(cls, *args, metadata=None, **kwargs):
        cache = get_internal_cache()
        if cache is None:
            return cls(dask.dataframe.read_csv(*args, **kwargs), metadata=metadata)
        else:
            cache_key = f"{cls.__module__}:{cls.__qualname__} {dask.base.tokenize((args, kwargs))}"
            df = cache.get(cache_key)
            if df is None:
                ddf = dask.dataframe.read_csv(*args, **kwargs)
                # A dask.dataframe does not know its byte size, so we use npartitions == 1
                # as a proxy for "small enough to cache".
                if ddf.npartitions == 1:
                    # Read the data now and cache it.
                    start_time = time.perf_counter()
                    df = ddf.compute()
                    cache.put(cache_key, df, cost=time.perf_counter() - start_time)
                    # Wrap the in-memory dataframe in dask for type stability.
                    return cls.from_pandas(df, npartitions=1, metadata=metadata)
                else:
                    # This is too large to cache. Defer reading the data.
                    return cls(ddf, metadata=metadata)
            else:
                # Return the cached result.
                # Wrap the in-memory dataframe in dask for type stability.
                return cls.from_pandas(df, npartitions=1, metadata=metadata)
```

Adapters can also provide a constructor that opts out of caching entirely.
```py
    @classmethod
    def read_csv_no_cache(cls, *args, metadata=None, **kwargs):
        return cls(dask.dataframe.read_csv(*args, **kwargs), metadata=metadata)
```

The details of what and when to cache, and what cache key to use, are highly situation dependent so they must be left up to each Adapter. There is a documented recommendation that the cache key SHOULD begin with `{cls.__module__}:{cls.__qualname__}` to avoid collisions between different Adapters.

The cache's maximum size can be configured via an environment variable, configuration settings, or CLI flag. It can also be switched off entirely. By default, it is on with a size of 1 GB (chosen to be small enough to work on modest laptops).

## Instrumentation

The cache keeps a running tally of hits and misses, which is useful for testing. It can also logs hits, misses, and stores to the standard error. This is quite verbose and is off by default, but it can be switched on via an environment variable or configuration setting. Sample logs:

```
$ TILED_INTERNAL_CACHE_LOG_LEVEL=INFO tiled serve directory --public example_files/
...

INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
Internal cache miss 'tiled.readers.dataframe:DataFrameAdapter 26667e1fc8e85aff4436b643aef3f11c'
Internal cache store 'tiled.readers.dataframe:DataFrameAdapter 26667e1fc8e85aff4436b643aef3f11c' cost=0.005780.3
INFO:     127.0.0.1:46078 - "GET /dataframe/full/another_table HTTP/1.1" 200 OK
Internal cache hit 'tiled.readers.dataframe:DataFrameAdapter 26667e1fc8e85aff4436b643aef3f11c'
INFO:     127.0.0.1:46082 - "GET /dataframe/full/another_table?columns=A HTTP/1.1" 200 OK
Internal cache hit 'tiled.readers.dataframe:DataFrameAdapter 26667e1fc8e85aff4436b643aef3f11c'
INFO:     127.0.0.1:46086 - "GET /dataframe/full/another_table?columns=B HTTP/1.1" 200 OK
```

Notice that the requests above are not exactly the same: they ask for the whole table, Column A, and Column B respectively, but all resolve to the same cache key. This is because the Adapter caches the DataFrame itself, which it can use as a basis for building responses to all of these requests. Again, a cache of requests and responses is too fine-grained to do this, which is why we need both this in-process cache and a shared response cache.

Tiled implements the [standard HTTP Server-Timing header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing). It always reports the total time that the application spends between receiving the request and sending the response. This lets us distinguish the network / HTTP overhead from the time spent inside our FastAPI app. On routes that return data payloads, we now additionally break out the time spent _reading_ data and the time spent _packing_ (a.k.a. serializing) it, reported alongside the total application time.

```
server-timing: read;dur=0.5, pack;dur=0.8, app;dur=6.4
```

This will make it easy to see when reading is the bottleneck and when additional Adapter-level caching may be helpful.

## Design Notes

**Why a global singleton?** We generally avoid global singletons, but in this case two things pushed me in that direction. (1) It is complex to pass the `cache` from the application all the way down into Adapters. It would require all the code in between to handle the cache and pass it down correctly. (2) A cache is a thing you really only want one of... you don't want to accidentally end up with two. The global-ness is slightly inconvenient for unit testing, but as you can see in the tests, it's not bad.

**Why is the cache used in the `read_csv` classmethod constructor and not more deeply inside the instance?** This makes it easy to make alternative constructors with different caching choices (e.g. different threshold, different cache key, ...).

**Should we add this to other Adapters beyond `read_csv`?** Yes, but let's do it in future PRs.


dylanmcreynolds:
Would it be possible to add a helper method to the cache that handles the getting, the putting and the various checks for `None` so that the caller doesn't have to?

Something like:
```
def get_cachable_value(key, cachable_function):
```
Then I guess each `cachable_function` would then have to return a tuple...the cachable object and a flag to indicate whether it's should actually be cached in this particular case?

tiled/server/compression.py
```diff
@@ -92,13 +94,17 @@ async def send_compressed(self, message: Message) -> None:
...
                     THRESHOLD = 1 / 0.9
                     if compression_ratio > THRESHOLD:
-                        headers[
-                            "X-Compression-Stats"
-                        ] = f"time={1000 * compression_time:.2f},ratio={compression_ratio:.2f}"
                         headers["Content-Encoding"] = self.encoding
                         headers["Content-Length"] = str(len(compressed_body))
                         headers.add_vary_header("Accept-Encoding")
                         message["body"] = compressed_body
+                        # The Server-Timing middleware, which runs after this
```
dylanmcreynolds, line 101: is this change related to caching? Or is it mostly a related improvement to instrumentation?

tiled/_tests/test_internal_cache.py
```diff
@@ -0,0 +1,59 @@
...
+    assert cache.hits == cache.misses == 0
+    client["data"]
+    assert cache.misses == 1
+    assert cache.hits == 0
+    client["data"]
+    assert cache.misses == 1
+    assert cache.hits == 1
+    client["data"]
+    assert cache.misses == 1
+    assert cache.hits == 2
```
dylanmcreynolds, line 35: Maybe add a test of the instrumentation of cached? A bug there could send future people on a terrible journey.

tiled/server/compression.py
danielballan: Yeah, just a related improvement to instrumentation.

tiled/_tests/test_internal_cache.py
danielballan: Cachey has at least [one test](https://github.com/dask/cachey/blob/789e03dbc64e4f39a56724f9b02643c02f704959/cachey/tests/test_cache.py#L99-L119) on this itself, but it's good idea to add an extra layer of protection here since it's important. Will do.

tiled/readers/dataframe.py
```diff
@@ -45,11 +45,50 @@ def read_csv(cls, *args, metadata=None, **kwargs):
...
+        --------
+
+        >>> DataFrameAdapter.read_csv("myfiles.*.csv")
+        >>> DataFrameAdapter.read_csv("s3://bucket/myfiles.*.csv")
+        """
+        cache = get_internal_cache()
+        ddf = dask.dataframe.read_csv(*args, **kwargs)
+        # A dask.dataframe does not know its byte size, so we use npartitions == 1
+        # as a proxy for "small enough to cache".
+        if (cache is not None) and ddf.npartitions == 1:
```
callumforrester, line 69: > Why is the cache used in the read_csv classmethod constructor and not more deeply inside the instance? This makes it easy to make alternative constructors with different caching choices (e.g. different threshold, different cache key, ...).

This sounds sensible but you might end up with many similar constructors across adapters. Perhaps a util module could reduce the duplication somewhat? You could have a shared `CacheSettings` dataclass and some functions which hold the non-`read_csv`-specific logic in here. You would still have room to add adapter-specific caching arguments to constructors.

danielballan: Yeah, this looks way too complex to be replicated into every Adapter. I think you are right. In fact, we can leverage dask's [opportunistic caching](https://docs.dask.org/en/latest/caching.html) mechanism for every adapter that uses dask. (But we don't have to restrict the cache to _only_ work with dask.) Commit on the way....

danielballan:
Refining my thinking further:

There are three kinds of things we want to cache:

(1) Chunks of array or DataFrame in memory that we have read and may want again later, in whole or in part
(2) HTTP responses with serialized data that we sent and may need to re-send if we receive an identical request again later
(3) File handles, h5py objects, etc.

Items of kind (1) and (2) could be stored in worker memory space or in external memory space (e.g. Redis) because they are serializable. I think we should name these two caches (1) "Data Cache" and (2) "Response Cache" respectively. We should make it possible, via configuration, to use storage inside or outside the worker for either of these. In a deployment where clients have an affinity to a specific worker, using worker memory will give better performance. In a deployment where client requests are balanced across workers, using an external memory space may give better performance.  

Only (3) must be stored in-process.

Both (1) and (2) must be configured with a memory limit: the resource of interest is RAM. (3) requires different resource limiting: something like "max files open". It's not clear whether it worth directly managing that limit.

For this PR, I intend to stay focused on one. I think it has evolved from being a general purpose "internal cache" to being in fact a "data cache" that may be in-process or out-of-process.

docs/source/explanations/caching.md
```diff
@@ -0,0 +1,100 @@
...
+   cache, on the server side. It stores the content of the most frequent responses.
+   This covers use cases such as, "Several users are asking for the exact same
+   chunks of data in the exact same format."
+3. **Service-side data cache.** The _response_ caches are very finely targeted.
+   Requests that ask for overlapping but distinct slices of data, or requests
+   that ask for the same but in varied formats, will not benefit from that
+   cache; they will "miss". The _data_ cache, however, stores chunks of (array,
+   dataframe) data, which is it can slice and encode differently for different
+   requests. It will not provide quite the same speed boost as a response cache,
+   but it has a broader impact.
```
dylanmcreynolds, line 30: Maybe something here about how 3 stores in-memory python objects? Also what would you think about calling it `Server-side object cache`?

dylanmcreynolds: Also...how about something about how it can be used by those coding readers?

danielballan: I'll grant you that "data cache" is a bit broad. I already went through and renamed it from "internal cache" (because of course it may not always be internal). I'm 50/50 on "object" being better, and I don't really have the heart to rename it again now but I am open to renaming it in the future.

danielballan: > Maybe something here about how 3 stores in-memory python objects?

I mention it below, but I think you are right it should be included here as well. Adding...

danielballan: I could foresee whee we tighten this down a bit, actually, and don't allow just _any_ Python object to be dumped in here. Our current usage is just numpy arrays and dataframes, which give us more options for serializing to Redis that we'll have if we fully embrace "anything Python".

danielballan: For the record: we agreed to sit on it and consider other names. "Data" is definitely overly broad --- all of these caches store data.

danielballan:
I think this is converging on "ready".

docs/source/how-to/tune-caches.md
```diff
@@ -0,0 +1,85 @@
+# Tune Caches to Balance Speed and Memory Usage
+
+This section describes a feature of the Tiled *server*. For client-side caching,
+see {doc}`../tutorials/caching`.
+
+## Data Cache
+
+The Tiled server Tiled stores chunks of recently-used data in worker memory.
```
danielballan, line 8: Typo

docs/source/explanations/caching.md
```diff
@@ -0,0 +1,125 @@
...
+
+with get_data_cache().dask_context:
+    # Any tasks that happen to already be cached will be looked up
+    # instead of computed here. Anything that _is_ computer here may
+    # be cached, depending on its bytesize and its cost (how long it took to
+    # compute).
+    dask_object.compute()
+```
+
+Items can be proactively clearly from the cache like so:
```
danielballan, line 99: Typo

docs/source/explanations/caching.md
```diff
@@ -0,0 +1,125 @@
...
+into dask in a similar way to make it easy for any Adapters that happen to use
+dask to leverage Tiled's data cache very simply, like this:
+
+```py
+from tiled.server.data_cache import get_data_cache
+
+
+with get_data_cache().dask_context:
+    # Any tasks that happen to already be cached will be looked up
+    # instead of computed here. Anything that _is_ computer here may
```
danielballan, line 93: Typo

docs/source/how-to/tune-caches.md
```diff
@@ -0,0 +1,85 @@
...
+  log_level: DEBUG  # case-insensitive
+```
+
+or the environment variable
+
+```
+TILED_DATA_CACHE_LOG_LEVEL=DEBUG  # case-insensitive
+```
+
+The debug interleave with the access logs from uvicorn like this.
```
danielballan, line 75: Missed word

danielballan (commented):
Copy editing notes to Future Me

tiled/_tests/test_internal_cache.py
danielballan: Done

tiled/readers/dataframe.py
danielballan: This is now much improved and simplified. Thanks for the feedback,  @callumforrester.

danielballan:
Over weekend chats with @dylanmcreynolds and @tacaswell I think we have a clear handle on the _scope_ for this cache, even if some details of its operation are still speculative, and "object cache" is a better name for it. I have done the renaming.

PR #86, opened 2021-09-23
Add ansible playbook for Ansible-based deployment

danielballan:
An alternative to Docker-based deployment. Build for and tested on https://tiled-demo.blueskyproject.io.


danielballan:
The approach in this PR still still valid, but I've come to prefer using Ansible+podman for this, and I think it's a more approachable way to do it for most people. I'm going to close this.

PR #89, opened 2021-09-28
Support 'Expires' header

danielballan:
In current `main` branch, we check the server every time using `If-None-Match` to confirm whether the content ID is unchanged. In this PR, server-side trees can signal that their metadata or content is unlikely to change until some given time. The client will then rely completely on its cache until that time, saving the latency of an `If-None-Match` / `304` handshake.

Needs tests.


tiled/client/cache.py
```diff
@@ -74,20 +89,55 @@ class Reservation:
...
 
+    @property
+    def etag(self):
+        return self.headers["ETag"]
+
+    @property
+    def expires(self):
+        expires = self.headers.get("Expires")
+        if expires is not None:
+            return datetime.strptime(expires, HTTP_EXPIRES_HEADER_FORMAT)
```
tacaswell, line 109:
```suggestion
            return datetime.strptime(expires, HTTP_EXPIRES_HEADER_FORMAT)
        return None
```

I think the explicit return is good for readability here.

tiled/client/context.py
```diff
@@ -247,27 +247,49 @@ def get_content(self, path, accept=None, timeout=UNSET, stream=False, **kwargs):
...
+        }
+        if "no-cache" in cache_control:
+            reservation = None
+        else:
+            reservation = self._cache.get_reservation(url)
         try:
             if reservation is not None:
+                if not reservation.is_stale():
+                    # Short-circuit. Do not even bother consulting the server.
+                    return reservation.load_content()
```
tacaswell, line 263: either do multiple returns or single return is this function, doing a direct return here and then setting up `content` to be the return value below was confusing.

danielballan: Done

tiled/client/cache.py
danielballan: Done

PR #97, opened 2021-10-12
Adjust for breaking change in tifffile 2021.10.10.

danielballan:


cgohlke:
Sorry, this wasn't intended to break anything. I don't think I ever used that feature... I could release a new version of tifffile if you can provide a small test case of the API and cases you rely on.

danielballan:
Thanks @cgohlke! I'm just catching up and you already fixed it. You are a hero!

dylanmcreynolds:
So this PR is now not needed?

danielballan:
Well, it's still worth having because the old name is deprecated and I think we want to support old versions of tifffile for awhile.

PIp doesn't always respect minimum version pins. It's getting better about that, but lab software environment often have old pip.

tiled/readers/tiff_sequence.py
```diff
@@ -151,3 +151,17 @@ def macrostructure(self):
                 (shape[2],),
             ),
         )
+
+
+def _safe_asarray(seq, index=None):
+    """
+    Wrap FileSequence.asarray to cope with this brekaing change in tifffile.
```
clintonroy, line 158: breaking

PR #110, opened 2021-10-18
Added orjson

jmaruland:


tiled/_tests/test_array.py
```diff
@@ -66,3 +68,30 @@ def test_shape_with_zero():
     client = from_tree(tree)
     actual = client["test"].read()
     assert numpy.array_equal(actual, expected)
+
+
+def test_nan_infinity_handler(tmpdir):
+    data = numpy.array([0, 1, numpy.NAN, -numpy.Inf, numpy.Inf])
+    metadata = {"infinity": math.inf, "-infinity": -math.inf, "nan": numpy.NAN}
```
danielballan, line 75: Can you add `numpy.inf`, `-numpy.inf`, and `math.nan` here too, so that we are testing both the `math` and `numpy` varieties of each thing here in `metadata`?

You don't need to test both with `data` ---- putting `math.inf` inside a numpy array would force the array to be "object" type, which Tiled doesn't handle by default --- so just testing both `math` and `numpy` in `metadata` is fine and testing only numpy in `data` is fine.

PR #111, opened 2021-10-18
Reduce number of requests required to load dataframes

jmaruland:


danielballan:
To capture some discussion from a video call here:

I'd like to enlarge the scope of this PR. Instead of placing the `"meta"` and `"divisions"` structure in separate endpoints, I think we should place it inline, as in:

```
# json
{
  ...,
  "structure_family": "dataframe",
  "structure":
    "macro": {
      "npartitions": 10,
      "columns": ["A", "B", "C"],
    },
    "micro": {
      "meta": "data:vnd.apache.arrow.file;base64,{base64_encoded_arrow_serialized_data}"
      "divisions": "data:vnd.apache.arrow.file;base64,{base64_encoded_arrow_serialized_data}"
    }
}
````

where we are using [Base64 encoding](https://developer.mozilla.org/en-US/docs/Glossary/Base64) to place the binary Arrow-serialized "meta" and "divisions" inside JSON and using [DATA URIs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) to make the format self-describing in JSON. For msgpack-encoded responses, I think the (unicode) Data URI standard does not make sense, so in that case we should just place the Arrow-encoded binary data there directly. Anyone using our msgpack API is one step beyond the basics and, I think we can assume, has consulted our docs and knows what they are getting.

```
# msgpack
{
  ...,
  "structure_family": "dataframe",
  "structure":
    "macro": {
      "npartitions": 10,
      "columns": ["A", "B", "C"],
    },
    "micro": {
      "meta": "{arrow_serialized_data}"
      "divisions": "{arrow_serialized_data}"
    }
}
````

danielballan:
A relevant part of the code is

https://github.com/bluesky/tiled/blob/279386062a56c8fa50ed7d307a152383db0b9910/tiled/server/core.py#L493-L502

danielballan:
Superseded by #132 which puts meta and divisions in line as proposed above and also caches structure information from the first request that gets it, making the rest of this PR moot.

PR #121, opened 2021-10-25
wrapped csv serialization in try/except case to include additional er…

jmaruland:
…ror cases

### DIFF
--- a/tiled/structures/array.py
+++ b/tiled/structures/array.py
@@ -167,11 +167,12 @@ if modules_available("orjson"):
 
 
 def serialize_csv(array, metadata):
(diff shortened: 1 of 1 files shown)


danielballan:
Can you give an example of a case this would catch that the current approach would not catch? I like that the current approach is more targeted, so I'd like to understand the reason for making it broader.

jmaruland:
I tracked an error all the way down to this method when Dylan was trying to read a TiffSequence dataset with the browser. The TiffSequence reader generates a 3D NumPy array and the CSV serialization fails at this point because it expects to receive a 2D array

Issue #88, opened 2021-09-24
Build intake interoperability


danielballan:
We know we want to do this, but I don't think we have an issue yet to track it. In broad strokes:

- [ ] Add an `IntakeAdapter` which wraps `intake.Catalog`. When it reaches a (nested) entry that is not a `Catalog`, it can use `to_dask()` to obtain a sufficiently self-describing structure.
- [ ] Add some kind of hook to `tiled serve ...` for serving intake YAML configs.
- [ ] Build a client-side adapter.

PR #122, opened 2021-10-26
POC: intake

martindurant:
Example code of how client-side integration with Intake might look, i.e., have Intake use the Tiled server seamlessly as a data portal.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,11 @@ setup(
     install_requires=[],  # Requirements depend on use case (e.g. client vs server).
     extras_require=extras_require,
     packages=find_packages(exclude=["docs", "tests"]),
-    entry_points={"console_scripts": ["tiled = tiled.commandline.main:main"]},
+    entry_points={
+        "console_scripts": ["tiled = tiled.commandline.main:main"],
+        'intake.drivers': [
+            'tiled = tiled.intake:TiledCatalog']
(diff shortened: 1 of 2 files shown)


danielballan:
Oh, cool! I just saw this. Thanks @martindurant!

martindurant:
@danielballan , related: you mentioned a generic IntakeReader that Tiled could use to input sources. Does this exist, or was it just an idea? A POC should be easy, but I don't know how you deal with sources where opening the metadata alone is expensive.

danielballan:
xref https://github.com/bluesky/tiled/issues/88

danielballan:
Notes on call with @tacaswell and @martindurant:

* Let's put the code from this PR into intake. This will add an option dependency on `tiled[client]` to intake. This enables intake to read data from a tiled server. We may also add `intake.open_tiled("https://...")`. For testing, it would be useful if it accepted a live tiled client object as well as a URL.
* Let's add intake adapters to tiled. This will not require an intake import. A factory function will consume an intake object (DataSource or Catalog). It will consult the `container` to map it onto a structure family and associated adapter class. It uses `to_dask()` to obtain what it needs for `macrostructure()` and `microstructure()` relatively cheaply. In the future, intake may add machine-readable schemas; in that case, Tiled should try those first.

PR #127, opened 2021-11-04
Enable opting in to SWMR mode for reading.

danielballan:


danielballan:
I think this is worth adding. We may decide to make SWMR opt _out_ rather than opt in in the future. But this PR is a useful backward-compatible step.

tacaswell (approved):

PR #130, opened 2021-11-05
read full dataframe directly in DataFrameClient

kleinhenz:
If we are going to get a pandas dataframe rather than a dask dataframe avoid going through the dask object as an intermediate. Avoids needing to make a separate metadata request. Improves throughput ~2x as expected. alternative to #111.

### DIFF
--- a/tiled/_tests/test_object_cache.py
+++ b/tiled/_tests/test_object_cache.py
@@ -74,19 +74,19 @@ a,b,c
     cache = get_object_cache()
     assert cache.hits == cache.misses == 0
     client["data"].read()
-    assert cache.misses == 2  # two dask objects in the cache
+    assert cache.misses == 1  # two dask objects in the cache
     assert cache.hits == 0
     client["data"].read()
-    assert cache.misses == 2
-    assert cache.hits == 2
+    assert cache.misses == 1
+    assert cache.hits == 1
     # Simulate eviction.
     cache.clear()
     client["data"].read()
-    assert cache.misses == 4
-    assert cache.hits == 2
+    assert cache.misses == 2
+    assert cache.hits == 1
     client["data"].read()
-    assert cache.misses == 4
-    assert cache.hits == 4
+    assert cache.misses == 2
+    assert cache.hits == 2
 
 
 def test_object_cache_disabled(tmpdir):
--- a/tiled/client/dataframe.py
+++ b/tiled/client/dataframe.py
@@ -230,7 +230,16 @@ class DataFrameClient(DaskDataFrameClient):
         """
         Access the entire DataFrame. Optionally select a subset of the columns.
         """
-        return super().read(columns).compute()
+        content = self.context.get_content(
+            f"/dataframe/full/{'/'.join(self.context.path_parts)}/{'/'.join(self._path)}",
+            params=self._params,
+            headers={"Accept": APACHE_ARROW_FILE_MIME_TYPE},
+        )
+        df = deserialization_registry("dataframe", APACHE_ARROW_FILE_MIME_TYPE, content)
+
+        if columns is not None:
+            df = df[columns]
+        return df
 
     def download(self):
         # Do not run super().download() because DaskDataFrameClient calls compute()


danielballan:
I believe this is faster for small dataframes but slower for large ones. Can you test it on a large (10s of GB) DataFrame to see?

kleinhenz:
yeah this is slower for a large partitioned dataframe I guess because you have to pay the cost to materialize the whole thing on the server instead of sending the chunks you already have. For the case with a single partition there is no difference as expected. I get ~6s vs ~12s for a dataframe with one hundred million rows, three float64 columns and 500 partitions. My argument would be that if you are dealing with giant data you should use the dask dataframe client and if you aren't you are saying that your data is a reasonable size to materialize all at once.

danielballan:
There is a use case for "My data fits in RAM so I want the `numpy` client, but it would still benefit from being fetched in parallelized partitions."

danielballan:
We want to support use cases where the structure of the data may be dynamic, such as rows appended to a table or images appended to an image series. With that in mind, the client always pulls the `structure` immediately before reading. That way, if the table had 3 partitions when we first create the `DataFrameClient` but now has 6, we will find out about it. Observe that the client already _knows_ `npartitions` is `1` before we `read()`. The only reason it asks again is to verify it hasn't changed under us.

```py
In [3]: c = from_uri("https://aimm.lbl.gov")

In [4]: d = c['wanli']['021879']

In [5]: d.item
Out[5]: 
{'id': '021879',
 'type': 'reader',
 'meta': None,
 'links': {'self': 'https://aimm.lbl.gov/metadata/wanli/021879',
  'full': 'https://aimm.lbl.gov/dataframe/full/wanli/021879',
  'partition': 'https://aimm.lbl.gov/dataframe/partition/wanli/021879?partition={index}'},
 'attributes': {'metadata': {},
  'structure_family': 'dataframe',
  'structure': {'macro': {'npartitions': 1,
    'columns': ['energy', 'i0', 'tey', 'tfy', 'i0_alt']},
   'micro': {'links': {'meta': 'https://aimm.lbl.gov/dataframe/meta/wanli/021879',
     'divisions': 'https://aimm.lbl.gov/dataframe/divisions/wanli/021879'}}},
  'specs': None}}
```

As you've shown, that consistency carries a 2X performance cost (likely from request latency). For the use case where the data is static, this is a very poor trade off. A couple months ago, I talked through with @tacaswell and @dylanmcreynolds (separately) the idea of designating which aspects of a structure the client should expect to be static. It may be time to implement that now. I believe we suggested something like:

```
"structure": {
    "micro": ...,
    "macro": {"npartitions": 3, "columns": ["A, "B", "C"]}},
    "dynamic": {"columns": "static", "rows": "static"}
```

where the choices may be something like `"static"` (nothing will change), `"appendable"` (append-only changes, no insertions or deletions), and `"dynamic"` (free for all).

The Adapters would grow a `dynamic_structure` method alongside their existing `microstructure` and `macrostructure` methods. In the case of AIMM data, they could use this to designate that everything is `static` and the client could then know it is safe to rely on the structure without re-checking it at each access.

In the short term, to test this approach we could punt on a _general_ solution and simply gives the AIMM adapters a spec (e.g. `"short_table"`) and use a `special_client` like so:

```py
# server side

class MyAdapter:
    spec = "short_table"
    ...


# client side

class ShortDataFrameClient(DataFrameClient):
    def read(self, ...):
        # Use /full route without re-fetching structure.
        ...

c = from_uri("...", special_clients={"short_table": ShortDataFrameClient})
```

That is, this idea of a new `dynamic` designation in `structure` would be providing a general way to provide this signal across specs, but clients are always free to do custom performance optimizations based on special knowledge.

You could even decide that, in your work, you know that you only deal with short tables, and do this:

```
c = from_uri("...", structure_clients={"dataframe": ShortDataFrameClient})
```

danielballan:
Revisiting this:

> We want to support use cases where the structure of the data may be dynamic, such as rows appended to a table or images appended to an image series. With that in mind, the client always pulls the structure immediately before reading.

I now appreciate that the latency cost of always (re-)pulling the structure information is too high. In #132, the client checks for a new field in the macrostructure, `resizable`. If `resizable` is `False`, the client caches the structure information. If `resizable` is not `False`, the client currently raises an `NotImplementedError`; we can revisit that in the future.

danielballan:
I think this is superseded by #132. If you agree @kleinhenz please close. Thanks!

PR #139, opened 2021-12-02
fix cors middleware

kleinhenz:
cors middleware can only be initialized after settings have been read


danielballan:
I see. Setting this by environment variable, [as we do in the tutorial](https://blueskyproject.io/tiled/tutorials/plotly-integration.html) and in many live demos we've given, works. Setting it via config has probably never worked.

danielballan (approved):

PR #140, opened 2021-12-05
GraphQL Implementation

dylanmcreynolds:
This draft PR represents thoughts related to creating a GraphQL search implementation in tiled. I'm creating the PR very early in development to maximize feedback.

This uses the [Ariadne](https://ariadnegraphql.org/) graphql package, which focuses on "schema first" development.

Currently, this only supports a "hello world" endpoint at `/graphql`.

The general theory is that this will always support a single endpoint into which Tree-specific schemas and code can be registered at startup. At app startup, the code in graphql.py will be notified of the `root tree` object. With that, it will interrogate top level `nodes` for registered queries. A `gqlquery` will provide:
- a query part (possibly a graphql `fragment` 
- one or more resolver iplementations

In this way, developers who register tiled a `Tree` will be able to optionally specify a graphql schema (string) and implementation (python).

### DIFF
--- a/requirements-server.txt
+++ b/requirements-server.txt
@@ -1,4 +1,5 @@
 # These are the requirements needed for basic server functionality.
+ariadne
 cachey
 cachetools
 dask
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -6,6 +6,7 @@ import sys
 import urllib.parse
 from functools import lru_cache, partial
 
+from ariadne.asgi import GraphQL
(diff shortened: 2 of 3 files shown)


danielballan:
I like this approach. It fits in well with our existing patterns.

danielballan:
Power-cycled to re-run CI now that docs build has been fixed on `main`.

danielballan:
I'm going to close this as stale. I am still interested in understanding use cases for GraphQL in Tiled.

PR #143, opened 2021-12-08
Support RGB TIFFs.

### LINKED ISSUE #133: RGB TIFF Support
Tried to use Tiled to access an RGB dataset of colored TIFF images and received the following error--
tiled.client.utils.ClientError: 400: Block index out of range

danielballan:
Closes #133.


danielballan:
Power-cycled to re-run CI now that docs build has been fixed on `main`.

PR #148, opened 2022-01-08
Obtain base_url in a more robust and standard way

### LINKED ISSUE #146: Fix base URLs.
When tiled is served behind gunicorn and/or nginx, it provides the wrong base URL in the `links` it hands the client. Example:

```
$ http https://tiled-demo.blueskyproject.io/metadata/bmm | jq .data.links
{
  "self": "http://localhost:8000/metadata/bmm",
  "search": "http://localhost:8000/search/bmm"
}
```

Somehow we need to propagate the relevant information through. Off the top of my head, maybe this could involve...

* Grab it from the `Host` header (seems simplest if it is a general solution)
* Do something with `X-Forwarded-For`
* Extract something from the ASGI protocol

danielballan:
Closes #146 

To test, I deployed this on https://tiled-demo.blueskyproject.io/, which serves Tiled using gunicorn behind nginx. The scheme and the host are both correct:

```
$ http https://tiled-demo.blueskyproject.io/node/metadata/ | jq .data.links
{
  "self": "https://tiled-demo.blueskyproject.io/node/metadata/",
  "search": "https://tiled-demo.blueskyproject.io/node/search/"
}
```

We also care about the niche use case of SSH port-forwarding to access a Tiled deployment that is only available on an internal network (such as `tiled.nsls2.bnl.gov` is now). HTTPS over port-forwarding is janky (it breaks SSL) but it has been useful in the early stages. Fortunately, it is supported by the approach in this PR:

```
$ ssh -L 8000:localhost:443 tiled-demo.blueskyproject.io
...
$ http --verify false "https://localhost:8000/node/metadata/" | jq .data.links
{
  "self": "https://localhost:8000/node/metadata/",
  "search": "https://localhost:8000/node/search/"
}
```

It works because the `Host` header in the request is set to `localhost:8000` and propagated through nginx as `X-Forwarded-Host`.

I had previously supported this use case with a custom `x-base-url` header, set by the Python client and observed by the server. This is not necessary, and it is removed in this PR.

### DIFF
--- a/tiled/server/router.py
+++ b/tiled/server/router.py
@@ -485,6 +485,8 @@ def _get_base_url(request):
     #   to specify a non-default port.
     #   https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23
     host = request.headers.get("x-forwarded-host", request.headers["host"])
-    scheme = request.headers.get("x-forwarded-scheme", request.url.scheme)
+    scheme = request.headers.get("x-forwarded-proto", request.url.scheme)
     root_path = request.scope.get("root_path") or "/"
+    if not root_path.endswith("/"):
+        root_path = f"{root_path}/"
     return f"{scheme}://{host}{root_path}"
--- a/tiled/server/router.py
+++ b/tiled/server/router.py
@@ -465,19 +465,26 @@ def node_full(
 
 
 def _get_base_url(request):
-    # Confusing thing:
-    # An httpx.URL treats netloc as bytes (in 0.18.2)
-    # but starlette.datastructures.URL treats netloc as str.
-    # It seems possible starlette could change their minds in
-    # the future to align with httpx, so we will accept either
-    # str or bytes here.
-    client_specified_base_url = request.headers.get("x-base-url")
-    if client_specified_base_url is not None:
-        return client_specified_base_url
-    url = request.url
+    # We want to get the scheme, host, and root_path (if any)
+    # *as it appears to the client* for use in assembling links to
+    # include in our responses.
+    #
+    # We need to consider:
+    #
+    # * FastAPI may be behind a load balancer, such that for a client request
+    #   like "https://example.com/..." the Host header is set to something
+    #   like "localhost:8000" and the request.url.scheme is "http".
+    #   We consult X-Forwarded-* headers to get the original Host and scheme.
+    #   Note that, although these are a de facto standard, they may not be
+    #   set by default. With nginx, for example, they need to be configured.
+    #
+    # * The client may be connecting through SSH port-forwarding. (This
+    #   is a niche use case but one that we nonetheless care about.)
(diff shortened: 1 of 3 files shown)


tiled/server/router.py
```diff
@@ -465,19 +465,26 @@ def node_full(
...
+    #   https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23
+    host = request.headers.get("x-forwarded-host", request.headers["host"])
+    scheme = request.headers.get("x-forwarded-scheme", request.url.scheme)
     root_path = request.scope.get("root_path") or "/"
-    if isinstance(url.netloc, bytes):
-        netloc_str = url.netloc.decode()
-    else:
-        netloc_str = url.netloc
-    return f"{url.scheme}://{netloc_str}{root_path}"
+    return f"{scheme}://{host}{root_path}"
```
dylanmcreynolds, line 490: This line depends on root_path, if present, always starting with a `/`. Is that always dependable?

tiled/server/router.py
```diff
@@ -465,19 +465,26 @@ def node_full(
...
+    #   set by default. With nginx, for example, they need to be configured.
+    #
+    # * The client may be connecting through SSH port-forwarding. (This
+    #   is a niche use case but one that we nonetheless care about.)
+    #   The Host or X-Forwarded-Host header may include a non-default port.
+    #   The HTTP spec specifies that the Host header may include a port
+    #   to specify a non-default port.
+    #   https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23
+    host = request.headers.get("x-forwarded-host", request.headers["host"])
+    scheme = request.headers.get("x-forwarded-scheme", request.url.scheme)
```
dylanmcreynolds, line 488: This looks wrong. The header is `x-forwarded-proto`, not `x-forwarded-scheme`

dylanmcreynolds (changes requested):

tiled/server/router.py
dylanmcreynolds: Also, can we depend on headers at this point being lower case? HTTP headers are not case sensitive.

dylanmcreynolds (changes requested):
I have some thoughts to look at.

dylanmcreynolds (changes requested):
Added a couple comments I'd like to review.

tiled/server/router.py
danielballan: Oops, I misspelled this consistently in the nginx config and here, which is why the test worked. Fixed.

The `headers` object is a special case-insensitive dict.

tiled/server/router.py
danielballan: It looks like there are no guarantees that it has a trailing slash, so I have added a check that ensures it gets one.

dylanmcreynolds (approved):

PR #150, opened 2022-01-10
Support multiple identity providers and API keys

### LINKED ISSUE #120: Support multiple Authenticators
- [ ] Adopt a richer model of an authenticated entity, upgrading from a simple `str` to a model with `entity` (user or service name), `entity_type` (enum: `user` or `service`), and `IdP`.
- [ ] Let an Authenticator own its routes and put them at scoped subpaths like `/auth/orcid/code`.
- [ ] Update the config to accept multiple Authenticators.
- [ ] Update the authentication info published at `/` and update the client to give the user a choice when there are multiple authentication methods available.

danielballan:
Closes #120 

This PR adds the concept of a Principal (a user or service) that can be authenticated via more than one identity provider, or via tiled API keys.

Checklist:

- [x] Support more than one identity provider (authenticator) in the server configuration.
- [x] Make the user model richer, moving from a string username to a model with `id` and `provider`. (And name it "Principal" to cover non-user entities.)
- [x] Update SimpleAccessPolicy to expect this richer model.
- [x] Provide more links in the `/` route and hard-code less in the client.
- [x] Remove notion of `authentication_uri`; let the server tell us what our options are.
- [x] Add multiple-choice prompt in the client when multiple providers are available.
- [x] Add sqlalehemy ORMs and alembic configuration.
- [x] Add pydantic models to match ORMs.
- [x] Use Principal model. Place `principal.uuid` instead of `username` in tokens. Let `/auth/whoami` return everything we know about the Principal, including the list of all Identities. But do not leak which Identity authenticated the current session. ("You are you.") 
- [x] Add scopes to routes.
- [x] Add route for revoking tokens.
- [x] Add routes for minting and revoking API keys.
- [x] Add logic for initializing database and verifying that it has the correct revision.
- [x] Tests


tiled/database/alembic_utils.py
```diff
@@ -0,0 +1,79 @@
...
+def run_alembic(args=None):
+    """
+    Run the alembic command against a temporary alembic.ini file.
+
+    The file will is generated from a template to include the database URI and
+    the location of the alembic migration scripts.
+    """
+    if args is None:
+        args = sys.argv[1:]
+    database_uri = "sqlite:///tiled.sqlite"  # TEMP
```
dylanmcreynolds, line 73: This looks odd, but I don't have any experience with alembic. Rather than hard-coding the database_uri, wouldn't you want to get it from settings so that you're upgrading the same database? Or do I misunderstand what's goign on?

tiled/database/alembic_utils.py
```diff
@@ -0,0 +1,79 @@
...
+                # YAY for nested templates?
+                database_uri=str(database_uri).replace("%", "%%"),
+            )
+        )
+
+
+@contextlib.contextmanager
+def temp_alembic_ini(database_uri):
+    """
+    Context manager for temporary JupyterHub tiled directory
```
dylanmcreynolds, line 42: Jupyterhub?

tiled/database/orm.py
```diff
@@ -0,0 +1,195 @@
...
+    principal = relationship("Principal", back_populates="identities")
+
+
+class Role(Timestamped, Base):
+    __tablename__ = "roles"
+
+    id = Column(Integer, primary_key=True, index=True, autoincrement=True)
+    name = Column(Unicode(255), index=True, unique=True)
+    description = Column(Unicode(1023), nullable=True)
+    scopes = Column(JSONList, nullable=False)
```
dylanmcreynolds, line 144: JSONlist and not JSONList(255)?

danielballan: Good catch!

danielballan: Actually, now that I think about it, I left this intentionally unbounded, following JupyterHub's example. The list of scopes may be long.

dylanmcreynolds: If the list of scopes is long and approaching unbounded, then should we put them in a mapping table rather than stuffing them in an unbounded column?

danielballan: Fair point. Given the choice I prefer putting a bound on the list.

danielballan: (For now)

tiled/database/alembic_utils.py
danielballan: This was leftover dead code. Deleted now.

danielballan:
This has received conceptual review from @dylanmcreynolds and @tacaswell and _some_ line review from @dylanmcreynolds. It has become fairly unreviewable, given many many changes related to supporting API keys and identities. I will merge so that follow-up work can proceed in reviewable chunks.

PR #155, opened 2022-01-17
HDF5 File open in swmr mode

OCopping:
Added env var to open hdf5 files in swmr mode by default, although can be disabled by setting env var to 0.  Throws warning if hdf5 file can't be opened.


danielballan:
Oh, I think I misread @jacobfilik on GitHub:

> i think we pretty much just open everything with libver latest and swmr True...and just handle the exception if the file cant be opened yet

I was thinking:

1. Try with `swmr=True` / `libver="latest"`.
2. Fall back to the default if the file can't be opened.
3. If that fails (because the file is being written to) raise.

I think warning and returning `None` is liable to create confusing errors downstream. If we can't open the file as directed, we should raise and let the caller decide what it wants to do. The correct response (Exit with failure? Skip the file for now but try again later?) depends on context that we cannot know at this layer.

I haven't got much hands-on experience with SWMR. What happens when you create a file _without SWMR_, close it (i.e. finish writing), and then try to open it with `swmr=True`?

danielballan:
Closing in favor of #127 which makes SWMR opt _in_ rather than opt out.

I'm open to change to opt-out behavior in the future, but opt-in feels like the right initial step.

PR #157, opened 2022-01-25
Fixed HDF5 dataset variable-length string handling

OCopping:


OCopping:
The weird thing is, both of the cases that trigger this error seem to already be in `array_cases`:
https://github.com/bluesky/tiled/blob/38b99563b20582278c603c9b0052e6e2bce0a5cb/tiled/_tests/test_array.py#L28-L29

danielballan:
To handle the attrs issue, could you try something like:

```py
class MockHDF5Dataset:
    "Mock just enough of the HDF5 Dataset interface to act as a placeholder."
    def __init__(self, array, attrs):
        self.array = array
        self.attrs = attrs
    def __getitem__(self, key):
        return self.array.__getitem__(key)
```

The numpy array subclass accomplishing the same thing, but someone encountering this codebase and finding an _ndarray subclass_ might quickly conclude that we are doing something much more complex than we really are and be put off.

tiled/utils.py
```diff
@@ -8,6 +8,7 @@
 import os
 import sys
 import threading
+import numpy
```
danielballan, line 11: This approach, falling back to `default`, is great.

My one note is that we try to avoid importing the "heavy" libraries like `numpy` unless/until they are used. This gets us two things:

1. For minimalist applications that only require access to basic metadata—say, a process that counts the number of scans done so far today—the environment doesn't need to install `numpy`/`pandas`/`xarray`.
2. The top-level import `import tiled` runs quickly, and tools like the CLI `tiled profile list` or `tiled serve ...` aren't slowed down by waiting for these large packages to import (and then not even be used).

So, the goal is to avoid importing numpy unless it is has already been imported somewhere else in the program (at which point `import numpy` is basically free). Here's what we can do. If numpy hasn't been imported yet, we can be sure that there are no instances of `numpy.str_` or `numpy.bytes_`. Therefore, we can protect the import like this:

```py
    def default(content):
        # If numpy has not been imported, we can be sure that there are no
        # instances of numpy objects, and we don't want to pay for the numpy
        # import / dependency if it's not otherwise used.
        if "numpy" in sys.modules:
            import numpy

            if isinstance(content[0], (numpy.str_, numpy.bytes_)):
                return content.tolist()
        return super().default(content)
```

danielballan (commented):
Looking good. I have one fiddly suggestion to help keep other parts of tiled fast.

A couple tests in `test_hdf5.py` and this will be good to merge. :+1:

tiled/utils.py
```diff
@@ -487,8 +486,16 @@ def safe_json_dump_array(array):
...
+        # If numpy has not been imported, we can be sure that there are no
+        # instances of numpy objects, and we don't want to pay for the numpy
+        # import / dependency if it's not otherwise used.
+        if "numpy" in sys.modules:
+            import numpy
+
+            if isinstance(content[0], (numpy.str_, numpy.bytes_)):
+                return content.tolist()
+
+        return list(bytes(content))
```
danielballan, line 498: I think the proper behavior in this line is

```py
raise TypeError
```

as illustrated in [the section on `default` in the orjson docs](https://github.com/ijl/orjson#default). We want to say, "We give up; we cannot convert this object to JSON," and let the user figure it out. Converting to `bytes(...)` may fail, depending on what type `content` is, and even if it works it's not clear that the result will be sensible in this context.

OCopping: Ah okay, I did read that in the docs and tried it, but wasn't sure if we wanted it to just accept defeat :sweat_smile: Will change it to that now

danielballan:
Cool. The :x: above from our linters means that you should

```
pre-commit install
pre-commit run --all-files
```

Once install `pre-commit` will prohibit you from committing work that has a linting issue. More info at https://pre-commit.com/.

With a couple tests, this will be ready to merge.

OCopping:
So it seems adding the ```python
raise TypeError
```
in `default()` now breaks the function. With that line included it now says it can't serialize `bytes` or even `numpy.ndarray`, which are both supported by orjson...
```
  File "/scratch/tiled/tiled/utils.py", line 505, in safe_json_dump_array
    return orjson.dumps(array, default=default)
TypeError: Type is not JSON serializable: bytes
```
Any ideas what could be causing this? I initially thought it could be causing a recursion with the `except TypeError:` below, but after testing this doesn't seem to be the case

**Comment by danielballan** [comment]
I think the problem occurs when we hand `safe_json_dump_array(…)` a dict that contains both a numpy str and a normal numpy. Maybe we need to remove the `except TypeError:` and consolidate into one orjson call that uses both OPT_SERIALIZE_NUMPY and default.

While we are at it we should rename this utility function to just safe_json_dump because we are handing it things other than an array, things like a dict that may contain arrays or other things.

**Comment by OCopping** [comment]
So it turns out the code probably wasn't working as intended as explained in the `default` section of the `orjson` docs: 
```
It is important that default raise an exception if a type cannot be handled. Python otherwise implicitly returns None, which appears to the caller like a legitimate value and is serialized:
```
Because of this addition, `TypeError`s for `bytes` began showing up. I've added a check for `bytes` to convert to `str` before heading back to `orjson.dumps()`, as there is no option to keep existing bytes objects untouched when serializing. I know this probably isn't ideal, so interested to hear your feedback.

**Comment by danielballan** [comment]
I think you are right: we have no better option than to convert that `bytes` to `str`. JSON is not a binary format, so there is no way to place `bytes` in it directly.

Playing around with some invented examples, I've come around to thinking that we should lean on `orjson`'s internal recursion more. As an example, trying to encode

```py
{"a": numpy.array([1,2,3]), "b": numpy.bytes_(b"abc"), "c": numpy.array([b"abc", b"def"], dtype=numpy.bytes_)}
```

with the implementation in this PR fails with

```
TypeError: Type is not JSON serializable: numpy.bytes_
```

because there is an `numpy.bytes_` object nested deeply enough (under the key `"c"`) that `default` doesn't see it. Instead of trying to look inside `content` using expressions like `content[0]`, we can lean on `orjson`'s recursion. If `default` returns an object that isn't serializable, `orjson` calls `default` again (and again, up to some high limit).

```py
In [1]: import numpy

In [2]: metadata = {"a": numpy.array([1,2,3]), "b": numpy.bytes_(b"abc"), "c": numpy.array([b"abc", b"def"], dtype=numpy.bytes_)}

In [3]: def default(content):
   ...:     if isinstance(content, numpy.ndarray):
   ...:         # If we make it here, OPT_NUMPY_SERIALIZE failed because we have hit some edge case.
   ...:         # Give up on the numpy fast-path and convert to Python list.
   ...:         # If the items in this list aren't serializable (e.g. bytes) we'll recurse on each item.
   ...:         return content.tolist()
   ...:     if isinstance(content, (bytes, numpy.bytes_)):
   ...:         return content.decode()
   ...:     raise TypeError
   ...: 

In [4]: import orjson

In [5]: orjson.dumps(metadata, option=orjson.OPT_SERIALIZE_NUMPY, default=default)
Out[5]: b'{"a":[1,2,3],"b":"abc","c":["abc","def"]}'
```

Does this solution address the examples you are looking at as well?

OCopping:
Ah I see, I misunderstood what `orjson` meant by the recursion. That solution works great!

OCopping:
I've written a test, although I'm not convinced this is the same as the real example we have at Diamond. I have to use `safe_json_dump()` in the h5py creation otherwise it gets the same `dtype("O")` issue as before. What do you think?

danielballan:
I think that nested metadata will arise in other context—such as with Bluesky, where metadata is stored as highly-nested Mongo documents—so it is good to have solid support for it in `safe_json_dumps`. But for HDF5 tests in particular, I think the _nested_ aspect isn't relevant because HDF5 attributes aren't allowed to be nested. As you observed, placing a `dict` in an attribute fails with ` dtype("O")` error. One solution, as you said, is to encode the content as JSON first and then make it an attribute, so that from HDF5's point of view it is just a single top-level attribute with a string. That's valid, but not likely to be seen in the wild.

My suggestion would be to test `safe_json_dumps` on its own without involving the rest of Tiled's machinery (maybe a new test file, `tiled/_tests/test_json.py`?) to cover general test cases, like the nestedness.

In `test_hdf5.py`, focus on recreating the key aspects of your real example. I guess that would include `numpy.str_`, `numpy.bytes_` in HDF5 attributes.

danielballan:
Documentation failure was unrelated, fixed in push to `main`. Ref: https://github.com/bluesky/tiled/issues/171

PR #160, opened 2022-01-28
Make API key secret larger, and store/expose last 8 for identification.

danielballan:
You only get the see the API key secret once. Later, it can be inconvenient to figure out what the key was. You can do:

```
http GET :8000/auth/apikey 'Authorization:Apikey SECRET
```

but it would be nice to be able to identify the key without _using_ the key. This PR stores the last 8 digits of the hex-encoded secret so that it can be cross-referenced easily.

Since this reduces the security of the secret, the PR adds 4 bytes to compensate.

I read that GitHub does this (or did this) for personal access tokens, but I can't find clear evidence that they _currently_ do. I don't feel confident that we should do this.

### DIFF
--- a/tiled/database/migrations/versions/481830dd6c11_initialize.py
+++ b/tiled/database/migrations/versions/481830dd6c11_initialize.py
@@ -84,6 +84,7 @@ def upgrade():
             index=True,
             nullable=False,
         ),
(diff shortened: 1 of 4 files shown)


danielballan:
Superseded by #162

PR #162, opened 2022-01-30
Rework API keys

danielballan:
- Simplify HTTP API response layouts (abandon attempting to apply JSON API for this)
- Remove uuid column from APIKey table; instead use first eight chars of hex-encoded secret to identify. (This seems to be how GitHub handles personal access tokens. See also https://www.freecodecamp.org/news/best-practices-for-building-api-keys-97c26eabfea9/.)
- Use `(first_eight, hashed_secret)` as primary key.
- Use `DELETE /apikey?first_eight={first_eight}` instead of `DELETE /apikey/{uuid}`.
- Add `tiled api_key {create, list, revoke}` CLI subcommands.
- Docs

### DIFF
--- a/docs/source/reference/scopes.md
+++ b/docs/source/reference/scopes.md
@@ -1,0 +1,23 @@
+# Scopes
+
+Tiled uses OAuth2 scopes to restrict the actions that users, services, and API keys can perform.
+
+## Scopes
+
+* `inherit` --- Default scope for API keys. Inherit scopes of the Principal associated with this key, resolved at access time.
+* `read:metadata` --- List and search metadata.
+* `read:data` --- Fetch (array, dataframe) data.
+* `apikeys` --- Manage API keys for the currently-authenticated user or service.
+* `metrics` --- Access Prometheus metrics.
+* `admin:apikeys` --- Manage API keys on behalf of any user or service.
+* `read:principals` --- Read list of all users and services and their attributes.
+
+An authenticated entity ("Principal") may be assigned roles:
+
+## Roles
+
+* `user` --- default role, granted scopes `["read:metadata", "read:data", "apikeys"]`
+* `admin` --- granted all scopes
+
+There is support for custom roles at the database level, but it is not yet
+exposed through the API.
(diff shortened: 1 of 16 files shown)


docs/source/reference/scopes.md
```diff
@@ -0,0 +1,23 @@
...
+* `read:metadata` --- List and search metadata.
+* `read:data` --- Fetch (array, dataframe) data.
+* `apikeys` --- Manage API keys for the currently-authenticated user or service.
+* `metrics` --- Access Prometheus metrics.
+* `admin:apikeys` --- Manage API keys on behalf of any user or service.
+* `read:principals` --- Read list of all users and services and their attributes.
+
+An authenticated entity ("Principal") may be assigned roles:
+
+## Roles
```
dylanmcreynolds, line 17: Probably should be a third level indent: `###`

tiled/database/orm.py
```diff
@@ -154,16 +154,15 @@ class Role(Timestamped, Base):
 class APIKey(Timestamped, Base):
     __tablename__ = "api_keys"
 
+    # Store the first_eight characters of the hex-encoded secret.
+    # The key holder can use this to identity the key.
+    # We do not store the full secret, only its sha256-hashed value.
+    # A primary key on (first_eight, hashed_secret) enables
+    # fast lookups.
+    first_eight = Column(Unicode(8), primary_key=True, index=True, nullable=False)
```
dylanmcreynolds, line 162: The full hashed value will be unique...8 bytes of the 32? Very low chance of collision, but would this be a case for an incrementing integer for that primary key?

danielballan: The primary key is on `(first_eight, hashed_secret)` so I think this is only a collision if _both_ columns collide.

Likewise with the `Identity` table where the key is `(provider, identity)`. Naturally there are many collisions on `provider` and there may be some collisions on `identity`, but that is OK.

docs/source/reference/scopes.md
danielballan: Good point. I resolved by restructuring the sections above this instead.

tiled/database/orm.py
dylanmcreynolds: Oh cool...I missed that it was a compound primary. Move along, nothing to see here.

PR #165, opened 2022-02-01
Document list of standards used.

danielballan:
During our presentation for NERSC, someone asked for more detail about our claim to "leverage standards." We gave a decent summary off the cuff, but it would be good to have a list to point to.


dylanmcreynolds (changes requested):

stuartcampbell (commented):
Do you want to mention standards like formatting and/or other coding standards that we adhere to ?  Or do you just want to list the 'technical' standards ?

danielballan:
Good question. I think that belongs in "contributing" / dev docs. The audience I have in mind for this page is someone who wants to know how "standard" the tiled service is, as opposed to the implementation/codebase.

PR #172, opened 2022-02-08
MNT: be a bit more cautious about importing numpy

tacaswell:
Small follow on to #157

A documented way to prevent a module from being imported is to set `sys.modules['foo'] = None` which would have passed the previous test and failed on import.  This will only try to use numpy if it is actually already imported.

I know this because we had a similar check for IPython in Matplotlib and get bit by someone who wanted to prevent importing IPython!


danielballan:
Always learning. Thanks @tacaswell.

PR #177, opened 2022-03-01
add missing anyio dependency

kleinhenz:
Adds missing anyio server dependency.

Also updates how the dockerfile installs dependencies to use the general requirements files instead of trying to freeze all versions. It's not clear to me that having frozen versions is worth the headache of keeping the requirements files in sync.


kleinhenz:
failing test looks unrelated. Not sure whats going on there.

danielballan:
RE: the failing test, I agree it looks unrelated. There could be a race condition with the (verrrry slow) CI filesystem. I've added sleeps in the past; we way need more and/or longer ones.

danielballan:
Interestingly, `pip install tiled[server]` _does_ install `anyio` in a clean env for me, presumably via a secondary dependency, so it's not totally clear under which conditions this fails.

But regardless, we now directly import `anyio` so we should declare that dependency explicitly.

PR #181, opened 2022-03-04
graphql

kleinhenz:
This is a concept implementation of graphql integration in tiled. Alternative to #140 using strawberry instead of ariadne. Strawberry has a dataclass approach to schema definition rather than the text based approach in ariadne. I don't have a strong preference between these approaches.

This pull request allows trees to provide a `graphql_queries` attribute which gets merged into the global `Query` object at initialization. This step is natively supported by strawberry (`strawberry.tools.merge_types`). From what I have seen this is not as easily supported in either graphene or ariadne but I could be just missing it.

As a proof of concept I implement a simple version of the `/node/search` route in graphql. This involves some code duplication of things in the rest api which have to be slightly modified to work correctly with graphql. In particular in the schema definitions all `dict` fields need to be changed to strawberry `JSON` scalars. Not sure if we even want this query but included just to get an idea of how interacting with the tree works.

Opening now to get feedback.

### DIFF
--- a/requirements-server.txt
+++ b/requirements-server.txt
@@ -24,3 +24,4 @@ toolz
 typer
 uvicorn[standard]
 watchgod
+strawberry-graphql[fastapi]
--- a/tiled/config.py
+++ b/tiled/config.py
@@ -143,6 +143,7 @@ def construct_build_app_kwargs(
             root_mapping = {}
             index = {(): root_mapping}
             include_routers = []
+            graphql_queries = []
             for segments, tree in trees.items():
                 for i in range(len(segments)):
                     if segments[:i] not in index:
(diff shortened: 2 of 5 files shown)


dylanmcreynolds:
I have never played with strawberry. I played initially with graphene and graphene-pydantic because I thought the pydantic link was cool (when you're a hammer, everything looks like a nail.) I found it to be terribly hard to use, and had some version mismatching that was just too hard to debug. So I switched to ariadne, and was able to reproduce my graphql demo in an hour and only a few lines of code (which is what #140 does.)

However, I do like the looks of this code. I'd like to play with it a little.  Thanks!

danielballan:
Agreed with @dylanmcreynolds. It looks nice. I look forward to playing with it for a bit. Your suggestion on Slack

> I'm not sure if building it in like this is the right way to go. If we have a bench of default queries that we provide then maybe it makes sense but if all of the queries are going to come from the tree then I'm not sure if this is better than just putting it in the tree and using the include_routers extension point. The case with multiple trees contributing different queries feels a bit marginal to me but I could be wrong.

is also reasonable. We might do well to play with this outside the codebase (injected via `include_routers`) and gain some experience before we commit to putting it into tiled itself.

dylanmcreynolds:
Small note...the strawberry dependency should be added to requirements-server.txt

kleinhenz:
I think it is? Does it not work for you?

danielballan:
I'm going to close this as stale. I am still interested in understanding use cases for GraphQL in Tiled.

PR #183, opened 2022-03-12
fix some auth routes

kleinhenz:
`json_or_msgpack` calls `.dict()` on content so passing a raw list does not work properly. Wrap in pydantic model first.


kleinhenz:
I actually think it makes more sense to remove the `.dict()` call in `json_or_msgpack` and make it the caller's responsibility to pass something encodable. updated to do this. Also fixes `whoami` route which previously returned json even if msgpack was requested in the public case breaking the python client assumptions.

PR #184, opened 2022-03-12
sqlite fix

kleinhenz:
* set `check_same_thread` to false for sqlite db
* follows fastapi docs and existing usage in `get_sessionmaker`
* fixes `sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread.` error when using local sqlite db

### DIFF
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -373,7 +373,11 @@ def build_app(
                 make_admin_by_identity,
             )
 
-            engine = create_engine(settings.database_uri)
+            connect_args = {}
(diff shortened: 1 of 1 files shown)


dylanmcreynolds:
Just curious, why was this found now? Something about how AIMM is packaging in Docker?

kleinhenz:
No I just found it running locally with the basic `DictionaryAuthenticator`

danielballan:
Was that a warning or an error? If an error, the auth tests should fail.

kleinhenz:
It's a warning I guess. Doesn't crash the server.

PR #185, opened 2022-03-18
Reorganize database settings. Use pessimistic disconnect handling.

danielballan:
Two changes:

First, a backward-incompatible change to the config structure for database-related config from

```yaml
database_uri: ...
database_settings:
  pool_size: ...
```

to

```yaml
database:
  uri: ...
  pool_size: ...
```

I started to do this in a backward-compatible way, but it got messy and error-prone, and I think a clean break is best. I believe the only deployments that would be affected are tiled-demo, AIMM, and NSLS2.

Second, the use of [pessimistic disconnection handling](https://docs.sqlalchemy.org/en/14/core/pooling.html#disconnect-handling-pessimistic) by default. It can be turned off via a new config setting:

```yaml
database:
  uri: ...
  pool_size: ...
  pool_pre_ping: False
```

This solves a critical issue that we observed in the NSLS2 deployment where long-running servers would end up with broken database connections in their pool. You have a small performance hit for this, but I think it's necessary. As another data point, JupyterHub uses this setting as well.

### DIFF
--- a/docs/source/how-to/tiled-database.md
+++ b/docs/source/how-to/tiled-database.md
@@ -34,7 +34,8 @@ in place of `...` below. Inject the password via an environment variable as show
 do not hard-code it in the configuration file. Be sure to quote the URI.
 
 ```yaml
-database_uri: "postgresql://tiled:${TILED_DATABASE_PASSWORD}@.../tiled"
+database
+  uri: "postgresql://tiled:${TILED_DATABASE_PASSWORD}@.../tiled"
 ```
 
 Initialize the database. Initialization only has to be done once ever. (If you
--- a/docs/source/how-to/tiled-database.md
+++ b/docs/source/how-to/tiled-database.md
@@ -49,6 +49,5 @@ The database is ready to use.
 
 ## Reference
 
-See `database` in the service configuration reference
-{doc}`../reference/service` for a comprehensive list of the options for tuning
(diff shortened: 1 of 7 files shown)


dylanmcreynolds:
I'm confused. I only see changes in documentation and test suite configuration. This doesn't require a change in tiled code?

docs/source/how-to/tiled-database.md
```diff
@@ -34,7 +34,8 @@ in place of `...` below. Inject the password via an environment variable as show
 do not hard-code it in the configuration file. Be sure to quote the URI.
 
 ```yaml
-database_uri: "postgresql://tiled:${TILED_DATABASE_PASSWORD}@.../tiled"
+database
+  uri: "postgresql://tiled:${TILED_DATABASE_PASSWORD}@.../tiled"
```
dylanmcreynolds, line 38: Should this describe pool size and pre pool ping?

danielballan: Good note.

I added a section to the bottom of this page that links to the service configuration reference, which already includes descriptions of the various options.

dylanmcreynolds (approved):

kleinhenz (approved):

danielballan:
Power-cycled to restart GH Actions after today's GH Actions outage.

PR #186, opened 2022-03-18
Explain what the identifying_file is up to.

danielballan:
This provides more documentation on our approach for locating `share/tiled` on the system.

### DIFF
--- a/share/tiled/.identifying_file_72628d5f953b4229b58c9f1f8f6a9a09
+++ b/share/tiled/.identifying_file_72628d5f953b4229b58c9f1f8f6a9a09
@@ -1,0 +1,2 @@
+This is file is used as a sentinel by tiled.utils.get_share_tiled_path to
+confirm that it has found shared/tiled on the system.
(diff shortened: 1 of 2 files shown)


kleinhenz:
Do you understand why these shouldn't just go in the package directory under e.g. `resources` or something like that?

danielballan:
From [packaging docs](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#data-files):

> In some cases you may need to place data files outside of your [packages](https://packaging.python.org/en/latest/glossary/#term-Import-Package). The `data_files` directive allows you to do that. It is mostly useful if you need to install files which are used by other programs, which may be unaware of Python packages.

I believe that `shared/X` is the standard place to put static assets that may be supplemented or altered for specific deployments. It's much safer and easier for a customized deployment or package to modify `share/tiled` than to reach inside the Python package itself. These static files are distinct from other static files in the package (like, say, our configuration schemas) that are more tightly coupled to the Python code in the package.

kleinhenz (approved):

PR #189, opened 2022-03-25
Require msgpack >=1.0.0.

danielballan:
Prompted by a user bug report where tiled was installed into an
existing env that had a too-old version (0.6.1) missing timestamp
support.


danielballan:
Power-cycled because GH Actions didn't run the first time.

tacaswell (approved):

PR #192, opened 2022-03-29
Support lookup like c['a', 'b'].

danielballan:
With this PR, these are equivalent:

```py
>>> node['a']['b']['c']
>>> node[('a', 'b', 'c')]
>>> node['a', 'b', 'c']
```

The last two, it should be noted, are equivalent on a Python level, nothing to do with tiled:

```py
In [1]: class A:
   ...:     def __getitem__(self, stuff):
   ...:         print(repr(stuff))
   ...: 

In [2]: a = A()

In [3]: a[1]
1

In [4]: a[1, 2]
(1, 2)

In [5]: a[(1, 2)]
(1, 2)
```

As noted  in the code comments, this could be made faster by eliding `c['a', 'b', 'c']` into a single request rather than issuing chained requests. But there are some tricky aspects to that (described in more detail in the comments) so we'll leave it for another time.

### DIFF
--- a/requirements-client.txt
+++ b/requirements-client.txt
@@ -1,5 +1,6 @@
 # These are the requirements needed for basic client functionality.
 appdirs
+click !=8.1.0
 entrypoints
 heapdict
 httpx >=0.20.0
--- a/requirements-server.txt
+++ b/requirements-server.txt
@@ -5,6 +5,7 @@ anyio
 cachey
 cached-property; python_version < '3.8'
 cachetools
+click !=8.1.0
 dask
 fastapi
 jinja2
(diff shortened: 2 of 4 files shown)


danielballan:
Added an unrelated commit to address an issue with a recent release of `click`: https://github.com/tiangolo/typer/issues/377

danielballan:
I'll add there is a clean analogy to numpy slicing here.

```py
>>> arr[0][1] 
>>> arr[0, 1]  # equivalent
```

tiangolo:
I just released Typer `0.4.1` that should handle the issue. :rocket: :nerd_face:

PR #202, opened 2022-04-20
Create server client shared structure_family class and pydantic server-only structures

jmaruland:
This PR will try to solve inconsistencies that are raised by FastAPI and pydantic when using strucure classes that are dependent of the `dataclasses.dataclass` package while trying to solve the new write feature on the server side

### DIFF
--- a/tiled/adapters/hdf5.py
+++ b/tiled/adapters/hdf5.py
@@ -1,4 +1,5 @@
 import collections.abc
+import os
 import warnings
 
 import dask.array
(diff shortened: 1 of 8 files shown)


danielballan:
I hope we can find a solution to the underlying issue in pydantic, and perhaps submit a patch upstream to them, but this is a good workaround to unblock us in the meantime.

danielballan:
This is a good description of the underlying pydantic issue we hit with `ArrayStrucutre`. https://github.com/samuelcolvin/pydantic/issues/3251

I think the issue with `DataFrameStructure` is unrelated and could be addressed without changes to pydantic.

danielballan:
This is a good description of the underlying pydantic issue we hit with `ArrayStrucutre`. https://github.com/samuelcolvin/pydantic/issues/3251

I think the issue with `DataFrameStructure` is unrelated and could be addressed without changes to pydantic.

danielballan:
This is a good description of the underlying pydantic issue we hit with `ArrayStrucutre`. https://github.com/samuelcolvin/pydantic/issues/3251

I think the issue with `DataFrameStructure` is unrelated and could be addressed without changes to pydantic.

PR #203, opened 2022-05-04
Fix locket usage.

danielballan:
I'm seeing CI failures on the daily scheduled runs, likely due to the recent release of locket 1.0.0: https://pypi.org/project/locket/#history


danielballan:
The CI is still failing because `tifffile` also recently released backward-incompatible changes, but this PR is a step in the right direction.

PR #204, opened 2022-05-04
Add [experimental] methods for writing arrays and dataframes

jmaruland:
NOTE:
The methods that were added to node.py and context.py are intended to be experimental. We are testing approaches to make tiled receive data and put it in a node in the tree. This might be subject to changes in the future. The idea behind this approach is to make it as easy as possible to the user to use a write command  for tiled while the data validation, data storage and management happens internally and with no complicated or convoluted steps for the user.

### DIFF
--- a/tiled/client/node.py
+++ b/tiled/client/node.py
@@ -550,9 +550,17 @@ class Node(BaseClient, collections.abc.Mapping, IndexersMixin):
 
     def write_array(self, array, metadata=None, specs=None):
         """
-        EXPERIMENTAL: write an array
+        EXPERIMENTAL: Write an array.
 
-        This is subject to change or removal without notice
+        Parameters
+        ----------
+        array : array-like
+        metadata : dict, optional
+            User metadata. May be nested. Must contain only basic types
+            (e.g. numbers, strings, lists, dicts) that are JSON-serializable.
+        specs : List[str], optional
+            List of names that are used to label that the data and/or metadata
+            conform to some named standard specification.
 
         """
 
@@ -593,15 +601,27 @@ class Node(BaseClient, collections.abc.Mapping, IndexersMixin):
             + "/"
             + uid
         )
-        self.context.put_content(full_path_data, content=array.tobytes())
+        self.context.put_content(
+            full_path_data,
+            content=array.tobytes(),
+            headers={"Content-Type": "application/octet-stream"},
+        )
 
     def write_dataframe(self, dataframe, metadata=None, specs=None):
         """
-        EXPERIMENTAL: write a dataframe
+        EXPERIMENTAL: Write a DataFrame.
 
         This is subject to change or removal without notice
 
-
+        Parameters
+        ----------
+        dataframe : pandas.DataFrame
+        metadata : dict, optional
+            User metadata. May be nested. Must contain only basic types
+            (e.g. numbers, strings, lists, dicts) that are JSON-serializable.
+        specs : List[str], optional
+            List of names that are used to label that the data and/or metadata
+            conform to some named standard specification.
         """
         from dask.dataframe.utils import make_meta
 
--- a/tiled/client/node.py
+++ b/tiled/client/node.py
@@ -6,7 +6,6 @@ import itertools
 import time
 import warnings
 from dataclasses import asdict, fields
-from io import BytesIO
 
 import entrypoints
 
@@ -549,7 +548,7 @@ class Node(BaseClient, collections.abc.Mapping, IndexersMixin):
             # Do not print messy traceback from thread. Just fail silently.
             return []
 
-    def create_array(self, array, metadata=None, specs=None):
+    def write_array(self, array, metadata=None, specs=None):
         """
         EXPERIMENTAL: write an array
 
@@ -582,7 +581,7 @@ class Node(BaseClient, collections.abc.Mapping, IndexersMixin):
         uid = document["uid"]
         self.context.put_content(f"/array/full/{uid}", content=array.tobytes())
 
-    def create_dataframe(self, dataframe, metadata=None, specs=None):
+    def write_dataframe(self, dataframe, metadata=None, specs=None):
         """
         EXPERIMENTAL: write a dataframe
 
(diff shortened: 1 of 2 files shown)


tiled/client/node.py
```diff
@@ -543,6 +549,87 @@ def _ipython_key_completions_(self):
             # Do not print messy traceback from thread. Just fail silently.
             return []
 
+    def create_array(self, array, metadata=None, specs=None):
```
danielballan, line 552: On second thought, I think `write_array` (matching `read()`) would be a better name here. From the user's point of view, they aren't _creating_ the array (they already have the array) they are writing it into tiled's storage.

tiled/client/node.py
```diff
@@ -543,6 +549,87 @@ def _ipython_key_completions_(self):
...
+            "structure": asdict(structure),
+            "structure_family": StructureFamily.array,
+            "specs": specs,
+            "mimetype": mimetype,
+        }
+        document = self.context.post_json("/node/metadata/", data)
+        uid = document["uid"]
+        self.context.put_content(f"/array/full/{uid}", content=array.tobytes())
+
+    def create_dataframe(self, dataframe, metadata=None, specs=None):
```
danielballan, line 585: Same here: `write_dataframe`.

tiled/client/node.py
```diff
@@ -543,6 +549,87 @@ def _ipython_key_completions_(self):
...
+            micro=BuiltinDtype.from_numpy_dtype(array.dtype),
+        )
+        data = {
+            "metadata": metadata,
+            "structure": asdict(structure),
+            "structure_family": StructureFamily.array,
+            "specs": specs,
+            "mimetype": mimetype,
+        }
+        document = self.context.post_json("/node/metadata/", data)
```
danielballan, line 581: If we write to a node like `/analyzed` this should go to `/node/metadata/analyzed/{uid}`. This example is relevant:

https://github.com/bluesky/tiled/blob/16478d9e55b6f1f401b98d9fe042e3d0ed4d7f80/tiled/client/array.py#L49-L53

tiled/client/node.py
```diff
@@ -543,6 +549,87 @@ def _ipython_key_completions_(self):
...
+        uid = document["uid"]
+        self.context.put_content(f"/array/full/{uid}", content=array.tobytes())
```
danielballan, line 583: Same issue here about including the path to this node.

tiled/client/node.py
```diff
@@ -543,6 +549,87 @@ def _ipython_key_completions_(self):
...
+            "structure_family": StructureFamily.dataframe,
+            "specs": specs,
+            "mimetype": mimetype,
+        }
+
+        data["structure"]["micro"]["meta"] = base64.b64encode(
+            bytes(serialize_arrow(data["structure"]["micro"]["meta"], {}))
+        ).decode()
+
+        document = self.context.post_json("/node/metadata/", data)
```
danielballan, line 624: Same issue here about including the path to this node.

tiled/client/node.py
```diff
@@ -543,6 +549,87 @@ def _ipython_key_completions_(self):
...
+        uid = document["uid"]
+
+        write_buffer = BytesIO()
+        dataframe.to_csv(write_buffer)
+        self.context.put_content(
```
danielballan, line 629: Same issue here about including the path to this node.

tiled/client/node.py
```diff
@@ -543,6 +549,87 @@ def _ipython_key_completions_(self):
...
+
+        data["structure"]["micro"]["meta"] = base64.b64encode(
+            bytes(serialize_arrow(data["structure"]["micro"]["meta"], {}))
+        ).decode()
+
+        document = self.context.post_json("/node/metadata/", data)
+        uid = document["uid"]
+
+        write_buffer = BytesIO()
+        dataframe.to_csv(write_buffer)
```
danielballan, line 628: Does this need to be Arrow, not CSV? The `serialize_arrow` function would be useful here.

danielballan (changes requested):

tiled/adapters/tiff.py
```diff
@@ -216,8 +216,9 @@ def _safe_asarray(seq, index=None):
     https://github.com/cgohlke/tifffile/commit/18315846ebf0b31126dfdeffee8e0ef30d81b31c#diff-d8d5cccae6533b861bdf6c0c20c6f49415978effec9a13797909535f3fc7390dL9563-R9746  # noqa
     """
 
-    try:
-        return seq.asarray(index=index)
-    except TypeError:
-        # usage for versions before 2021.10.10
-        return seq.asarray(file=index)
+    return seq.asarray()[index]
```
danielballan, line 219: This reads the entire sequence (i.e. all the TIFF files) and then slices out the part we want. The existing code reads only exactly the data we need, which is of course faster. I am sure that tifffile still supports this somehow, but it's not immediately clear how. Let's address this tifffile change in a separate PR.

tiled/client/node.py
```diff
@@ -624,11 +623,12 @@ def create_dataframe(self, dataframe, metadata=None, specs=None):
...
         uid = document["uid"]
 
-        write_buffer = BytesIO()
-        dataframe.to_csv(write_buffer)
-        self.context.put_content(
-            f"/dataframe/full/{uid}", content=write_buffer.getvalue()
-        )
+        # write_buffer = BytesIO()
+        # dataframe.to_csv(write_buffer)
+        serilized_data = base64.b64encode(
```
danielballan, line 628: Why base64 encode?

danielballan (changes requested):

tiled/adapters/tiff.py
jmaruland: Fixed in https://github.com/bluesky/tiled/pull/206

danielballan (approved):

tiled/client/node.py
```diff
@@ -543,6 +548,111 @@ def _ipython_key_completions_(self):
...
+        uid = document["uid"]
+
+        full_path_data = (
+            "/array/full"
+            + "".join(f"/{part}" for part in self.context.path_parts)
+            + "".join(f"/{part}" for part in self._path)
+            + "/"
+            + uid
+        )
+        self.context.put_content(full_path_data, content=array.tobytes())
```
kleinhenz, line 596: Should we send `npy` instead? The metadata would be redundant with the micro/macro structure but that is also true for arrow.

kleinhenz: Should there be a general mechanism for providing a serializer to turn your object into bytes? For reading we allow you to request data in many different formats. Maybe the same should be true for writing?

danielballan (changes requested):
Taking a closer look at this I have more questions before we merge. Will comment in detail later.

tiled/client/node.py
danielballan: As a wire format, I think the server should accept just the array buffer because that is a more cross-language concept, where the `npy` format (with some dtype header in there) is a Python-specific thing.

I think it would be sensible for the server to _persist_ the data as npy though.

danielballan (approved):
We resolved our concerns over a phone call. Will merge once @jmaruland confirms that this can integrate with his experimental databroker server extension.

PR #205, opened 2022-05-05
provide tighter control over number of database connections

kleinhenz:
This provides tighter control over the number of db connections established by tiled.

In particular, it makes the sqlalchemy max_overflow parameter configurable and avoids creating a separate engine during app startup. This limits the number of established connections to the value expected from the configuration.


danielballan:
Power-cycled to re-run CI now that the `main` branch is passing again.

PR #206, opened 2022-05-06
Adjust to backward incompatible tifffile changes

jmaruland:
Recent update of tifffile raised errors for our previous implementation of asarray() on TiffSequence objects. Fixed.


danielballan:
This is great. Thank you for handling this @jmaruland.

PR #207, opened 2022-05-10
Ldap authenticator

dmgav:
Implementation of LDAP authenticator. The authenticator could be configured in .yml file as

```
authentication:
    providers:
        - provider: ldap_local 
          authenticator: tiled.authenticators:LDAPAuthenticator
          args:
              server_address: localhost
              server_port: 1389
              bind_dn_template: "cn={username},ou=users,dc=example,dc=org"
              use_tls: false
              use_ssl: false
    tiled_admins:
        - provider: ldap_local 
          id: user02
```

### DIFF
--- a/tiled/authenticators.py
+++ b/tiled/authenticators.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import re
 import secrets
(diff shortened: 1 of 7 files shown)


tiled/authenticators.py
```diff
@@ -1,8 +1,11 @@
 import logging
```
dylanmcreynolds, line 1: Since not every tiled instance will integrate with LDAP, would you consider moving these imports so that they are only done if an ldap authenticator is needed?

danielballan: It took a few tries there, but this is now fixed.

dylanmcreynolds (commented):
Let preface this comment by acknowledging this is the very first test_authenticators.py commit, so thanks for adding it! 

Possibly for a later PR, but have you thought of mocking the ldap3 code? I have no experience mocking ldap, but there even seems to be some support in ldap3 https://ldap3.readthedocs.io/en/latest/mocking.html

danielballan:
* I am grateful that this has unit tests, period. (More than I did with PAM, OIDC, SAML!)
* I agree with @dylanmcreynolds that this would be a good use for mocking.
* Given internal NSLS-II time pressures I think I'm in favor of refactoring to mocking in a separate PR.

PR #212, opened 2022-05-18
invalidate length cache after calling write methods

kleinhenz:
The write_* methods of node are expected to change the number of entries so the length cache should be invalidated.


danielballan:
Good catch!

Implementation suggestion: I think a faster and simpler way to invalidate the cache would be to set `_cached_len` to `None`. It saves a syscall to the clock and is a one-liner so it doesn’t require a new method.

PR #216, opened 2022-05-25
Improve usability of slices into Nodes' keys, values, and items.

### LINKED ISSUE #215: Better names for the *_indexer accessors
Tiled client `Node` object satisfy the Python [`Mapping` interface](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping); they are "dict-like". This is nice because all the things you can do with dicts

```py
d["my_data"]
len(d)
d.items()
# etc.
```

work on `Node` just the same.

The `Mapping` interface supports look up by _name_ like `d["my_data"]` but not look up by _position_ like `d[0]`. For `Node`, we want position-based access too; we want to support the Python [`Sequence` interface](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence).

What are our options for doing this?

## Accessors with log names (current approach)

A `Node` _is_ a `Mapping` and it _has_ accessors that satisfy the `Sequence` interface.

```py
c.values_indexer[...]
c.keys_indexer[...]
c.items_indexer[...]
```

Pro:

* Explicit switching between named-based lookup and positional lookup
* Unambiguous

Con:

* Way too verbose to just get the first/last item

## Single Dispatch

```py
c[0]  # int -> Sequence-like access
c[:5]  # slice -> Sequence-like access
c["image"]  # str -> Mapping-like access
```

Pro:

* No new names to learn as the user

Cons:

* This kind of overloading can be confusing.
* Databroker has a _strong_ commitment to interpreting integer keys as `scan_id`. Breaking that would be a huge problem for Databroker. For this reason, this is a nonstarter.

## Overload `.values`, `.keys`, `.items`

Continue to support `.values()`, `.keys()`, `.items()`. _Also_ support `.values[]`, `.keys[]`, `.items[]`.

Pro:

* No new names to learn

Cons:

* Overloading interface methods like this with `__getitem__` is _very_ strange. (I can't think of a single time I've seen this done.)

## Give `.values_indexer` (the most common one) a shorter alias

Ideas for ideas include:

```
.s  # not obvious, but nice and short! meant to invoke "slice" or "sequence"
.slice
.seq  # for "sequence..."
.iloc  # familiar if you know pandas, but inscrutable otherwise
.val  # good name, but not obvious how to use it, or which is which between "val" and "values"
```

I really want a _very short_ way to type "Get me the first one," so I currently lean toward `.s`.

## Make `.values()`, `.keys()`, `.items()` return iterators that are indexable.

```py
c.values()[0]  # first
c.values()[:5]  # first 5
# ditto for keys, items
```

h/t @kleinhenz for this suggestion, which is very much in the spirit of [slicerator](https://github.com/soft-matter/slicerator) and [PIMS](https://github.com/soft-matter/pims). I can't believe I didn't think of it!

danielballan:
Closes #215. See that issue for justification and discussion.

### DIFF
--- a/tiled/iterviews.py
+++ b/tiled/iterviews.py
@@ -1,0 +1,205 @@
+"""
+Iterables for KeysView, ValuesView, ItemsView that are sliceable.
+"""
+
+
+class IterViewBase:
+
+    __slots__ = ("_get_length",)
+
+    def __init__(self, get_length):
+        self._get_length = get_length
+
+    def __repr__(self):
+        return f"<{type(self).__name__}>"
+
+    # Convenience aliases
+
+    def first(self):
+        return self[0]
+
+    def last(self):
+        return self[-1]
+
+    def head(self, n=5):
+        return self[:n]
+
+    def tail(self, n=5):
+        return list(reversed(self[-1 : -(n + 1) : -1]))  # noqa: E203
+
+    def __init_subclass__(cls, *args, **kwargs):
+        cls.first.__doc__ = f"Get the first {cls._name}."
+        cls.last.__doc__ = f"Get the last {cls._name}."
+        cls.head.__doc__ = f"Get the first N {cls._name}s."
+        cls.tail.__doc__ = f"Get the last N {cls._name}s."
+
+    def __len__(self):
+        return self._get_length()
+
+
+class KeysView(IterViewBase):
+    """
+    A sliceable, iterable view of keys.
+    """
+
+    __slots__ = ("_keys_slice",)
+    _name = "key"
+
+    def __init__(self, get_length, keys_slice):
+        self._keys_slice = keys_slice
+        super().__init__(get_length)
+
+    def __getitem__(self, index_or_slice):
+        if isinstance(index_or_slice, int):
+            if index_or_slice < 0:
+                index_or_slice = -1 - index_or_slice
+                direction = -1
+            else:
+                direction = 1
+            keys = list(self._keys_slice(index_or_slice, 1 + index_or_slice, direction))
+            try:
+                (key,) = keys
+            except ValueError:
+                raise IndexError("Index out of range")
+            return key
+        elif isinstance(index_or_slice, slice):
+            start, stop, direction = slice_to_interval(index_or_slice)
+            return list(self._keys_slice(start, stop, direction))
+        else:
+            raise TypeError(
+                f"{index_or_slice} must be an int or slice, not {type(index_or_slice)}"
+            )
+
+    def __iter__(self):
+        yield from self._keys_slice(0, None, 1)
+
+
+class ItemsView(IterViewBase):
+    """
+    A sliceable, iterable view of (key, value) pairs.
+    """
+
+    __slots__ = ("_items_slice",)
+    _name = "item"
+
+    def __init__(self, get_length, items_slice):
+        self._items_slice = items_slice
+        super().__init__(get_length)
+
+    def __getitem__(self, index_or_slice):
+        if isinstance(index_or_slice, int):
+            if index_or_slice < 0:
+                index_or_slice = -1 - index_or_slice
+                direction = -1
+            else:
+                direction = 1
+            items = list(
+                self._items_slice(index_or_slice, 1 + index_or_slice, direction)
+            )
+            try:
+                (item,) = items
+            except ValueError:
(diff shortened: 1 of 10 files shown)


danielballan:
Example usage:

```py

In [2]: c = from_uri("https://tiled-demo.blueskyproject.io/api")

In [3]: bmm = c['bmm']['raw']

In [6]: bmm.values().first()
Out[6]: <BlueskyRun {'primary', 'baseline'} scan_id=22521 uid='f8c83910' 2020-03-02 23:18>

In [7]: bmm.values().head()
Out[7]: 
[<BlueskyRun {'primary', 'baseline'} scan_id=22521 uid='f8c83910' 2020-03-02 23:18>,
 <BlueskyRun {'primary', 'baseline'} scan_id=22524 uid='51f9eb19' 2020-03-02 23:33>,
 <BlueskyRun {'primary', 'baseline'} scan_id=22525 uid='d96c0436' 2020-03-02 23:45>,
 <BlueskyRun {'primary', 'baseline'} scan_id=22526 uid='a7d56444' 2020-03-02 23:57>,
 <BlueskyRun {'primary', 'baseline'} scan_id=22528 uid='599b29f3' 2020-03-03 00:21>]

In [8]: bmm.values()[3]
Out[8]: <BlueskyRun {'primary', 'baseline'} scan_id=22526 uid='a7d56444' 2020-03-02 23:57>

In [9]: bmm.values()[:3]
Out[9]: 
[<BlueskyRun {'primary', 'baseline'} scan_id=22521 uid='f8c83910' 2020-03-02 23:18>,
 <BlueskyRun {'primary', 'baseline'} scan_id=22524 uid='51f9eb19' 2020-03-02 23:33>,
 <BlueskyRun {'primary', 'baseline'} scan_id=22525 uid='d96c0436' 2020-03-02 23:45>]

In [10]: c['um2022']['olds']['ethanol_fast'].keys().first()
Out[10]: 'rand_Ethanol_short_20210913-215007_fa938e_0001_mean_tth'

In [11]: c['um2022']['olds']['ethanol_fast'].keys().head()
Out[11]: 
['rand_Ethanol_short_20210913-215007_fa938e_0001_mean_tth',
 'rand_Ethanol_short_20210913-215009_16c12d_0001_mean_tth',
 'rand_Ethanol_short_20210913-215011_d2cc70_0001_mean_tth',
 'rand_Ethanol_short_20210913-215014_746e44_0001_mean_tth',
 'rand_Ethanol_short_20210913-215018_09f455_0001_mean_tth']
```

tiled/iterviews.py
```diff
@@ -0,0 +1,205 @@
...
+        return self[0]
+
+    def last(self):
+        return self[-1]
+
+    def head(self, n=5):
+        return self[:n]
+
+    def tail(self, n=5):
+        return list(reversed(self[-1 : -(n + 1) : -1]))  # noqa: E203
```
jmaruland, line 28: I like the idea of using methods like `first()`, `last()`, `head()` and `tail()`. It might be silly but I think we should check for cases smaller than the default value. After dealing with some cases where people saved datasets with no data or just one sample, I always try and check for size before indexing. For example, what if the dataset is len(self)<5 and the method uses the default value for n (n=5)?

kleinhenz (approved):

kleinhenz:
This looks good to me. I tried it out in aimmdb and didn't run into any problems.

tiled/iterviews.py
danielballan: Good question. We are protected there by Python's slicing behavior. If you slice "the first 100" in a list of length 3, you get 3 items back, not an error. The slicing is interpreted to mean _up to_ the first 100. This usually the desired behavior.

```py
>>> ['a', 'b', 'c'][:100]
['a', 'b', 'c']
```

On the other hand if you ask for the 100th item specifically, not a slice by an integer lookup, then you get an `IndexError`.

```py
>>> ['a', 'b', 'c'][100]
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
```

I added some tests just now ([91b3f9f](https://github.com/bluesky/tiled/pull/216/commits/91b3f9fcbd24ed1625d7a5109527b19923d3c707)) to verify that calling `.head()` on an empty tree returns an empty list, whereas calling `.first()` raises an `IndexError`.

jmaruland (approved):

PR #218, opened 2022-05-26
Add binary operator queries.

danielballan:
## New queries

This adds queries for equality, binary comparison (>, <, <=, >=), and collection membership.

By extending the vocabulary of queries built into Tiled, this PR:

* Provides a larger set of queries that can, potentially work across different Tiled servers
* Make it easier to construct simple queries without learning Mongo query syntax
* Lays track for removing support for direct Mongo query syntax (See https://github.com/bluesky/databroker/issues/701.)

## New query encoding mechanism

To support `GET` requests, we encode search queries in URL query parameters.

In the main branch, the logic for encoding and decoding queries is hard-coded inside the client and server respectively. In this PR, we move that logic to instance and class methods (respectively) on the query objects themselves. This put the query objects in charge of encoding themselves into URL query parameters, so that we don't need to cover every possibility in the client and server code.

This is a backward-incompatible change. Custom queries defined in external libraries must add `encode` and `decode` methods.

## Usage of the new queries

The new queries have a straightforward usage. This looks similar to how `FullText`, `TimeRange`, and other existing queries look

```py
from tiled.queries import Contains, Comparison, Eq

c.search(Eq("color", "red"))  # i.e. color == "red"
c.search(Eq("sample.name", "Ni"))  # nested key sample.name == "Ni"
c.search(Comparison("gt", "temperature", 100))  # i.e temperature > 100
c.search(Contains("detectors", "ccd"))  # i.e. "ccd" in detectors
```

## Experimental high-level usage

The experimental `Key` object is not a query, but it can be used to build queries in expressions.

```py
from tiled.queries import Key

c.search(Key("color") == "red"))  # i.e. color == "red"
c.search(Key("sample.name") == "Ni"))  # nested key sample.name == "Ni"
c.search(Key("temperature") > 100))  # i.e temperature > 100
```

What's happening here is, `Key` defined magic methods `__eq__`, `__gt__`, (etc.) such that the expression `Key("color") == "red")` returns `Eq("color", "red")` and the expression `Key("temperature") > 100)` returns `Comparison("gt", "temperature", 100)`. It's just a cute shorthand.

This may be too magical or odd. Normally expressions with a binary operator are expected to return a boolean, not a custom object. I think it is significant that the language allows this; it could call `bool` on the result to coerce it be either `True` or `False` but it does not. (In other cases, like `__contains__`, it does.)

I know of at least one precedent: [SQL Alchemy ORMs](https://docs.sqlalchemy.org/en/14/orm/queryguide.html).

Also, one or two users at databroker tutorials over the years have asked why they can't use normal Python operators in searches. There are at least some users out there who would find this natural.

I propose to experiment with this and rip it out if it causing more confusion than help.

### DIFF
--- a/tiled/adapters/mapping.py
+++ b/tiled/adapters/mapping.py
@@ -5,7 +5,7 @@ import operator
 from datetime import datetime
 
 from ..iterviews import ItemsView, KeysView, ValuesView
-from ..queries import Comparison, Contains, Eq, FullText
+from ..queries import Comparison, Contains, Eq, FullText, Regex
 from ..query_registration import QueryTranslationRegistry
 from ..utils import UNCHANGED, DictView, SpecialUsers, import_object
 from .utils import IndexersMixin
@@ -312,6 +312,28 @@ def full_text_search(query, tree):
 
 
 MapAdapter.register_query(FullText, full_text_search)
+
+
+def regex(query, tree):
+    import re
+
+    matches = {}
+    flags = 0 if query.case_sensitive else re.IGNORECASE
+    pattern = re.compile(query.pattern, flags=flags)
+    for key, value in tree.items():
+        # Find the queried key in the metadata.
+        term = value.metadata
+        for subkey in query.key.split("."):
+            if subkey not in term:
+                break
+            term = term[subkey]
+        else:
+            if isinstance(term, str) and pattern.search(term):
+                matches[key] = value
+    return tree.new_variation(mapping=matches)
+
+
+MapAdapter.register_query(Regex, regex)
 
 
 def eq(query, tree):
--- a/tiled/_tests/test_queries.py
+++ b/tiled/_tests/test_queries.py
@@ -5,7 +5,7 @@ import numpy
 from ..adapters.array import ArrayAdapter
 from ..adapters.mapping import MapAdapter
 from ..client import from_tree
-from ..queries import Comparison, Contains, Eq, Key
+from ..queries import Comparison, Contains, Eq, FullText, Key, Regex
 
 keys = list(string.ascii_lowercase)
 mapping = {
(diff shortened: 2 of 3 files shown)


danielballan:
I received very positive feedback on this plan from five power users when I posted in the NSLS2 Slack. A good signal that this is worth trying.

tiled/adapters/mapping.py
```diff
@@ -313,6 +314,67 @@ def maybe_lower(s):
...
+        # Find the queried key in the metadata.
+        term = value.metadata
+        for subkey in query.key.split("."):
+            if subkey not in term:
+                break
+            term = term[subkey]
+        else:
+            if term == query.value:
+                matches[key] = value
+    return tree.new_variation(mapping=matches)
```
jmaruland, line 329: I feel that it would be a more "Pythonesque" approach if the method returns `None` when there is no match

tiled/adapters/mapping.py
```diff
@@ -313,6 +314,67 @@ def maybe_lower(s):
...
+                break
+            term = term[subkey]
+        else:
+            if (
+                isinstance(term, collections.abc.Iterable)
+                and (not isinstance(term, str))
+                and (query.value in term)
+            ):
+                matches[key] = value
+    return tree.new_variation(mapping=matches)
```
jmaruland, line 351: Same with the match result of this method

tiled/adapters/mapping.py
```diff
@@ -313,6 +314,67 @@ def maybe_lower(s):
...
+            if subkey not in term:
+                break
+            term = term[subkey]
+        else:
+            if query.operator not in {"le", "lt", "ge", "gt"}:
+                raise ValueError(f"Unexpected operator {query.operator}.")
+            comparison_func = getattr(operator, query.operator)
+            if comparison_func(term, query.value):
+                matches[key] = value
+    return tree.new_variation(mapping=matches)
```
jmaruland, line 372: Same with the match result of this method

tiled/adapters/mapping.py
danielballan: I think it's better to retain type stability. The code calling this expects a `tree`, an object with certain methods. If we _sometimes_ return `None` every piece of code that handles this will have to check if it's `None`, and if we forget somewhere, that will be an error. An empty `tree` is a valid thing to have, and it behaves sensibly. It has `len(tree) == 0` and so on.

tacaswell:
Mostly a conceptual review (but did skim the code).

I think (ab)using the binary operators like this is a bit funny, but is justified from a usability point of view.

I would be interested if we could find any reasoning from upstream about why the comparisons behave this way and `__contains__` forces to bool.

PR #220, opened 2022-05-27
moved write paths to tiled server

jmaruland:


tiled/server/router.py
```diff
@@ -531,3 +534,60 @@ def node_full(
...
+    body: schemas.PostMetadataRequest,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
+):
+    if body.structure_family == StructureFamily.dataframe:
+        # Decode meta for pydantic validation
+        body.structure.micro.meta = deserialize_arrow(
+            base64.b64decode(body.structure.micro.meta)
+        )
+
+    try:
```
danielballan, line 551: I think this is too broad. If a node _tries_ to support writing but then encounters some error on the way, we'll tell the user "This node is not writable," when the actual problem is that writing should work but failed for some reason.

Perhaps

```py
if hasattr(entry, "post_metadata"):
    ...
else:
    raise ...
```

instead.

tiled/server/schemas.py
```diff
@@ -8,6 +8,8 @@
 import pydantic.generics
 
 from ..structures.core import StructureFamily
+from .pydantic_array import ArrayStructure
+from .pydantic_dataframe import DataFrameStructure
```
danielballan, line 12: This can be changed to import from `tiled.structures.dataframe` once #229 is merged.

tiled/server/router.py
```diff
@@ -531,3 +534,60 @@ def node_full(
...
+
+
+@router.put("/array/full/{path:path}")
+async def put_array_full(
+    request: Request,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
+):
+    data = await request.body()
+
+    try:
```
danielballan, line 571: Ditto

tiled/server/router.py
```diff
@@ -531,3 +534,60 @@ def node_full(
...
+
+
+@router.put("/node/full/{path:path}")
+async def put_dataframe_full(
+    request: Request,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
+):
+    data = await request.body()
+
+    try:
```
danielballan, line 587: Ditto

danielballan (changes requested):

tiled/server/router.py
kleinhenz: What is preferred between checking `hasattr` and catching `AttributeError`?

danielballan: "Look before you leap" (`hasattr`) is preferred unless the case where it's missing is rare/exceptional. Catching exceptions is comparatively expensive and generally not the right way to do interface checking.

Another option, which has grown in popularity is to make a Protocol and then do `isinstance(entry, Writable)`. I don't think we need to get into that yet, but maybe later.

danielballan (approved):

PR #225, opened 2022-06-03
Enforce a minimum pool_size in schema.

danielballan:
I tried a `pool_size` of 1 and send my workers into a confusing crash loop (with no useful output). Enforcing this at the schema level will save others from the same confusion.


kleinhenz (approved):

danielballan:
CI failure is a test race condition, unrelated to the changes.

PR #227, opened 2022-06-03
Improve documentation on queries

danielballan:

PR #231, opened 2022-06-08
Raise clear error with 'and' or 'or'.

danielballan:
As noted in #227 queries cannot be made to work with the Python keywords `and` and `or`. Currently, they just silently ignore all but for the first or last query. We documented that, but it still seems likely to cause user confusion, silently giving them an answer to a different question than the one they meant to ask.

There is no hook for `and` or `or` specifically, but we can hook `bool` to ensure that attempts to use `and` or `or` raise an error:

```py
In [1]: from tiled.queries import Key

In [2]: Key("color") == "red"
Out[2]: Eq(key='color', value='red')

In [3]: (Key("color") == "red") and (Key("sample") == "Ni")
TypeError                                 Traceback (most recent call last)
<ipython-input-3-b3c7952f4b70> in <module>
----> 1 (Key("color") == "red") and (Key("sample") == "Ni")

~/Repos/bnl/tiled/tiled/queries.py in __bool__(self)
     18 class NoBool:
     19     def __bool__(self):
---> 20         raise TypeError(
     21             """Queries are not "truth-y" or "false-y". They must be passed to search().
     22 

TypeError: Queries are not "truth-y" or "false-y". They must be passed to search().

You may be seeing this error message because you tried to use a tiled query
with the Python keywords `and` or `or`. This does not (cannot) work.
To compose queries, chain search calls like:

    c.search(...).search(...).search(...)
```

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -89,7 +89,7 @@ release = tiled.__version__
 #
 # This is also used if you do content translation via gettext trees.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
--- a/docs/source/reference/queries.md
+++ b/docs/source/reference/queries.md
@@ -43,7 +43,7 @@ c.search(...).search(...).search(...)
 
 ````{warning}
 
-**You cannot use queries with the Python keywords `and` or `or`.**
+**You cannot use queries with the Python keywords `not`, `and`, or `or`.**
 
 In Python, `and` and `or` have a particular behavior:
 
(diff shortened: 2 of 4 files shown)


danielballan:
The one test failure is the race involving slow CI runner disk.

```
_________________________ test_collision_after_startup _________________________
example_data_dir = local('/tmp/pytest-of-runner/pytest-0/temp5')
    def test_collision_after_startup(example_data_dir):
        """Test that files which produce key collisions are ignored until the collision is resolved."""
        config = {
            "trees": [
                {"tree": "files", "path": "/", "args": {"directory": str(example_data_dir)}}
            ]
        }
        client = from_config(config)
        assert "a" in client
        # Add a.tiff which will collide with a.tif.
        p = Path(example_data_dir, "a.tiff")
        with pytest.warns(UserWarning):
            tifffile.imwrite(str(p), data)
            force_update(client)
>       assert "a" not in client
E       AssertionError: assert 'a' not in <Node {'more', 'another_table', 'b', 'tables', 'c', 'a'}>
```

Unrelated to changes.

PR #233, opened 2022-06-10
add delete route

kleinhenz:
This adds a DELETE route at `/node/delete` which calls `entry.delete()` for deleting data from the server. Not sure if the route should be `/node/delete` or just `/node/metadata`.

### DIFF
--- a/tiled/client/node.py
+++ b/tiled/client/node.py
@@ -363,7 +363,7 @@ class Node(BaseClient, collections.abc.Mapping, IndexersMixin):
         self._cached_len = None
 
         path = (
-            "/node/delete/"
+            "/node/metadata/"
             + "".join(f"/{part}" for part in self.context.path_parts)
             + "".join(f"/{part}" for part in self._path)
             + "/"


tiled/server/router.py
```diff
@@ -556,11 +556,25 @@ def post_metadata(
...
+@router.delete("/node/delete/{path:path}")
+async def delete(
+    request: Request,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
+):
+    if hasattr(entry, "delete"):
+        entry.delete()
+    else:
+        raise HTTPException(
+            status_code=404, detail="This path does not support deletion."
```
danielballan, line 573: Method not allowed (405) would be better here.

tiled/server/router.py
```diff
@@ -556,11 +556,25 @@ def post_metadata(
             specs=body.specs,
         )
     else:
-        raise HTTPException(status_code=404, detail="Failed to write in this node.")
+        raise HTTPException(status_code=404, detail="This path cannot accept metadata.")
 
     return json_or_msgpack(request, {"key": key})
 
 
+@router.delete("/node/delete/{path:path}")
```
danielballan, line 564: I think sticking with `/node/metadata/` is better.

The name is slightly confusing but we essentially treat that route as the canonical home for the data so I think we should carry that pattern on here.

danielballan (changes requested):

tiled/server/router.py
kleinhenz: is that true for the other writing routes as well?

danielballan: Ah, it is! I didn’t think of this at first because the original toy implementation added new routes. Mind updating while you are here?

kleinhenz: updated all of them

danielballan (approved):
Merge at will when CI passes.

jmaruland (commented):
Came here late while I was fixing other bugs. But I want to highlight that I like this feature

Issue #135, opened 2021-11-30
Add "license" as a first-class concept


danielballan:
A license could be recorded _now_ as metadata, but this issue proposes to make a separate first-class entity in Tiled that is separately exposed in the API. It should be possible to integrate Access Control Policies.

The license should include for example:

* a canonical license name
* a link to the full license text
* a link to the source (e.g. DOI)
* an author name (CC BY ...)

danielballan:
Collecting links to user agreements:

* https://als.lbl.gov/wp-content/uploads/2016/07/nonpropuseragreement_sample-2019.pdf

danielballan:
We should consider whether this should be added as a top-level key _beside_ `metadata`, `specs`, and `references` or as a name with special significance _within_ `references`. (I think we should keep it out of `metadata` because it's important that that be a user-space namespace with keys that mean nothing to Tiled.)

danielballan:
See comments on the rationale for removing `references`: https://github.com/bluesky/tiled/issues/432#issue-1657348968

Likewise, I think license should just live within `metadata`, marked up using some existing standard like JSON-LD.

jmaruland:
NERDm uses license as part of its schema. You can see it here:
https://data.nist.gov/od/dm/nerdm/#Resource.license

PR #242, opened 2022-06-30
Add "references".

danielballan:
This PR does a couple loosely related things:

* Document what types are allowed in `metadata` and how we handle them in JSON and msgpack respectively. Nothing has _changed_ here; I'm just documenting it belatedly.
* Document the intent behind `specs`.
* Make it possible to pass `specs` in `__init__`. Previously, you had to write a subclass if you wanted to provide `specs`.
* Add `references`, about which more below.
* Add properties `c.specs` and `c.references` as aliases for `c.item["specs"]` and `c.item["references"]`. This gives brevity, obviously, but maybe more importantly it gives discoverability and a place for docstrings to describe what these are.

### Motivation

We want to explicitly link processed data to raw data. We could provide a uid like this:

```py
c.write_array(arr, metadata={"raw_uid": "..."}))
```

That is a useful think to do, but I think we should also provide a complete link. It's not always clear how to resolve that `uid`: Is that a Tiled uid? Where? In which Tiled deployment? Or is it something external? It would be best to provide a link to `https://.../node/metadata/...` or whatever external service so it's clear.

I think that, much like the proposal #135, these deserve to be a first-class concept in Tiled, not lumped in with generic `metadata`. By placing them in a dedicated flat dict we can:

* Ensure that they are well-formed URLs
* Create generic (i.e. facility-wide) services which verify that the links are not dead, both at insertion time and periodically after if we like
* Display them as links in UIs

### Implementation

This PR adds a dict named `references` as a _peer_ to `metadata`, `specs`, `structure_family`, and the other attributes. Clients can provide references when they write, as in:

```py
c.write_array(arr, metadata={"raw_uid": "..."}), references={"raw": "https://..."})
```

This will require an update to all Adapters that implement `post_metadata`. They need to accept (and presumably store) a `references` parameter. They should expose a `references` property.

### Naming

We could have called it `links`, but this would create confusion with the existing `links` (one level of nesting up) which are trusted, tiled-internal links which are used by the client to fetch data.

I think the name "references" correctly suggests that this is intended to be used to reference external entities, which could include references to other nodes in Tiled, references to external data or metadata providers, and even references to academic publications.

### Internal specification?

I think that, to start, we should not establish any naming scheme for the keys. Even something as simple as `"raw"` for raw data becomes complex when we consider processed data with multiple inputs (e.g. light and dark frames) which need to be differentiated. I think that the meaning of these references will be highly situation dependent. We may develop conventions or recommendations over time, but I don't think we should be prescriptive until we get some use in the wild.


docs/source/explanations/metadata.md
```diff
@@ -0,0 +1,63 @@
+# Metadata, Specs, and References
+
+## Metadata
+
+Each Tiled node carries an optional dictionary of metadata. This is fully
+under the control of the user: Tiled itself is opinionated about its content
```
kleinhenz, line 6: not opinionated?

danielballan: Inded. Thanks. :- )

x94carbone:
@danielballan I think this is a great idea. I agree with the principle that these should be a forefront concept. Indeed, the `references` can be optional since not every datum will have a reference, but it is critical to create an unbroken reference chain (a DAG if you will) of parent and child data points.

I think it fits into your framework, but it should definitely be the case that there can be multiple references. For example, while multiple children can each share the same parent, there are some operations (such as my classic example: taking the average) in which one child can have more than one parent. This will allow for that I'm guessing?

Finally, there are a ton of "uuid-like" objects floating around. Have we put that to rest? I.e., is there only one `uid` now? 😄 

I'm happy to provide more detailed feedback if you'd like!

danielballan:
Yes, references are optional and yes there is room for as many as you need. I agree that we should expect "children" to commonly have multiple "parents".

I'm not sure what you're referring to regarding UUIDs. Tiled doesn't have a concept for that built in. Every node (i.e. "datum") has a unique address, like `https://example.com/node/metadata/path/to/datum` that identifies it within Tiled. That path may or may not include a UUID. For example, you are serving files like

```
$ tiled serve directory my_files/
```

it likely does not. If you are serving data stored in MongoDB, it likely may, as that's a useful way to uniquely label entries in a database.

x94carbone:
@danielballan what I mean in terms of the unique id's is what is put in the `reference` key. I suppose that's the address you're talking about?

danielballan:
Oh, I see. Yes. Each node in Tiled has a canonical URL, which from Python you can get via

```py
node.uri
```

and that would be the best thing to use in the reference.

dylanmcreynolds:
I like this idea. I have a couple of thoughts. 

First, in your example:
```python
c.write_array(arr, metadata={"raw_uid": "..."}), references={"raw": "https://..."})
```
the type of reference is also the key, precluding you from having multiple references of the same type. I could see use cases where a derived data set was created from more than one raw data set, and vice versa.

Second, there are a number of specifications in the world of the semantic web for establishing links between objects wherein the relationship is described (RDF, OWL, XLink, JSON-LD). I'm not an expert in this area, but I think supporting JSON-LD should get some consideration, if we're growing first-class support for linking.

To bring the two thoughts together, if we made `references` a list, then we could start growing more structure into it.

```python
c.write_array(arr, metadata={"raw_uid": "..."}), 
   references=[
        {"raw": "https://food"},
        {"raw": "https://bar])
```

danielballan:
Oh, I like that.

In some cases I think it's good to nudge users to differentiate their inputs with distinct names. For example, if we are linking to "light frames" and "dark frames" `[{"light": ...}, {"dark": ...}]` is much better than `[{"raw": ...}, {"raw": ...}]`. But in cases like @x94carbone's example (averaging) you really do just want an undifferentiated list of inputs---basically `*args` in Python terms, and inventing arbitrary names would be anti-helpful.

List of dicts for now, with the possibility of expanding to richer values later for JSON-LD and the like.

gwbischof:
Adding references changes tiled from a tree structure to a graph.  A possible use case is that maybe want to export a run and its children, the runs children are accessed by references.  If we can't differentiate between parent and child references then the export can fail due to hitting cycle.  So for this reason I think references should have a direction in the first pass.  And we can decide to add more metadata in the future.

Also, I propose adding a more general version of `write_array`, that writes a document that can have `data`, `metadata`, and `references`.

gwbischof:
Actually, export could still work with a cycle, if it keeps track of which nodes are visited.
So I don't have a good reason direction is needed at this point.

danielballan:
Yes, a method for uploading structured datasets, such as

```py
create_node(dict_like, metadata, specs, references)
```

is a natural next step. @jmaruland had been working on a `create_dataset` method in #237 for uploading xarrays specfically, with the thought of broading from there. I think we've learned from that work that we may instead want a general method (`create_node`) that can upload any object that can be traversed like a dict. Those may include `xarray.Dataset`, `dict`, `h5py.Group`, and maybe other thing we haven't thought of yet.

Based on what we learned in the attempt in #237, I think we need to first:

* Simplify the xarray handling. (That's in progress now, and it's making the codebase _much_ simpler.)
* Reconcile @kleinhenz's work exploring the duality of search and tree-traversal.

I think the references _do_ have a direction, in a sense: they point from the node that holds the reference to the node in the reference. I suspect that we will mostly point from children to parents; otherwise we would be updating parents to inform them when analysis results are added or removed.

If we find that really need a hard vocabulary for this for one reason or another, I think we should explore JSON-LD and others. I don't see much downside in layering that on later once we know more.

danielballan:
Now I understand the question. I think we are using the word "parents" in two distinct ways.

Tiled already tracks the parent--child relationship for nodes and their children, and there is no ambiguity about the direction. You can query a node's children with

```
GET /node/search/{path}
```

and you can see a nodes parents in its path, like `/node/metadata/a/b/c`

This PR adds a way to mark additional relationships (a graph, as you say) that can provide a directional link between nodes and, further, label the nature of that link (`"raw"`). These extra relationships are not used to encode the parent--child relationship that builds the tree.

I think this is similar to how HDF5 file have a tree-based structure but additionally allow "links".

tiled/_tests/test_specs_and_references.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # a dict instead of a list of dicts
+    tree = MapAdapter({}, references={"ref_test": "https://example.com"})
+    with pytest.raises(pydantic.error_wrappers.ValidationError):
+        from_tree(tree)
+
+    # dict has too many items
+    tree = MapAdapter(
+        {}, references={"ref_test": "https://example.com", "extra": "oops"}
+    )
```
maffettone, line 41:
```suggestion
    tree = MapAdapter(
        {}, references=[{"ref_test": "https://example.com", "extra": "oops"}]
    )
```
This should be a list of dict.

tiled/_tests/test_specs_and_references.py
```diff
@@ -0,0 +1,48 @@
...
+    with pytest.raises(pydantic.error_wrappers.ValidationError):
+        from_tree(tree)
+
+    # dict has not enough items
+    tree = MapAdapter({}, references=[{}])
+    with pytest.raises(pydantic.error_wrappers.ValidationError):
+        from_tree(tree)
```
maffettone, line 48:
```suggestion
        from_tree(tree)
    
    # Not a dict
    tree = MapAdapter({}, references=[("ref_test": "https://example.com")])
    with pytest.raises(pydantic.error_wrappers.ValidationError):
        from_tree(tree)
    tree = MapAdapter({}, references=[["ref_test": "https://example.com"]])
    with pytest.raises(pydantic.error_wrappers.ValidationError):
        from_tree(tree)
```

maffettone:
Everyone's notes thus far make good sense. This is a very useful addition to tracking the source of derived data. 

I'm curious to why use a list of single item dictionaries instead of a `List[Tuple[str, URL]]`? I also suggested some extra tests. 

Thinking of this as a directed graph, I'm thinking about how I could break this nefariously and/or accidentally. 

## Creating a cycle (similar to/same as @gwbischof note)
We assemble a directed tree of nodes in search. When we start to write nodes with references, do we ever intend on assembling the graph or tracing relevant data? This will be challenging (though not impossible) with cases like where I can know or specify the uri, or want two derived datasets to know about eachother:
```python
c.write_array(arr1, metadata={"raw_uid": "..."}), 
    references=[
        {"raw": "raw_uri"},
        ...
        {"other_derived": "uri2"])
c.write_array(arr2, metadata={"raw_uid": "..."}), 
    references=[
        {"raw": "raw_uri"},
        ...
        {"other_derived": "uri1"])
```

## Dead links
What is the strategy for handling bad or dead links? Pydantic naturally shouldn't be checking links beyond validity. But if there is a link traversal (I don't see one yet in the code changes), then caution should be taken to elegantly handle dangling edges. In the averaging problem, it'd be a mess if a sum over 9999 failed to draw the raw data because 1 moved or wasn't created. 

## Nefarious links 
Most of my experiences with Tiled thus far has been serving up a database or filestore, and creating accessible URI/URLs. With user control over links, what is stopping me from putting in dangerous links?

danielballan:
Thanks for the extra test cases! The best kind of review comment. <3

**Tuples?** When I read @dylanmcreynolds's suggestion of list-of-dicts, I thought of list of tuples as well. I couldn't think of a clear reason to choose one over the other. Can anyone think of one?

**Cycles?** I am not sure how much we should worry about this in advance. What we're adding here is pretty basic and can be extended or (optionally) restricted in the future as we get some experience with what we'd want to build on it.

**Dead links** Yes, one of the motivations for creating a dedicated namespace with a clear structure is to enable automated proactive dead-link detection. I can imagine setting up periodic agents to check links and provide lists of dead ones, for example.

**Nefarious links** This occurred to me too. When we build client tools that follow links, we might apply [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) rules --- i.e. only follow links within in the subdomain of the provider, such as `tiled.nsls2.bnl.gov`, or else from some explicitly allowed list. I think it would be very useful to be able to include links to random publications or data sources, but if that's concerning we could add optional server configuration that rejects any links not matching a list of trusted subdomains. Perhaps that's worth doing now. I'm curious what @dylanmcreynolds thinks.

danielballan:
> I couldn't think of a clear reason to choose [between list-of-tuples vs list-of-dicts]. Can anyone think of one?

I just remembered that we have run into this before with parameters sorting in #161. OpenAPI has a problem with mixed-type collections. You can spell `Tuple[str, URL]` in pydantic but it does not translate into OpenAPI. Instead one should do `[{"key": "raw", "value": "https://…"}, …]`. That is valid, and putting user-provided strings in values instead of keys is more idiomatic anyway.

The above is a bit verbose but it applies only to the wire protocol, the HTTP API. No one generally has to type that.

Fortunately, in the Python API we can accept both the list-of-tuples and list-of-dicts spelling because converting either of them into the wire protocol representation is simple and unambiguous. Before Python 3.6/3.7 when dicts were unordered, you could say they were slightly different, but now they are equivalent.

maffettone:
My thought was more so for conveying intent. A tuple has a fixed length and can be easily zipped in a list. Dictionaries can grow, even when they shouldn't, and may confuse the early user. "Give me a list of tuples of (x, y) coordinates" seems much more natural and prone to expected behavior than "give me a list of dictionaries of {x:y} pairs, but make sure there is only one pair per dictionary".

Would it be better, (but also more work) to to do `List[Tuple[str, URL]]` and handle the conversion for OpenAPI internally?

danielballan:
Yeah, I think I agree. This is exactly what we do for sorting: the Python API looks like `[("plan_name", 1), ("time", -1)]` and the HTTP API looks like `[{"key": "plan_name", "direction": 1}, {"key": "time", "direction": -1}]`.

https://github.com/bluesky/tiled/blob/50cbfdd7296d6366053e489ec63c44936c8d05ec/tiled/client/node.py#L134-L140

Looks like the same approach could be applied here and satisfy both our collective preferences and the technical constraints.

danielballan:
I noted above that this was "on ice" for awhile, waiting for disruptive refactors to finish. Those are done, and now @jmaruland is picking this up to first rebase it and then rework it in accordance with the discussion above.

tiled/adapters/array.py
```diff
@@ -1,5 +1,4 @@
 import dask.array
-import numpy
```
danielballan, line 2: I think this is a rebase mistake and should be restored.

tiled/adapters/array.py
```diff
@@ -41,15 +53,16 @@ def __init__(self, array, *, chunks=None, metadata=None, dims=None, specs=None):
         self._metadata = metadata or {}
         self._dims = dims
         self.specs = specs or []
+        self.references = references or []
 
     @classmethod
-    def from_array(cls, array, *, chunks=None, metadata=None, dims=None, specs=None):
+    def from_array(cls, data, *, metadata=None, dims=None, specs=None, references=None):
```
danielballan, line 59: The only change here should be adding `references`.

tiled/adapters/array.py
```diff
@@ -41,15 +53,16 @@ def __init__(self, array, *, chunks=None, metadata=None, dims=None, specs=None):
...
         return cls(
-            numpy.asarray(array),
-            chunks=chunks,
+            dask.array.from_array(data),
```
danielballan, line 61: Ditto here. Nothing in this section should change: only adding `references` at the end.

tiled/adapters/dataframe.py
```diff
@@ -28,19 +28,37 @@ class DataFrameAdapter:
     structure_family = "dataframe"
 
     @classmethod
-    def from_pandas(cls, *args, metadata=None, specs=None, **kwargs):
+    def from_pandas(cls, *args, metadata=None, specs=None, references=None, **kwargs):
         ddf = dask.dataframe.from_pandas(*args, **kwargs)
-        return cls.from_dask_dataframe(ddf, metadata=metadata, specs=specs)
```
danielballan, line 33: The only change here should be adding `references`.

tiled/adapters/dataframe.py
```diff
@@ -61,7 +79,15 @@ def read_csv(cls, *args, metadata=None, specs=None, **kwargs):
         cache = get_object_cache()
         if cache is not NO_CACHE:
             cache.discard_dask(ddf.__dask_keys__())
-        return cls.from_dask_dataframe(ddf, metadata=metadata, specs=specs)
```
danielballan, line 64: The only change here should be adding `references`.

tiled/adapters/excel.py
```diff
@@ -7,34 +7,33 @@
...
-    This maps the Excel file, which may contain one of more spreadsheets,
-    onto a tree of tabular structures.
-
-    Examples
-    --------
+    @classmethod
+    def from_file(cls, file, *, specs=None, references=None):
+        """
+                Read the sheets in an Excel file.
+        >>>>>>> a731c70 (Implement 'references'.)
```
danielballan, line 14: This needs cleaned up.

tiled/examples/xdi.py
```diff
@@ -8,7 +8,6 @@
 import io
 import re
 
-import dask
```
danielballan, line 11: I don't think anything in this file should be changed.

danielballan (commented):
Hints on fixing up the rebase.

danielballan:
Summarizing discussion from yesterday evening:

* In `__init__`, references sent from server (as JSON object) should be rearranged as list-of-tuples and stored in `self._references`.
* A public property, `references`, should give read-only access to `self._references`. Something we did _not_ discuss but I'll add here: following the pattern used elsewhere, `ListView` should be used to communicate to the user that this is only a _view_ of the references that they cannot be modified/mutated here. (Use `git grep ListView` to find examples.)
* In the methods `new` and `update_metadata` references should be rearranged from list-of-tuples to JSON object for communication with the server.

danielballan:
I think holding just a URL is enough for our immediate needs...

In Python:

```py
[("raw", "https://..."), ...]
```

In JSON:

```json
[{"label": "raw", "url": "https://..."}, ...]
```

But we may want to grow more just a URL, so maybe we want to build in that flexibility to make room to expand the content of a reference in a backward-compatible fashion:

In Python:

```py
[("raw", {"url": "https://...", "future additional key goes here": "stuff"}), ...]
```

In JSON:

```json
[{"label": "raw", "reference": {"url": "https://..."}}, ...]
```

If we start with just the URL, there will be no backward-compatible way to grow. So we should make a decision now about the likelihood that we'll want to grow.

danielballan:
Another option is to back away from using list-of-tuples in Python and just do

```py
[{"label": "raw", "url": "https://..."}, ...]
```

and handle that the same in Python and JSON. That is, naturally, extensible:

```py
[{"label": "raw", "url": "https://...", "future additional key goes here": "stuff"}, ...]
```

And it's nice that we don't have to worry about transforming between different representations in Python and JSON.

dylanmcreynolds:
> And it's nice that we don't have to worry about transforming between different representations in Python and JSON.

I fully support this.  You reduce future bugs introduced by people having to transform data structures

> which had the pitfall "give me a list of dictionaries of {x:y} pairs, but make sure there is only one pair per dictionary" 

That's true. In addition, it's more future proof, allowing for additional properties.

Going back to my previous (comment)[https://github.com/bluesky/tiled/pull/242#issuecomment-1171363735]. This structure also opens the door for creating references that are JSON-LD-compliant. Is that important? I haven't really come across a system that uses JSON-LD, but it's nice to keep that door open.

danielballan:
Yes I think keeping the door open, to JSON-LD, is a big benefit.

tiled/server/schemas.py
```diff
@@ -316,6 +307,19 @@ class GetDistinctResponse(pydantic.BaseModel):
...
+    url: pydantic.AnyUrl
+
+    @classmethod
+    def from_json(cls, json_doc):
+        return cls(label=json_doc["label"], url=json_doc["url"])
 
 
+NodeAttributes.update_forward_refs()
 NodeStructure.update_forward_refs()
+PostMetadataRequest.update_forward_refs()
```
danielballan, line 324: If you move `ReferenceDocument` so that it is defined above these models in the module, it will not be necessary to call this. This is only needed to resolve references that are _circular_ such that there is no way to define everything in the right order.

tiled/_tests/test_specs_and_references.py
```diff
@@ -19,9 +19,11 @@ def test_spec_is_converted_to_str():
...
-    tree = MapAdapter({}, references=[{"ref_test": "https://example.com"}])
+    tree = MapAdapter(
+        {}, references=[{"label": "ref_test", "url": "https://example.com"}]
+    )
     c = from_tree(tree)
-    assert c.references == [{"ref_test": "https://example.com"}]
+    assert c.references == [{"label": "ref_test", "url": "https://example.com"}]
 
 
 def test_bad_reference():
```
danielballan, line 29: Can you update these test cases as well? They fail as expected but not for the right reasons.

danielballan:
Note to self: Rebase this after #318 is in, and then add tests on top of that.

PR #246, opened 2022-07-06
Change xarray from structure family to spec.

danielballan:
This PR breaks every API (Python server API, Python client API, HTTP API) in exchange for a major reduction in conceptual complexity and code.

## Motivation

* Continue to support xarray datasets with as good or better performance as we presently have.
* Reduce the minimum complexity required of a client to receive nested data structures, including currently supported structures (e.g xarray dataset) and anticipated future ones (e.g. SCIPP).
* Remove complex special-case code for xarray server and client implementation.
* Remove inconsistencies about when data is loaded, like this:

```py
# Arrays (correctly) give you a client which you can export or slice or introspect. They do not greedily start sending you data.
c["some_array"]  # no data yet...
c["some_darray"][:]  # data!

# Xarray datasets on the other hand start sending data right away.
c["some_xarray_dataset"]["some column"]  # surprise! you just downloaded a TB
```

Instead, we move to:

```py
c["some_xarray_dataset"]["some_column"]  # no data yet...
c["some_xarray_dataset"]["some_column"][:]  # data!
```

This leaves room for slicing, or calling `.export(...)` or other inspection. (It's possible to do this now, but the API is hard to guess/discover.)

The same change applies to DataFrames as well. Old usage:

```py
c["some_dataframe"].read()  # pandas.DataFrame
c["some_dataframe"][.read(["A"])  # pandas.DataFrame
c["some_dataframe"][.read(["A, B"])  # pandas.DataFrame
c["some_dataframe"]["A"]  # pandas.Series
```

New usage:

```py
c["some_dataframe"].read()  # pandas.DataFrame (unchanged)
c["some_dataframe"][.read(["A"])  # pandas.DataFrame (unchanged)
c["some_dataframe"][.read(["A, B"])  # pandas.DataFrame (unchanged)
c["some_dataframe"]["A"]  # ArrayClient
c["some_dataframe"]["A"][:]  # array
```

## Conceptual Justification

Arrays, DataFrames, and (to-be-implemented) COO sparse arrays are fundamentally different structures, and if a client doesn't understand them there is no way to send that client data in a way that it can interpret. Each of these structures has wire protocols for multiple lanuages.

Xarray Datasets, on the other hand, _can_ be represented as a "directory" of Arrays. If you recognize that it's an xarray Dataset, you can take advantage of certain constraints and assumptions, but you don't _need_ to understand what this means in order to receive the data. Therefore, Datasets should be a _spec_ (an optional, upgraded description) not a structure family.

We foresee more structures in this space, like SCIPP and various NeXus layouts. We don't want to grow a structure family for each of these because it would generate a lot of work for client implementations. The lowest common denominator of all these is the `node` structure family.


danielballan:
I am happy with the documentation and the unit tests.

I want to benchmark before we merge. I will post results here.

tiled/adapters/xarray.py
```diff
@@ -1,223 +1,82 @@
...
+                    array_adapter.macrostructure().dims,
+                    array_adapter.read(),
+                )
+            elif "xarray_coord" in array_adapter.specs:
+                coords[key] = (
+                    array_adapter.macrostructure().dims,
+                    array_adapter.read(),
+                )
+            else:
+                assert False, "Expected a spec"
```
kleinhenz, line 47: I think this should probably raise instead of assert

tiled/adapters/xarray.py
```diff
@@ -1,223 +1,82 @@
...
+    def __len__(self):
+        return len(self._dataset.data_vars) + len(self._dataset.coords)
 
-    def read(self, fields=None):
-        ds = self._dataset
-        if fields is not None:
-            ds = ds[fields]
-        return ds.compute()
+    def __iter__(self):
+        yield from (list(self._dataset.data_vars) + list(self._dataset.coords))
```
kleinhenz, line 65: `itertools.chain`?

tiled/serialization/xarray.py
```diff
@@ -0,0 +1,88 @@
...
+        pass
+
+
+if modules_available("scipy"):
+
+    def serialize_netcdf(node, metadata):
+        file = _BytesIOThatIgnoresClose()
+        # This engine is reportedly faster.
+        # Also, by avoiding the default engine, we avoid a dependency on 'scipy'.
+        node.as_dataset().to_netcdf(file, engine="scipy")
```
kleinhenz, line 28: did you mean to change the engine here? this looks like it depends on scipy...

tiled/server/core.py
```diff
@@ -22,36 +22,32 @@
...
 _FILTER_PARAM_PATTERN = re.compile(r"filter___(?P<name>.*)___(?P<field>[^\d\W][\w\d]+)")
 _LOCAL_TZINFO = dateutil.tz.gettz()
 
+# Pragmatic limit on how "wide" a node can be
+# before the server refusing to inline its contents
+INLINED_CONTENTS_LIMIT = 100
+
+# Pragmatic limit on how deep the server will recurse into
+# nodes that request inlined contents
+DEPTH_LIMIT = 20
```
kleinhenz, line 49: This seems very aggressive. Maybe something more like 3-4?

tiled/client/xarray.py
```diff
@@ -1,540 +1,94 @@
...
-            if (
-                optimize_wide_table
-                and (data_shape[0] < LENGTH_LIMIT_FOR_WIDE_TABLE_OPTIMIZATION)
-                and (len(data_shape) < 2)
-            ):
-                data_vars_clients[name] = wide_table_fetcher.register(
-                    name, data_array, metadata
+            array_structure = array_client.structure()
+            shape = array_structure.macro.shape
+            if optimize_wide_table and (
```
kleinhenz, line 57: Wide table optimization also fails for "jagged" datasets but this is currently not caught by the server.
Consider: `dset = array.Dataset({f"{i}" :  xarray.DataArray(i*numpy.ones(2*i), dims=f"dim{i}") for i in range(10)})`

kleinhenz: I guess the wide table optimization should only be applied per dimension or more conservatively only if the dataset only has one dimension so it really is just a dataframe which can be transmitted using arrow.

tiled/server/core.py
danielballan: This is a hard upper bound meant to protect the server from being crashed by badly designed or buggy Adapters. It is up to Adapters to opt in to this behavior and decide on a reasonable depth.

I have seen small but highly nester NeXus files, deeper than 3–4 layers, that would be reasonable to marshall greedily.

But I think you are right that 20 is too high. I’ll dial it down to around 5, and we can always raise it later.

tiled/client/xarray.py
danielballan: Good call: I had forgotten it was even possible to construct a ragged Dataset because the Bluesky Datasets always share a time coordinate and cannot be ragged.

This optimization really helps in cases where some (most) of the columns are scalar but some are ND.

tiled/serialization/xarray.py
danielballan: It turns out that the scipy engine is the only one that can write to file-like objects (i.e. buffers). So here we use the scipy engine, but this code block runs conditionally on scipy being installed.

Fixed out-of-date comment in 6df6150.

tiled/adapters/xarray.py
danielballan: It's true that `assert` should not be used for user input validation, but in this case the class is responsible for construction of the `array_adapters` and it should be impossible to reach this branch unless there is some bug internally to the class, so I think this is a correct use of `assert`.

kleinhenz: I think you can reach this path if you construct `DatasetAdapter` from an improper mapping just using `__init__`.

kleinhenz: You might actually want this if for example you are storing all the arrays separately instead of in a single netcdf and don't want to bother constructing the xarray object on the server side first which is probably how upload will have to work.

danielballan: Fair point. I have made it less strict.

danielballan:
Here is a before/after of the traffic issued by the client to download an xarray Dataset with a mixture of 4D image data and 1D columns.

Before: 7 requests, 1008 +/- 345 ms

```
[<Request('GET', 'https://tiled.nsls2.bnl.gov/api/node/search/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504?filter%5Blookup%5D%5Bcondition%5D%5Bkey%5D=primary&sort=_')>,
 <Request('GET', 'https://tiled.nsls2.bnl.gov/api/node/search/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary?filter%5Blookup%5D%5Bcondition%5D%5Bkey%5D=data&sort=_')>,
 <Request('GET', 'https://tiled.nsls2.bnl.gov/api/array/block/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data/coords/time/variable?block=0&expected_shape=29')>,
 <Request('GET', 'https://tiled.nsls2.bnl.gov/api/array/block/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data/data_vars/Wide%20Angle%20CCD%20Detector_image/variable?block=0%2C0%2C0%2C0&expected_shape=10%2C1%2C1024%2C1026')>,
 <Request('GET', 'https://tiled.nsls2.bnl.gov/api/array/block/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data/data_vars/Wide%20Angle%20CCD%20Detector_image/variable?block=1%2C0%2C0%2C0&expected_shape=10%2C1%2C1024%2C1026')>,
 <Request('GET', 'https://tiled.nsls2.bnl.gov/api/array/block/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data/data_vars/Wide%20Angle%20CCD%20Detector_image/variable?block=2%2C0%2C0%2C0&expected_shape=9%2C1%2C1024%2C1026')>,
 <Request('GET', 'https://tiled.nsls2.bnl.gov/api/node/full/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data?format=application%2Fvnd.apache.arrow.file&field=WAXS+fullframe_total&field=RSoXS+Shutter+Opening+Time+%28ms%29&field=en_energy&field=en_energy_setpoint&field=en_polarization&field=en_polarization_setpoint')>]
```

After: 6 requests: 977 +/- 162 ms

```
[<Request('GET', 'https://tiled-staging.nsls2.bnl.gov/api/node/metadata/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary')>,
 <Request('GET', 'https://tiled-staging.nsls2.bnl.gov/api/node/metadata/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data')>,
 <Request('GET', 'https://tiled-staging.nsls2.bnl.gov/api/array/block/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data/Wide%20Angle%20CCD%20Detector_image?block=0%2C0%2C0%2C0&expected_shape=10%2C1%2C1024%2C1026')>,
 <Request('GET', 'https://tiled-staging.nsls2.bnl.gov/api/array/block/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data/Wide%20Angle%20CCD%20Detector_image?block=1%2C0%2C0%2C0&expected_shape=10%2C1%2C1024%2C1026')>,
 <Request('GET', 'https://tiled-staging.nsls2.bnl.gov/api/array/block/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data/Wide%20Angle%20CCD%20Detector_image?block=2%2C0%2C0%2C0&expected_shape=9%2C1%2C1024%2C1026')>,
 <Request('GET', 'https://tiled-staging.nsls2.bnl.gov/api/node/full/rsoxs/raw/a5050f88-9929-4f1d-a82c-ca7fa4b8f504/primary/data?format=application%2Fvnd.apache.arrow.file&field=time&field=WAXS+fullframe_total&field=RSoXS+Shutter+Opening+Time+%28ms%29&field=en_energy&field=en_energy_setpoint&field=en_polarization&field=en_polarization_setpoint')>]
```

The URLs of the requests are also simpler, easier to follow.

dylanmcreynolds (approved):
I won't be able to give this a detailed review. I do like the concept that we don't have to grow structure families with every new representation, though. Simplifications like this make sense to me.

kleinhenz (approved):

PR #248, opened 2022-07-06
Created .delete() from __delitem__ in node

jmaruland:
Re-open https://github.com/bluesky/tiled/pull/245


tacaswell:
I am 👎🏻 on this change.

The `Node` class is derived from `Mapping` so we should stick with the dict-like API of (e.g. `n[k]` (`__getitem__`), `n[k] = v` (`__setitem__`), and `del d[k]` (`__delitem__`)) without providing extra aliases for them. 

If there were additionally a `delete` method I would wonder if it was _different_ that `__delitem__` and if so why?

tacaswell (dismissed):
Do not think this is a good idea baring a use-case where using `def n[k]` or `n.__delitem__(k)` would not be usable.

danielballan (approved):
This PR has been reduced to a small bug-fix PR.

PR #252, opened 2022-07-11
Enforce reasonable limits on API keys and Sessions per Principal

### LINKED ISSUE #250: Limit API keys per Principal
If a user creates a huge number of API keys the route to list their API keys could become to expensive. In the future we could paginate that route, but for now I propose to put a reasonable limit on the number of active API keys per user. Also, we should verify that the route does not include deleted/inactive keys.

When a user hits the limit, the remedy is to delete disused keys.

### LINKED ISSUE #251: Limit Sessions per Principal
There should be a high but finite limit on the number of unexpired and active Sessions per Principal.

danielballan:
Put limits on the number of API keys and Sessions per Principal.

1. Ensure that the routes which list API keys and sessions, which are not paginated, returns in a reasonable time.
2. Avoid unintentional or intentional abuse.

<hr />

Closes #250 
Closes #251

### DIFF
--- a/tiled/_tests/test_authentication.py
+++ b/tiled/_tests/test_authentication.py
@@ -353,15 +353,11 @@ def test_session_limit(enter_password, config):
     original_limit = authentication.SESSION_LIMIT
     authentication.SESSION_LIMIT = 10
     try:
-        with enter_password("secret2"):
-            from_config(config, username="bob", token_cache={})
-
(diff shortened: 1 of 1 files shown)


cryos (approved):
Looks good to me, I might be stingier with the default limits personally. A normal user should rarely need more than ten or twenty API keys in my opinion.

PR #256, opened 2022-07-18
Enable matching wildcards in file_ext.

danielballan:
This should enable matching files like

```
thing.txt.1
thing.txt.2
```

with

```yaml
mimetypes_by_file_ext:
  .txt.*: ...
```

and files like

```
thing.1.txt
thing.2.txt
```

with

```yaml
mimetypes_by_file_ext:
  .*.txt: ...
```

This still only has access to the file _extension_ (everything after the first dot).

### DIFF
--- a/tiled/adapters/files.py
+++ b/tiled/adapters/files.py
@@ -7,6 +7,7 @@ import queue
 import re
 import threading
 import warnings
+from fnmatch import fnmatch
 from pathlib import Path
 
 import cachetools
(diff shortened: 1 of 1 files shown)


danielballan:
I will close this. The hook added in #259 can be used instead. This could provide a lighter-touch way to handle certain use cases, but I think they are rare enough that it isn't worth adding complexity to the logic.

PR #266, opened 2022-07-27
Support updating metadata and accessing metadata revision history

jmaruland:
This PR supports the client and server features that are being tested alongside the experiments in https://github.com/bluesky/databroker/pull/716 which, by itself, allows the client to change the content of the `metadata` and `specs` of an existing sample

### DIFF
--- a/tiled/client/base.py
+++ b/tiled/client/base.py
@@ -1,4 +1,3 @@
-import copy
 import importlib
 import time
 
@@ -6,7 +5,7 @@ from ..utils import UNCHANGED, DictView, ListView, OneShotCachedMap
 from .cache import Revalidate, verify_cache
 
 
-class Metadata_Revisions:
+class MetadataRevisions:
     def __init__(self, context, path):
         self._cached_len = None
         self.context = context
@@ -49,9 +48,11 @@ class Metadata_Revisions:
             offset = item_
             limit = 1
 
-            return self.context.get_json(
+            content = self.context.get_json(
                 path, params={"page[offset]": offset, "page[limit]": limit}
             )
+
+            return content["data"]
 
         elif isinstance(item_, slice):
             offset = item_.start
@@ -65,20 +66,18 @@ class Metadata_Revisions:
                 params = f"page[offset]={offset}&page[limit]={limit}"
 
             next_page = path + params
-            result = None
+            result = []
             while next_page is not None:
                 content = self.context.get_json(path)
-                if result is None:
-                    result = copy.deepcopy(content)
+                if len(result) == 0:
+                    result = content.copy()
                 else:
                     result["data"].append(content["data"])
                 next_page = content["links"]["next"]
 
-            return result
+            return result["data"]
 
-    def __delitem__(self, n):
-
-        self._cached_len = None
+    def delete_revision(self, n):
 
         path = (
             "/node/revisions"
@@ -241,7 +240,7 @@ class BaseClient:
     @property
     def metadata_revisions(self):
         if self._metadata_revisions is None:
-            self._metadata_revisions = Metadata_Revisions(self.context, self._path)
+            self._metadata_revisions = MetadataRevisions(self.context, self._path)
 
         return self._metadata_revisions
 
--- a/tiled/client/base.py
+++ b/tiled/client/base.py
@@ -1,7 +1,91 @@
 import importlib
+import time
 
 from ..utils import UNCHANGED, DictView, ListView, OneShotCachedMap
 from .cache import Revalidate, verify_cache
+
+
+class Metadata_Revisions:
+    def __init__(self, context, path):
+        self._cached_len = None
+        self.context = context
+        self._path = path
+
+    def __len__(self):
+
+        LENGTH_CACHE_TTL = 1  # second
+
+        now = time.monotonic()
+        if self._cached_len is not None:
+            length, deadline = self._cached_len
+            if now < deadline:
+                # Used the cached value and do not make any request.
+                return length
+
+        path = (
+            "/node/revisions"
+            + "".join(f"/{part}" for part in self.context.path_parts)
+            + "".join(f"/{part}" for part in (self._path or [""]))
+        )
+
+        content = self.context.get_json(
+            path, params={"page[offset]": 0, "page[limit]": 0}
+        )
+        length = content["meta"]["count"]
+        self._cached_len = (length, now + LENGTH_CACHE_TTL)
+        return length
+
+    def __getitem__(self, item_):
+        self._cached_len = None
+
+        path = (
+            "/node/revisions"
+            + "".join(f"/{part}" for part in self.context.path_parts)
+            + "".join(f"/{part}" for part in (self._path or [""]))
+        )
+
+        sample = []
+        if isinstance(item_, int):
+            offset = item_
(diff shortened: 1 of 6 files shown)


tiled/client/base.py
```diff
@@ -121,6 +121,39 @@ def formats(self):
...
+            User metadata. May be nested. Must contain only basic types
+            (e.g. numbers, strings, lists, dicts) that are JSON-serializable.
+        specs : List[str], optional
+            List of names that are used to label that the data and/or metadata
+            conform to some named standard specification.
+        """
+
+        self._cached_len = None
+
+        data = {  # noqa: F841
```
danielballan, line 142: This F841 seems off. The variable is used below.

My guess is that you were trying to commit partial work, when you planned to use the variable but were not yet using it. In that situation, it is useful to know about 

```
git commit --no-verify ...
```

which disables the pre-commit hooks and enables you to commit partial work. This is IMO better than putting # noqa exemptions in place temporarily because it is easy to forget to remove them, as may have happened in this case.

tiled/server/router.py
```diff
@@ -617,3 +617,18 @@ async def put_dataframe_full(
...
     return json_or_msgpack(request, None)
+
+
+@router.put("/node/metadata/{path:path}")
+async def put_metadata(
+    request: Request,
+    body: schemas.PutMetadataRequest,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
+):
+    if hasattr(entry, "put"):
```
danielballan, line 628: Should this check for the method we are about to use?

tiled/server/router.py
```diff
@@ -617,3 +617,18 @@ async def put_dataframe_full(
...
+        entry.put_metadata()
```
danielballan, line 629: Does this need the metadata and specs as parameters?

danielballan (changes requested):

tiled/server/router.py
```diff
@@ -617,3 +616,21 @@ async def put_dataframe_full(
             status_code=405, detail="This path cannot accept dataframe data."
         )
     return json_or_msgpack(request, None)
+
+
+@router.put("/node/metadata/{path:path}")
+async def put_metadata(
+    request: Request,
+    body: schemas.PutMetadataRequest,
+    entry=Security(entry, scopes=["write:data", "write:metadata"]),
```
danielballan, line 625: The scopes listed on some of the writing routes were too broad, which I just fixed in existing routes in 5c79c37. This route should only require `write:metadata`.

```suggestion
    entry=Security(entry, scopes=["write:data", "write:metadata"]),
```

tiled/server/router.py
jmaruland: The right parameters were added in the last commit

tiled/server/router.py
jmaruland: Fixed in new commit

tiled/client/base.py
```diff
@@ -1,9 +1,94 @@
...
+            "/node/revisions"
+            + "".join(f"/{part}" for part in self.context.path_parts)
+            + "".join(f"/{part}" for part in (self._path or [""]))
+        )
+
+        if isinstance(item_, int):
+            offset = item_
+            limit = 1
+
+            return self.context.get_json(
```
danielballan, line 52: Extra `["data"]` here.

tiled/client/base.py
```diff
@@ -1,9 +1,94 @@
...
+                offset = 0
+
+            if item_.stop is None:
+                params = f"?page[offset]={offset}"
+            else:
+                limit = item_.stop - offset
+                params = f"page[offset]={offset}&page[limit]={limit}"
+
+            next_page = path + params
+            result = None
```
danielballan, line 68: Since we now plan to return the content of `"data"`, I think this can be initialized as a list.

tiled/client/base.py
```diff
@@ -1,9 +1,94 @@
+import copy
 import importlib
+import time
 
 from ..utils import UNCHANGED, DictView, ListView, OneShotCachedMap
 from .cache import Revalidate, verify_cache
 
 
+class Metadata_Revisions:
```
danielballan, line 9:
```suggestion
class MetadataRevisions:
```

danielballan:
We may need to revisit the usage of `del`. We target revisions based on revision number, not on their position in the list. I think it's correct to target the revision number, bu I now realize that using `del` for is very confusing because

```py
del x[5]
```

always means "Delete the fifth element," not, "Delete an element with an internal label that is equal to `5`."

danielballan:
Also, reminder to selves:

As we noticed in the demo, we need `update_metadata(...)` to update the local item as well (if the update has succeeded).

jmaruland:
> We may need to revisit the usage of `del`. We target revisions based on revision number, not on their position in the list. I think it's correct to target the revision number, bu I now realize that using `del` for is very confusing because
> 
> ```python
> del x[5]
> ```
> 
> always means "Delete the fifth element," not, "Delete an element with an internal label that is equal to `5`."

I can change it to something like
```python
x.delete_revision(revision=5)
```
This gives a better a idea of what we are trying to do, right?

danielballan:
Yeah that seems clear.

danielballan:
This needs a rebase.

tiled/client/base.py
```diff
@@ -121,6 +205,54 @@ def formats(self):
...
+        full_path_meta = (
+            "/node/metadata"
+            + "".join(f"/{part}" for part in self.context.path_parts)
+            + "".join(f"/{part}" for part in (self._path or [""]))
+        )
+
+        content = self.context.put_json(full_path_meta, data)
+
+        if metadata is not None:
+            if len(content["metadata"]) > 0:
```
danielballan, line 240: If we make it here, we know the client submitted a change that was accepted. The only additional question we have now is, "Did the server modify the `metadata` (via a validator) before accepting it?" This should be answer by `content["metadata"] is None`, not be an `len()` check.

tiled/server/router.py
```diff
@@ -740,3 +740,93 @@ async def put_dataframe_partition(
...
+        entry.put_metadata(
+            metadata=metadata,
+            specs=specs,
+        )
+
+        response_data = {"id": entry.key}
+        if metadata_modified:
+            response_data["metadata"] = metadata
+        else:
+            response_data["metadata"] = {}
```
danielballan, line 789: Returning `{}` would mean, "I _did_ modify the metadata; I modified it to be empty." We want `None`.

tiled/client/base.py
```diff
@@ -1,9 +1,92 @@
...
+
+        if isinstance(item_, int):
+            offset = item_
+            limit = 1
+
+            content = self.context.get_json(
+                path, params={"page[offset]": offset, "page[limit]": limit}
+            )
+
+            return content["data"]
```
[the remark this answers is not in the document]
jmaruland: This is tested in `test_experimental.py` when the client brings a single document to validate it against the document schema.
I'm applying the changes in `base.py` and adding the respective change in `test_experimental.py` to receive one object
https://github.com/jmaruland/databroker/blob/Add-timestamps-to-experimental-document/databroker/tests/test_experimental.py#L329-L334

danielballan (approved):

PR #291, opened 2022-08-10
Fix error message formatting.

danielballan:


danielballan:
@jklynch confirmed via Slack that this fixes the issue.

PR #295, opened 2022-08-11
Avoid using deserialization registry for simple case.

danielballan:
We defer registering (de)serializers unless we need them because it triggers imports of many optional dependencies.

I think the best solution is to incorporate dask's clever "lazy registration" mechanism, but for now this is a simpler path.


danielballan:
Merging as hotfix for regression in #293

Issue #288, opened 2022-08-09
Remove _async_bridge module 


danielballan:
Use `TestClient` to replace the async bridge. It does not introduce a Starlette dependency because if we have an app, we already have Starlette.

https://fastapi.tiangolo.com/tutorial/testing/

Issue #281, opened 2022-08-07
Refactor client cache to use httpx transport


danielballan:
By using a custom httpx Transport we can remove significant complexity from the caching logic.

There is a lot about how this is factored that makes sense to me: https://github.com/obendidi/httpx-cache/tree/main/httpx_cache

I like:
* Use of a Transport to inject into Client at the right layer
* Cache Controller (which status codes, methods, etc)
* Pluggable serializer
* Avoid the complexity of a separate URL -> ETag, ETag -> content mappings. That optimizes for a pretty rare situation (same content at different URLs) and adds lookup overhead as well as complexity. I should strip that out.
* A default cache location in ~/.cache/tiled? Maybe? Once this is robust, and as long as NFS does not actually make this slower.

I question:
* There is no mechanism for bounding the cache that I can find. We want to keep our scoring mechanism.
* Serializing with msgpack forces us to deserialize the whole (maybe large body) before we know whether the cache entry is expired. Maybe we can pack the content and/or stream in a second, separate msgpack chunk such that we do not read it if we know we won’t use it.
* The RW locks are correct (and better than my messy “reservation” idea) but do not work for multiprocess clients. I think I need a RW lock on top of locket.

danielballan:
Our current implementation makes a half-hearted attempt at multi-process support but it's flawed in a couple ways. I think it would be better to explicitly make the file-based cache single-process --- maybe even added a pidfile lock. If we do that, that opens up the possibility of using sqlite to store header info, size, etc.

danielballan:
I think we should change the internal cache layout from many-small-separate-files to a single SQLite database file, referencing some external files only for large data (MB-scale image chunks, not KB-scale metadata). This will enable better read and write performance:

https://www.sqlite.org/fasterthanfs.html
https://www.sqlite.org/intern-v-extern-blob.html

SQLite supports concurrent reads from multiple processes. It manages concurrent writes with an internal locking mechanism, though I have questions about how that performs on distributed file systems (e.g. NFS, Lustre), both in terms of speed and risk of corruption.

I think it makes sense to steer multiprocess _writing_ to a dedicate bulk download utility. It will manage multiple process, perhaps writing to separate SQLite files and merging them at the end. The merged file will be safe for multiprocess reading.

danielballan:
Notes from a design conversation with @tacaswell, working from the top down...

[...]

Issue #116, opened 2021-10-23
Add async client objects


danielballan:
- [x] Remove `tiled/client/_async_bridge.py`, a temporary hack that runs the `httpx.AsyncClient` on a thread to make ASGI work.
- [x] Rework the refresh flow code to httpx's [sans I/O mechanism for pluggable authentication](https://www.python-httpx.org/advanced/#customizing-authentication).
- [ ] Rework the rest of the client code following a pattern similar to the authentication one, aiming for sans I/O utilities, breaking out sync/async cases as needed to handle I/O. I'm not sure whether following httpx in using subclasses is best here. Consider composition.
- [x] Use this opportunity to make `tiled.client.node.Node` inherit from `tiled.client.base.BaseClient`.

canismarko:
Thanks, Dan. I see that `Context()` already has two `@classmethod` constructors *from_app()* and *from_any_uri()*. I'll see if the I/O calls fit in there (or if not a new one is fine with me too).

I'm not sure yet where I fall on the ``__getitem__()`` issue. I like `async def get()` since that mirrors the usual interface for mappings, so I'll add that either way. I agree that `await container[…]` is strange if you're not expecting it, though if you're using the async client you're probably okay with that kind of strangeness? I could go either way.

danielballan:
Following up on my quick suggestion that I gave, standing in @genematx's office with @canismarko on Zoom:

- For httpx's [client objects](https://github.com/encode/httpx/blob/master/httpx/_client.py), the sans-I/O behavior is defined in a base class, and subclasses (sync) `Client` and (async) `AsyncClient` add the I/O.
- The httpx [Auth](https://www.python-httpx.org/advanced/authentication/#custom-authentication-schemes) uses a generator-based approach to construct requests that are then sent sync or async. See [`httpx._auth`](https://github.com/encode/httpx/blob/master/httpx/_auth.py) for implementation.
- Interestingly, the [`httpx.Response` object](https://github.com/encode/httpx/blob/b5addb64f0161ff6bfe94c124ef76f6a1fba5254/httpx/_models.py#L515) uses a single object to expose both sync and async methods. I suspect that is pragmatic but would get messy for us, because the validity of sync or async methods would be context-dependent.

canismarko:
I was working on an API client for a different project and had this discussion in mind when writing it.

[...]

PR #296, opened 2022-08-12
Revamp client

danielballan:
This is a placeholder for what I expect to be a large PR with many interdependent aspects, effectively ripping the client apart and reassembling better.

- [x] Remove vestigial `path` parameter. (This information is superseded by `item`).
- [ ] Factor authenticaiton logic from `Context` into a separate `httpx.Auth` subclass.
- [ ] Factor caching and `offline` ("airplane mode") logic into the transport layer, and rethink caching. More planning notes here: https://github.com/bluesky/tiled/issues/281
- [ ] Make async varieties of the client objects. This is a good thing to have in general in 2022, but in particular there is evidence it may unlock better throughput on batch downloading.
- [ ] https://github.com/bluesky/tiled/issues/288

### DIFF
--- a/tiled/client/array.py
+++ b/tiled/client/array.py
@@ -91,13 +91,8 @@ class DaskArrayClient(BaseStructureClient):
             expected_shape = ",".join(map(str, shape))
         else:
             expected_shape = "scalar"
-        full_path = (
-            "/array/block"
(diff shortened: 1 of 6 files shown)


danielballan:
Superseded by #346. Two items in the check list above still need to be addressed (#116 and #281) but they will be handled in separate PRs.

PR #301, opened 2022-08-19
add configurable validation registry

kleinhenz:
This adds a configurable validation registry which allows attaching a function to validate data with a given spec posted to the `/node/metadata` route.

The validation configuration consists of a new `validation` section in the config file which maps specs to importable functions.
The validation function is called as `f(metadata, structure_family, structure, spec)` and must return `metadata` (possibly modified) or else throw a `tiled.validation_registration.ValidationError` exception.

Unfortunately not really possible to test without a builtin writable tree but hopefully not too complicated. It think this is generally useful and will improve the validation in aimmdb.

### DIFF
--- a/tiled/_tests/test_validation.py
+++ b/tiled/_tests/test_validation.py
@@ -1,0 +1,64 @@
+"""
+This tests tiled's validation registry
+"""
+
+import httpx
+import numpy as np
+import pandas as pd
+import pytest
+
+from ..client import from_tree
+from ..validation_registration import ValidationError, ValidationRegistry
+from .writable_adapters import WritableMapAdapter
+
+API_KEY = "secret"
+
+
+def validate_foo(metadata, structure_family, structure, spec):
+    if structure_family != "dataframe":
+        raise ValidationError(f"structure family for spec {spec} must be dataframe")
+
+    if list(structure.macro.columns) != ["a", "b"]:
+        raise ValidationError(f"structure for spec {spec} must have columns ['a', 'b']")
+
+    if "foo" not in metadata:
+        raise ValidationError("metadata for spec {spec} must contain foo")
+
+
+def test_validators():
+
+    validation_registry = ValidationRegistry()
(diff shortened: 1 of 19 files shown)


danielballan:
Looks right from what I can see, but the new file wasn’t committed.

There is a built-in in-memory writer now, just for testing. Can you add a test or two there? https://github.com/bluesky/tiled/blob/main/tiled/_tests/test_writing.py

The `from_config` constructor will let you inject a custom validator; grep for examples.

danielballan:
A thought: currently, when the client writes, the server returns only the id (that is, the key) and some links. The Python client merges this with other information that it just submitted and therefore already knows to instantiate a new client object.

If the metadata is altered by validation/normalization the client would currently have no way to know. One option is to always return the metadata. Another option is to have the validator return `None` to indicate “validated without any changes”.

The first option is simpler, but wastes serialization time and network time in the majority of cases. Metadata can be around 0.1–1 MB is real cases. I think it might be worth adding the complexity of the second option. Validators can of course ignore this optimization (always return the dict) at a performance cost.

tiled/_tests/test_validation.py
```diff
@@ -0,0 +1,64 @@
...
+        tree,
+        api_key=API_KEY,
+        authentication={"single_user_api_key": API_KEY},
+        validation_registry=validation_registry,
+    )
+
+    with pytest.raises(httpx.HTTPStatusError) as err:
+        a = np.ones((5, 7))
+        client.write_array(a, metadata={}, specs=["foo"])
+    assert err.match("400")
```
danielballan, line 44: There is a nice way to target the status code specifically. It might be worth moving this into `_tests/utils.py` for reuse here. https://github.com/bluesky/tiled/blob/1e78c2e2b700f8092865d5aa9a5e5b8cbf65233e/tiled/_tests/test_authentication.py#L23

danielballan (commented):
Added suggestion about texts. See also question about about how to send modifications to the client.

Issue #264, opened 2022-07-26
Expose a way to query `SELECT DISTINCT`


danielballan:
## Motivation

When formulating a search query, users want to know what the possible values of a given metadata key are. This could be used to:

* Expose a Python API that lists possible values for a given key
* Populate a select box or [autocomplete UI](https://mui.com/material-ui/react-autocomplete/)
* Provide a directory-like view on the data (drilling through proposals, sample names, etc.)

@kleinhenz has prototyped the directory-like view by adding special logic to paths. This is neat. I like that it's possible to do this without changes to Tiled core. For a couple weeks I was arguing in favor of pushing this approach more generally and moving supporting code into core. However, that approach blurs the line between drilling into a given structure and filtering collections of structures in a way that may be confusing, especially for catalogs that include nested structures. It also tends to limit the `SELECT DISTINCT` queries to keys the that server maintainer has decided in advance are important.

## Proposal

Add a new route `/node/distinct/{path}?metadata=...` that provides the distinct values for one or more metadata keys. The query parameter should be named something like `metadata` so that in the future we can query for distinct specs, structure families, etc. as well.

It may also be useful to include the _count_, the number of items that have each value. This might be used to, for example, show many items are contained in each psuedo-"directory" in a directory-like view. MongoDB provides a nice way to do this: https://stackoverflow.com/questions/14924495/mongodb-count-num-of-distinct-values-per-field-key/33418582#33418582 If the count imposes significant performance overhead compared to the list of, we could in the future add an opt-out query parameter.

The MongoDB query linked about also imposes a limit of the number of distinct values collected, which I think is sensible. If a key has more than 100 or so distinct values, perhaps we should return a sample and indicate that it is such.

The response may look something roughly like:

```json
{"data":
    {"metadata": {"Element": {"Ni": 100, "Au": 30}}}
}
```

following the JSON API-like response structure used by other JSON routes.

## Rejected Alternatives

We could avoid a dedicated route, but since the query may tend to be:

* a bit expensive because it is not necessarily indexed
* highly parameterized

it does not fit naturally into our existing routes.

kleinhenz:
closed by #302

PR #302, opened 2022-08-26
WIP add /node/distinct route for querying distinct metadata

kleinhenz:
addresses #264. Still not entirely decided on the proper response model.


danielballan (commented):
Looks good.

Can you add a test case for `specs` and `structure_families`? And can you add `distinct()` to the reference docs at https://github.com/bluesky/tiled/blob/main/docs/source/reference/python-client.md#client-node? With that, I think we can merge.

danielballan (approved):
Good to merge, once CI passess

PR #307, opened 2022-09-01
add queries to search specs and structure_families

### LINKED ISSUE #305: Add a query on Specs
Add a query that filters results to nodes that satisfy given spec(s). Usage might be:

```py
Specs(includes=[...], excludes=[...])
```

kleinhenz:
adds queries to search specs and structure_families. closes #305.

### DIFF
--- a/docs/source/reference/queries.md
+++ b/docs/source/reference/queries.md
@@ -22,6 +22,8 @@ Follow the links in the table below for examples specific to each query.
    tiled.queries.In
    tiled.queries.NotIn
    tiled.queries.Regex
+   tiled.queries.Specs
+   tiled.queries.StructureFamily
 ```
 
 ## Query expressions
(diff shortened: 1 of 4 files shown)


danielballan:
Looks great! Please add these to https://github.com/bluesky/tiled/blob/main/docs/source/reference/queries.md.

Also, I wonder if it's worth adding a convenience function

```py
def Spec(spec):
    return Specs([spec])
```

so that the common use case

```py
c.search(Spec("foo"))
```

is nice and clean, as opposed to `c.search(Specs(["foo"]))`. A minor thing, but it might be nice.

Addressing some alternatives that occurred to me: I _really_ dislike the pattern where a class/function accepts a string and quietly wraps it in a list because that makes it hard to discover that it _also_ accepts lists. We could use variadic args --- `Specs(*includes, excludes)` so that `Specs("foo")` and `Specs("foo" "bar")` work, but when `excludes` comes in it will look oddly internally inconsistent: `Specs("foo", "bar", excludes=["baz"])`.

danielballan (approved):

PR #312, opened 2022-09-06
Support varargs in search().

danielballan:
This supports

```py
c.search(q1, q2, q3)
```

as equivalent to

```py
c.search(q1).search(q2).search(q3)
```

_This has no performance benefits_ because no network traffic occurs until you actually start fetching results from the search. But it does not have a succinctness benefit. And it _may_ have a readability benefit as long as the behavior is sufficiently obvious: that is, that the queries are being logically AND-ed.

### DIFF
--- a/docs/source/reference/queries.md
+++ b/docs/source/reference/queries.md
@@ -26,6 +26,13 @@ Follow the links in the table below for examples specific to each query.
 
 ## Query expressions
 
+```{eval-rst}
+.. autosummary::
+   :toctree: generated
+
+   tiled.queries.Key
+```
+
 The `Key` object can be used to construct queries in a readable way using
 standard Python comparison operators, as in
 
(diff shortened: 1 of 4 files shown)


danielballan:
This PR also documents explicitly that there is no support for logical OR, but it provides a recommendation on how to do that nicely on the client side, which I have previously suggested in Slack messages but never documented.

danielballan:
The best reason I can think of _not_ to do this is that we might want multiple arguments to mean something else, like logical OR. But:

1. It seems pretty confusing to make `c.search(q1, q2)` mean a _different_ thing that `c.search(q1).search(q2)`. If we implement it at all, it should mean the same thing.
2. If we decide to add logical OR we can add an explicit modifier like `c.search(Or(q1, q2))`.
3. For reasons discussed elsewhere, we probably do not want to ever add support for logical OR in the server anyway.

danielballan:
Initial feedback from beamline staff is that this is too potentially confusing. Feedback on `search(q1& q2 & q3)` was also negative because the syntax is comparatively obscure.

PR #321, opened 2022-09-13
API: Use varargs in distinct().

danielballan:
Seeing the demo of distinct, I realize this is a good place to use variadic arguments, changing

```py
c.distinct(["foo"])
```

to

```py
c.distinct("foo")
```

Multiple metadata keys may be passed like

```py
c.distinct("foo", "bar")
```

This side-steps the potential issue where `c.distinct("foo")` becomes `c.distinct(["f", "o", "o"])`. It's also generally more memorable and readable, especially for Python novices.

I think the downside is that it's less obvious that multiple keys _can_ be queried at once. But in this case, on balance, I think variadic args are a net win.

If needed we can do a deprecation cycle here (accepting the old way `["foo"]` and issuing a warning) but given that this is very new I think we can just tell the AIMM folks and make a backward-incompatible change.

### DIFF
--- a/tiled/_tests/test_distinct.py
+++ b/tiled/_tests/test_distinct.py
@@ -39,7 +39,7 @@ def test_distinct():
 
     # test without counts
     distinct = client.distinct(
-        metadata_keys=["foo.bar"], structure_families=True, specs=True, counts=False
+        "foo.bar", structure_families=True, specs=True, counts=False
     )
     expected = {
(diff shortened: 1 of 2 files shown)


danielballan:
@kleinhenz pointed out that users may want to query entries where a _group_ of values are distinct, as in `["foo", "bar"]`. This would provide a way to do that, though perhaps it may not be obvious enough what the behavior is. Food for future thought.

PR #322, opened 2022-09-14
improve docker

kleinhenz:
* update Dockerfile so sparse deps are cached
* don't copy whole source directory into container
* make gunicorn config configurable using GUNICORN_CONF env var
* require check that single_user_api_key is set even in public mode

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,7 +17,7 @@ WORKDIR /app
 RUN apt-get update && apt-get install -y git
 
 # copy requirements over first so this layer is cached and we don't have to reinstall deps when editing src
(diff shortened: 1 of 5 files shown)


danielballan (commented):
Thanks for refurbishing this.

My only reservation is `GUNICORN_CONF`. I wish we didn't have to introduce a new env var. I wonder if we should defer to the gunicorn env var [`GUNICORN_CMD_ARGS`](https://docs.gunicorn.org/en/stable/configure.html) which can be used to point to a custom config location (among other things).

kleinhenz:
I was following [this](https://github.com/tiangolo/uvicorn-gunicorn-docker/blob/master/docker-images/start.sh) from the tiangolo uvicorn-gunicorn docker container but happy to change if you would rather depend on the built in env vars.

danielballan:
That’s good enough for me. Just an idle thought.

dylanmcreynolds (commented):
Seems solid.

PR #326, opened 2022-09-21
Deprecate Specs query. Add Spec query.

danielballan:
Upon reflection, I think my initiate suggestion for `Specs` in #305 was overbuilt. The most common use case will be to look for one spec, as in:

```py
c.search(Spec("XDI"))
```

Typically, because it seems that specs tend to follow an "inheritance" pattern of increasing specificity, one spec will often be enough. But, when the need arises, multiple specs can be filtered the same way we would do multiple rounds of filtering (logical AND) with Tiled queries generally, by chaining calls to `search`:

```py
c.search(Spec("XDI")).search(Spec("HasBatteryChargeMetadata"))
```

The behavior is completely unambiguous. The original proposal provided a way to do this that was more succinct

```py
c.search(Specs(["XDI", "HasBatteryChargeMetadata"]))
```

but its behavior was ambiguous. Does this match items that have both specs or just one? Also, as a minor point, in the situation where you want just one spec `Specs(["XDI"])` with the `[]` is a little ungainly.

The original proposal also provides an `excludes` option, but it's not yet shown that this would really be used.

Therefore, in this PR, I deprecate `Specs` and un-document it. I add `Spec` as a first-class query (not just a wrapper function around `Specs`). Let's see how it goes. If we need `Specs` we can reinstate it easily.

### DIFF
--- a/docs/source/reference/queries.md
+++ b/docs/source/reference/queries.md
@@ -23,7 +23,6 @@ Follow the links in the table below for examples specific to each query.
    tiled.queries.NotIn
    tiled.queries.Regex
    tiled.queries.Spec
-   tiled.queries.Specs
    tiled.queries.StructureFamily
 ```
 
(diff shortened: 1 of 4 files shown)


danielballan:
I think I’ll let this go.

PR #327, opened 2022-09-21
Require specs be declared, but validation optional.

danielballan:
Backward-incompatible change to config from

```yaml
validators:
  my-spec: validators:validate_my_spec
```

to

```yaml
specs:
- spec: my-spec
  validator: validators:validate_my_spec
```

where `validator:` is now optional.

The server rejects uploads with an unrecognized spec. Thus, you must at least _declare_ the spec in the server configuration to support client upload, but you don't have to write a validator. This will catch spelling mistakes and prohibit a client from exploding the spec namespace without consulting the server admin.

Note that Adapters can still declare specs (from the server side) without declaration. This only affects client upload.

### DIFF
--- a/tiled/config.py
+++ b/tiled/config.py
@@ -195,9 +195,13 @@ def construct_build_app_kwargs(
         for ext, media_type in config.get("file_extensions", {}).items():
             serialization_registry.register_alias(ext, media_type)
 
-        for spec, validator_path in config.get("validation", {}).items():
-            validator = import_object(validator_path)
-            validation_registry.register(spec, validator)
+        for item in config.get("specs", []):
+            if "validator" in item:
+                validator = import_object(item["validator"])
+            else:
+                # no-op
+                validator = _no_op_validator
+            validation_registry.register(item["spec"], validator)
 
     # TODO Make compression_registry extensible via configuration.
     return {
@@ -450,3 +454,7 @@ def direct_access_from_profile(name):
 
 class ConfigError(ValueError):
     pass
+
+
+def _no_op_validator(*args, **kwargs):
+    return None
--- a/tiled/server/router.py
+++ b/tiled/server/router.py
@@ -617,27 +617,29 @@ def post_metadata(
         body.references,
     )
 
+    metadata_modified = False
+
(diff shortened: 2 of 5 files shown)


tiled/server/router.py
```diff
@@ -779,21 +781,28 @@ async def put_metadata(
...
-                if result is not None:
-                    metadata_modified = True
-                    metadata = result
-
-            except ValidationError as e:
-                raise HTTPException(
-                    status_code=400,
-                    detail=f"failed validation for spec {spec}:\n{e}",
-                )
+    # Specs should be ordered from most specific/constrained to least.
```
dylanmcreynolds, line 784: Will it be obvious to the caller that if submit ["spec1", "spec2"] the result might be different from ["spec2", "spec1"]?

tiled/config.py
```diff
@@ -195,9 +195,13 @@ def construct_build_app_kwargs(
         for ext, media_type in config.get("file_extensions", {}).items():
             serialization_registry.register_alias(ext, media_type)
 
-        for spec, validator_path in config.get("validation", {}).items():
-            validator = import_object(validator_path)
-            validation_registry.register(spec, validator)
+        for item in config.get("specs", []):
```
dylanmcreynolds, line 198: Was this change necessary because you want to be able to support more configuration with specs in the future?

danielballan: Yes, also, just low this comment, notice that validation becomes optional. If the key new `validator` is omitted, then a no-op function is used.

tiled/server/router.py
danielballan: Probably not obvious, but I can't think of a better alternative.

The idea that specs should go from specific to general has been in there from the beginning. The first practical consequence is that, if there is a special client (e.g. the `BlueskyRun` client object) for multiple specs the earliest spec in the list gets precedence.

The second practical consequence is that we have to choose _some_ order for the validators to run in.

tiled/config.py
dylanmcreynolds: That makes sense. Thanks.

dylanmcreynolds (approved):

PR #332, opened 2022-10-05
Support chunked-encoding streaming export.

### LINKED ISSUE #70: Support chunked responses
Currently, the media type serialization registry contains _functions_. It should also accept _generator functions_ that yield the data in chunks. This may reduce the server's memory usage, and it will enable us to start servicing a request faster.

1. Teach this codepath that it might receives bytes (current behavior) or a generator instance (new behavior). https://github.com/bluesky/tiled/blob/70a8e1b687c27a52c37ee7aabaaf6da5d2bfc248/tiled/server/core.py#L425
   We can check whether the return value is a generator using `isinstance(content, types.GeneratorType)`. If it's a generator, return a `PatchedStreamingResponse` instead of a `PatchedResponse`. https://github.com/bluesky/tiled/blob/70a8e1b687c27a52c37ee7aabaaf6da5d2bfc248/tiled/server/core.py#L437-L441
2. Convert some of the existing serializers to generator functions. This may be done by taking the existing functions like `serialize_csv` and building generators that wrap them, as illustrated in my comment below. Then, the function version becomes an internal utility and the generator version takes its place in the registry here: https://github.com/bluesky/tiled/blob/70a8e1b687c27a52c37ee7aabaaf6da5d2bfc248/tiled/structures/array.py#L177-L178

danielballan:
Closes #70 

This extends the export functionality to support _generators_ that yield a chunk of bytes at a time. It applies this feature to the newline-delimited JSON exporter (`application/json-seq`). It is a really strong fit for that streaming-oriented format. In future PRs we can explore applying it to other formats.

Demo below. Notice `transfer-encoding: chunked` in the response header. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding.

```
$ http ":8000/api/node/full/short_table?format=application/json-seq"
HTTP/1.1 200 OK
content-type: application/json-seq
date: Wed, 05 Oct 2022 16:35:18 GMT
etag: c8adf2361038a9c936ab5d916bb219dd
server: uvicorn
server-timing: read;dur=0.5, tok;dur=0.1, pack;dur=0.2, app;dur=6.1
set-cookie: tiled_csrf=CFuHBOsYkdEnRoTvVCwG4JmfOHBDPnJytzAthdlG-FY; HttpOnly; Path=/; SameSite=lax
transfer-encoding: chunked

{"A":0.8393166285564699,"B":1.6786332571129399,"C":2.5179498856694096}
{"A":0.5688795868599656,"B":1.1377591737199313,"C":1.706638760579897}
{"A":0.6336226887581975,"B":1.267245377516395,"C":1.9008680662745925}
{"A":0.16052909847062302,"B":0.32105819694124604,"C":0.48158729541186907}
{"A":0.9191461227402795,"B":1.838292245480559,"C":2.7574383682208383}
{"A":0.8768191851466749,"B":1.7536383702933498,"C":2.6304575554400245}
{"A":0.38387163231094434,"B":0.7677432646218887,"C":1.151614896932833}
{"A":0.5438938247574789,"B":1.0877876495149579,"C":1.6316814742724368}
{"A":0.9415108546995818,"B":1.8830217093991637,"C":2.8245325640987455}
{"A":0.7571506359189891,"B":1.5143012718379782,"C":2.2714519077569673}
```

### DIFF
--- a/docs/source/how-to/custom-export-formats.md
+++ b/docs/source/how-to/custom-export-formats.md
@@ -216,6 +216,20 @@ In that case, it must be given as a MIME type, in accordance with the standard.
 The file extension alias is not accepted.
 ```
 
+## Advanced: Streaming export
+
+HTTP supports chunked responses, where data is streamed incrementally. This
+is a good fit for streaming-oriented formats such as newline-delimited JSON.
+
+To create a chunked exporter, implement your exporter as a Python generator
+that yields bytes.
+
+```python
+def export(array, metadata):
+    for ... in ...:
+        yield b"..."
+```
+
 ## Further examples
 
 At the bottom of each of the modules in `tiled/structures`, you will
--- a/tiled/_tests/test_export.py
+++ b/tiled/_tests/test_export.py
@@ -33,6 +33,9 @@ tree = MapAdapter(
             ),
             npartitions=3,
         ),
+        "empty_table": DataFrameAdapter.from_pandas(
+            pandas.DataFrame({"A": []}), npartitions=1
+        ),
         "structured_data": MapAdapter(
             {
                 "pets": ArrayAdapter.from_array(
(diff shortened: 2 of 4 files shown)


dylanmcreynolds (commented):
Exciting. I don't really have any comments other than that I can people this for more structure families.

dylanmcreynolds (approved):

Issue #325, opened 2022-09-20
Review unit tests for speed


danielballan:
The unit tests have gotten slower, locally, and _much_ slower on GitHub. It seems that we may be surpassing some threshold where GH Actions begins throttling us and making are slow tests even slower.

Let's look at speeding up the tests by:

* Using server instances shared across a whole module where the downside of coupled tests is low
* Reducing example data size

danielballan:
I learned that pytest provides [a CLI option](https://docs.pytest.org/en/latest/how-to/usage.html#profiling-test-execution-duration) for profiling test speed. This command shows all tests that take longer than one second to run, sorted from longest to shortest:

```
pytest --durations=0 --durations-min=1.0
```

These are the results on my local machine. Presumably results on CI are even slower, especially for the tests that involve I/O from disk.

```
98.32s call     tiled/_tests/test_xarray.py::test_wide_table_optimization_off
45.53s call     tiled/_tests/test_xarray.py::test_url_limit_handling
34.73s call     tiled/_tests/test_writing.py::test_write_large_array_full
27.71s call     tiled/_tests/test_xarray.py::test_dataset_column_access[image]
19.58s call     tiled/_tests/test_tiff.py::test_tiff_sequence[slice_input2-correct_shape2]
17.71s call     tiled/_tests/test_authentication.py::test_session_limit
15.35s call     tiled/_tests/test_sort.py::test_sort[letter-sorted_list0]
13.86s call     tiled/_tests/test_authentication.py::test_api_keys
13.46s call     tiled/_tests/test_object_cache.py::test_detect_content_changed_or_removed
12.86s call     tiled/_tests/test_queries.py::test_structure_families
11.51s call     tiled/_tests/test_hdf5.py::test_from_file
11.46s call     tiled/_tests/test_tiff.py::test_tiff_sequence_with_directory_walker
10.86s call     tiled/_tests/test_xarray.py::test_xarray_dataset[ragged]
10.44s call     tiled/_tests/test_writing.py::test_write_array_chunked
10.24s call     tiled/_tests/test_xarray.py::test_dataset_column_access[weather]
10.05s call     tiled/_tests/test_export.py::test_export_table[spreadsheet.xlsx-numpy]
9.38s call     tiled/_tests/test_tiff.py::test_tiff_sequence[slice_input4-correct_shape4]
8.99s call     tiled/_tests/test_xarray.py::test_xarray_dataset[image]
[...]

PR #333, opened 2022-10-05
Faster tests

danielballan:
This should make some progress on #325. Strategies:

* Switch some (not all) file-export tests to write to in-memory buffers instead of disks, which can be very slow on CI workers.
* Use smaller example data sets.
* In place we artificially reduce limits and then test that we hit them. We can reduce them more.
* Reuse the same `app` and `client` objects where possible---where there is no coupling between tests.


danielballan:
Tests run in 9 minutes locally now.

jmaruland (approved):

PR #339, opened 2022-10-07
missing dependencies in docker build

dylanmcreynolds:
Trying to build the docker image locally ran into a few challenges with missing dependencies. I also added an example comment of how to build the docker locally.


danielballan:
Test failures are flaky and unrelated.

PR #346, opened 2022-10-18
Refactor authentication

### LINKED ISSUE #145: If there is a cached refresh token, reauthenticate automatically.
When a server has optional authentication, the client does not currently automatically reauthenticate. It should.

### LINKED ISSUE #288: Remove _async_bridge module

danielballan:
Major refactor:
* Use `httpx.Auth` to inject refresh flow via coroutine.
* Store tokens on disk split out by both domain (`netloc`, to be precise) and provider, username.
* Store tokens in a tiered cache, which always comprises a dict in memory and a directory on disk.
* Store `access_token` as well as `refresh_token` and do not automatically refresh when a client is created. If the `access_token` is still valid, use it.
* `Context.__init__` only interacts with _public_ data (the server info at `/api/`). It does not attempt authentication. The constructor convenience methods `from_*` do the authentication, if given an `api_key` and/or a `username`.
* When given a `uri` to a specific node, like `https://.../api/node/metadata/x/y/z/`, we have been use an unusual trick to find our way to the root `/api/` route. We now just Look for `/node/metadata` in the string, and we prohibit servers to be launched at a prefix that contains `/node/metadata`.
* ~Enable Tiled to be launched at a sub-prefix with the new top-level config item, `prefix`.~ [Moved into a separate branch.]
* Remove the janky `_async_bridge` module and replace it with a vendor of (not yet relased) code in `starlette`.
* Prompt for authentication only at construction time, never in the middle of a computation. (Raise instead.)
* Detect whether the stdin is a TTY, and do not prompt if it is not. (Raise instead.)
* Remove vestigial `path` state from client objects. Use `item["links"]` instead.
* Move cached tokens from `~/.config/tiled/tokens` to `~/.cache/tiled/tokens`. This is especially helpful if `~/.config/tiled` is version-controlled. (You don't have to remember to `gitignore` the tokens.)
* Add CLI-specific state in `~/.config/tiled/cli` to store (1) the currently "active" / default tiled server to use and (2) which username/provider to authenticate with for each server, if there is an active session. Improve the CLI UX.

Closes #288 
Closes #145


danielballan:
To do:

- [ ] Work out issue with UI when the server is served from a prefix.
- [ ] Revisit Tiled CLI, now that tokens are split out by provider and username. Likely we want a notion of the current context (as `kubectl` has, perhaps a file at `~/.config/tiled/current_context` with a profile name.

danielballan:
Alright. I don't really have anyone sufficient "read in" to review this, so I'm going to "YOLO" merge it.

@dylanmcreynolds and @tacaswell have both been subjected to a high-level overview, but bear no responsibility for whatever mess I make here. :-D

PR #347, opened 2022-10-21
Support serving Tiled from a prefix

### LINKED ISSUE #283: Support serving Tiled from a sub-path.

danielballan:
Closes #283 

- [ ] This needs a rebase. It was branched off of #346 part way through.
- [ ] What to do about the React UI? The API route is set at build time.

### DIFF
--- a/example_configs/toy_authentication.yml
+++ b/example_configs/toy_authentication.yml
@@ -30,3 +30,4 @@ access_control:
 trees:
   - path: /
     tree: tiled.examples.toy_authentication:tree
+prefix: /a/b/c
--- a/tiled/_tests/test_client.py
+++ b/tiled/_tests/test_client.py
@@ -11,3 +11,8 @@ def test_configurable_timeout():
     assert c.context.http_client.timeout.connect != 17
     c = from_tree(tree, timeout=httpx.Timeout(17))
     assert c.context.http_client.timeout.connect == 17
+
+
+def test_prefix():
+    c = from_tree(tree, server_settings={"prefix": "/a/b/c"})
+    list(c)
--- a/tiled/_tests/test_client.py
+++ b/tiled/_tests/test_client.py
@@ -1,4 +1,5 @@
 import httpx
+import pytest
 
 from ..adapters.mapping import MapAdapter
 from ..client import from_tree
@@ -16,3 +17,7 @@ def test_configurable_timeout():
 def test_prefix():
     c = from_tree(tree, server_settings={"prefix": "/a/b/c"})
     list(c)
+
+    # Not allowed to use '/node/metadata' in the prefix.
+    with pytest.raises(Exception):
+        from_tree(tree, server_settings={"prefix": "/a/b/node/metadata/c"})
--- a/tiled/client/constructors.py
+++ b/tiled/client/constructors.py
@@ -170,7 +170,7 @@ def from_tree(
         settings = app.dependency_overrides[get_settings]()
         api_key = settings.single_user_api_key or None
     context = Context(
-        uri="http://local-tiled-app/api",
+        uri=f"http://local-tiled-app{(server_settings or {}).get('prefix', '')}/api",
         headers=headers,
         api_key=api_key,
         cache=cache,
@@ -418,7 +418,7 @@ def from_config(
         settings = app.dependency_overrides[get_settings]()
         api_key = settings.single_user_api_key or None
     context = Context(
-        uri="http://local-tiled-app/api",
+        uri=f"http://local-tiled-app{config.get('prefix', '')}/api",
         headers=headers,
         api_key=api_key,
         cache=cache,
--- a/tiled/commandline/main.py
+++ b/tiled/commandline/main.py
@@ -460,6 +460,7 @@ def serve_directory(
         ),
     ),
     port: int = typer.Option(8000, help="Bind to a socket with this port."),
+    prefix: Optional[str] = typer.Option(None, help="Serve Tiled on a sub-path."),
     object_cache_available_bytes: Optional[float] = typer.Option(
         None,
         "--data-cache",
@@ -476,7 +477,7 @@ def serve_directory(
     from ..server.app import build_app, print_admin_api_key_if_generated
 
     tree_kwargs = {}
-    server_settings = {}
+    server_settings = {"prefix": prefix or ""}
     if keep_ext:
         from ..adapters.files import identity
 
@@ -520,6 +521,7 @@ def serve_pyobject(
         ),
     ),
     port: int = typer.Option(8000, help="Bind to a socket with this port."),
+    prefix: Optional[str] = typer.Option(None, help="Serve Tiled on a sub-path."),
     object_cache_available_bytes: Optional[float] = typer.Option(
         None,
         "--data-cache",
@@ -536,7 +538,7 @@ def serve_pyobject(
     from ..utils import import_object
 
     tree = import_object(object_path)
-    server_settings = {}
+    server_settings = {"prefix": prefix or ""}
     if object_cache_available_bytes is not None:
         server_settings["object_cache"] = {}
         server_settings["object_cache"][
@@ -580,6 +582,7 @@ def serve_config(
     port: int = typer.Option(
         None, help="Bind to a socket with this port. Uses value in config by default."
     ),
+    prefix: Optional[str] = typer.Option(None, help="Serve Tiled on a sub-path."),
 ):
     "Serve a Tree as specified in configuration file(s)."
     import os
@@ -598,6 +601,10 @@ def serve_config(
         if "authentication" not in parsed_config:
             parsed_config["authentication"] = {}
         parsed_config["authentication"]["allow_anonymous_access"] = True
+
+    # Let prefix option override config.
+    if prefix is not None:
+        parsed_config["prefix"] = prefix
 
     # Delay this import so that we can fail faster if config-parsing fails above.
 
--- a/tiled/config.py
+++ b/tiled/config.py
@@ -162,6 +162,7 @@ def construct_build_app_kwargs(
             root_tree = MapAdapter(root_mapping, access_policy=root_access_policy)
             root_tree.include_routers.extend(include_routers)
         server_settings = {}
+        server_settings["prefix"] = config.get("prefix", "")
         server_settings["allow_origins"] = config.get("allow_origins")
         server_settings["object_cache"] = config.get("object_cache", {})
         server_settings["response_bytesize_limit"] = config.get(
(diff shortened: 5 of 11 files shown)


example_configs/toy_authentication.yml
```diff
@@ -30,3 +30,4 @@ access_control:
 trees:
   - path: /
     tree: tiled.examples.toy_authentication:tree
+prefix: /a/b/c
```
dylanmcreynolds, line 33: What do you think of `path_prefex` to make it a little clearer?

pshafer-als: <butting in here with unsolicited opinion...>  :-)

Aren't there several "paths" involved also?  The public path (URL) the backend path where the data resides (file path), and maybe more?

Maybe `url_path_prefix` or `url_prefix` or `http_root_path` helps distinguish without being too long?

danielballan: Yeah, agreed `prefix` is too overloaded to stand. Will change to one of these.

pshafer-als:
I think it will also be necessary to update the path to the Swagger docs, so that they also use the prefix.

https://github.com/bluesky/tiled/blob/489c63926aa2b562daa95f6991d6a268aa52fb39/tiled/server/app.py#L136

would become...
```
    app = FastAPI(docs_url=f"{global_prefix}/docs")
```

Rather than manually adding the prefix in all these locations though, would using the [`root_path` argument](https://fastapi.tiangolo.com/advanced/behind-a-proxy/?h=root+path#setting-the-root_path-in-the-fastapi-app) to FastAPI work?  I've used that option successfully for paths that are directly controlled by FastAPI as well as WSGI subapps like Dash (Flask) that get mounted into a FastAPI app.

danielballan:
This got stuck because it seemed like we had to know the API route (including the prefix) at UI _build_ time, which is awkward. But what if the UI fetched its API route from the front-end config, which we are currently using to describe specs and formats? The default would be as it is now (a relative path), but it could be overridden by config to be an absolute URL with the prefix. I think that would work but haven't looked at the code yet.

tacaswell:
This is a partial attempt to do the rebase "by hand" (asking git to do it goes very very badly), but given that there is already a `root_path` in the uvicorn config that is backed in all the way at the bottom, do we need a second layer of configurable prefix?

I got stuck because a couple places in the authenticator code we need to have access to the settings that I don't quickly see how to thread back through.

```diff
diff --git a/example_configs/toy_authentication.yml b/example_configs/toy_authentication.yml
index 6f96d50e..b8bf4ca3 100644
--- a/example_configs/toy_authentication.yml
+++ b/example_configs/toy_authentication.yml
@@ -31,3 +31,4 @@ access_control:
 trees:
   - path: /
     tree: tiled.examples.toy_authentication:tree
+    prefix: /a/b/c
diff --git a/tiled/_tests/test_client.py b/tiled/_tests/test_client.py
index 11b3cdaf..ce42a7fd 100644
--- a/tiled/_tests/test_client.py
+++ b/tiled/_tests/test_client.py
@@ -151,3 +151,9 @@ def test_jump_down_tree():
     with record_history() as h:
         client["e"]["d"]["c"]["b"]["a"]
     assert len(h.requests) == 5
+
+
+def test_prefix():
+    with Context.from_app(build_app(tree)) as context:
+        c = from_context(context, server_settings={"prefix": "/a/b/c"})
+    list(c)
diff --git a/tiled/config.py b/tiled/config.py
index 22d76dce..93f8aedd 100644
--- a/tiled/config.py
+++ b/tiled/config.py
@@ -183,6 +183,7 @@ See documentation section "Serve a Directory of Files"."""
             root_tree = MapAdapter(root_mapping, access_policy=root_access_policy)
             root_tree.include_routers.extend(include_routers)
         server_settings = {}
+        server_settings["prefix"] = config.get("prefix", '')
         server_settings["allow_origins"] = config.get("allow_origins")
         server_settings["object_cache"] = config.get("object_cache", {})
         server_settings["response_bytesize_limit"] = config.get(
@@ -235,6 +236,7 @@ def merge(configs):
     authentication_config_source = None
     access_control_config_source = None
     uvicorn_config_source = None
+    prefix_config_source = None
     object_cache_config_source = None
     metrics_config_source = None
     database_config_source = None
@@ -279,6 +281,16 @@ def merge(configs):
                 )
             uvicorn_config_source = filepath
             merged["uvicorn"] = config["uvicorn"]
+
+        if "prefix" in config:
+            if "prefix" in merged:
+                raise ConfigError(
+                    "prefix can only be specified in one file. "
+                    f"It was found in both {prefix_config_source} and "
+                    f"{filepath}"
+                )
+            prefix_config_source = filepath
+            merged["prefix"] = config["prefix"]
         if "object_cache" in config:
             if "object_cache" in merged:
                 raise ConfigError(
diff --git a/tiled/config_schemas/service_configuration.yml b/tiled/config_schemas/service_configuration.yml
index 909c623b..d45274e1 100644
--- a/tiled/config_schemas/service_configuration.yml
+++ b/tiled/config_schemas/service_configuration.yml
@@ -387,6 +387,12 @@ properties:
         from Mozilla's web developer documentation.
 
         https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+   prefix:
+     type: string
+     description: |
+       Serve the Tiled application on a sub-path. By default, Tiled is served at
+       `/`. If this is set to a path like `/a/b/c`, it will be served at `/a/b/c/`.
+       This must have a leading slash and it must _not_ have a trailing slash.
   uvicorn:
     type: object
     properties:
diff --git a/tiled/server/app.py b/tiled/server/app.py
index 0c3ac7a7..d44bb9bd 100644
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -124,6 +124,7 @@ def build_app(
         for spec in authentication.get("providers", [])
     }
     server_settings = server_settings or {}
+    global_prefix = server_settings.get("prefix", "")
     query_registry = query_registry or get_query_registry()
     compression_registry = compression_registry or default_compression_registry
     validation_registry = validation_registry or default_validation_registry
@@ -247,7 +248,7 @@ or via the environment variable TILED_SINGLE_USER_API_KEY.""",
                 {
                     "request": request,
                     # This is used to construct the link to the React UI.
-                    "root_url": get_root_url(request),
+                    "root_url": get_root_url(request, global_prefix),
                     # If defined, this adds a Binder link to the page.
                     "binder_link": os.getenv("TILED_BINDER_LINK"),
                 },
@@ -298,14 +299,14 @@ or via the environment variable TILED_SINGLE_USER_API_KEY.""",
         allow_headers=["*"],
     )
 
-    app.include_router(router, prefix="/api/v1")
+    app.include_router(router, prefix=f"{global_prefix}/api/v1")
 
     # The Tree and Authenticator have the opportunity to add custom routes to
     # the server here. (Just for example, a Tree of BlueskyRuns uses this
     # hook to add a /documents route.) This has to be done before dependency_overrides
     # are processed, so we cannot just inject this configuration via Depends.
     for custom_router in getattr(tree, "include_routers", []):
-        app.include_router(custom_router, prefix="/api/v1")
+        app.include_router(custom_router, prefix=f"{global_prefix}/api/v1")
 
     if authentication.get("providers", []):
         # Delay this imports to avoid delaying startup with the SQL and cryptography
@@ -371,7 +372,7 @@ or via the environment variable TILED_SINGLE_USER_API_KEY.""",
                     custom_router, prefix=f"/provider/{provider}"
                 )
         # And add this authentication_router itself to the app.
-        app.include_router(authentication_router, prefix="/api/v1/auth")
+        app.include_router(authentication_router, prefix=f"{global_prefix}/api/v1/auth")
 
     # The /search route is defined after import time so that the user has the
     # opporunity to register custom query types before startup.
@@ -816,7 +817,7 @@ Back up the database, and then run:
 
         from . import metrics
 
-        app.include_router(metrics.router, prefix="/api/v1")
+        app.include_router(metrics.router, prefix=f"{global_prefix}/api/v1")
 
         @app.middleware("http")
         async def capture_metrics_prometheus(request: Request, call_next):
diff --git a/tiled/server/authentication.py b/tiled/server/authentication.py
index 0d30e982..7e951e76 100644
--- a/tiled/server/authentication.py
+++ b/tiled/server/authentication.py
@@ -194,7 +194,7 @@ async def get_api_key(
     return None
 
 
-def headers_for_401(request: Request, security_scopes: SecurityScopes):
+def headers_for_401(request: Request, security_scopes: SecurityScopes, settings: BaseSettings):
     # call directly from methods, rather than as a dependency, to avoid calling
     # when not needed.
     if security_scopes.scopes:
@@ -203,7 +203,7 @@ def headers_for_401(request: Request, security_scopes: SecurityScopes):
         authenticate_value = "Bearer"
     headers_for_401 = {
         "WWW-Authenticate": authenticate_value,
-        "X-Tiled-Root": get_base_url(request),
+        "X-Tiled-Root": get_base_url(request, settings.prefix),
     }
     return headers_for_401
 
@@ -222,7 +222,7 @@ async def get_decoded_access_token(
         raise HTTPException(
             status_code=401,
             detail="Access token has expired. Refresh token.",
-            headers=headers_for_401(request, security_scopes),
+            headers=headers_for_401(request, security_scopes, settings),
         )
     return payload
 
diff --git a/tiled/server/utils.py b/tiled/server/utils.py
index 0c4368d0..17486227 100644
--- a/tiled/server/utils.py
+++ b/tiled/server/utils.py
@@ -27,21 +27,21 @@ def record_timing(metrics, key):
     metrics[key]["dur"] += time.perf_counter() - t0  # Units: seconds
 
 
-def get_root_url(request):
+def get_root_url(request, global_prefix):
     """
     URL at which the app is being server, including API and UI
     """
-    return f"{get_root_url_low_level(request.headers, request.scope)}"
+    return f"{get_root_url_low_level(request.headers, request.scope, global_prefix)}"
 
 
-def get_base_url(request):
+def get_base_url(request, global_prefix):
     """
     Base URL for the API
     """
-    return f"{get_root_url(request)}/api/v1"
+    return f"{get_root_url(request, global_prefix)}/api/v1"
 
 
-def get_root_url_low_level(request_headers, scope):
+def get_root_url_low_level(request_headers, scope, global_prefix):
     # We want to get the scheme, host, and root_path (if any)
     # *as it appears to the client* for use in assembling links to
     # include in our responses.
@@ -66,7 +66,7 @@ def get_root_url_low_level(request_headers, scope):
     root_path = scope.get("root_path", "")
     if root_path.endswith("/"):
         root_path = root_path[:-1]
-    return f"{scheme}://{host}{root_path}"
+    return f"{scheme}://{host}{root_path}{global_prefix}"
 
 
 def filter_for_access(entry, principal, scopes, metrics):

```

danielballan:
Superseded by the much simpler https://github.com/bluesky/tiled/pull/646

PR #348, opened 2022-10-21
First pass at making client pickleable

### LINKED ISSUE #340: Make client objects pickle-able
To facilitate multi-process workflows, client objects should be pickle-able. This will involve revisiting `tiled.client.constructors` and `tiled.client.context`.

danielballan:
Closes #340 

- [x] Context is pickleable
- [ ] Client objects are pickleable

To do this right, we should imitate `SerializableLock`, using a registry of weak references to ensure that if a client is moved out of process and then moved _back_ we can reuse the same HTTP connection pool.


danielballan:
This doesn't quite work with `dask.distributed` yet, which is the driving motivation, so it needs more work.

danielballan:
This is where I left off:

```py
c = from_uri("http://tiled-demo.blueskyproject.io/api", "dask")
import distributed
dask_client = distributed.Client()
result = c['generated']['big_image'].read().sum()
result
result.compute()
```

```
TypeError: ('Could not serialize object of type tuple', "(<bound method DaskArrayClient._get_block of <DaskArrayClient shape=(10000, 10000) chunks=((2500, 2500, 2500, 2500), (2500, 2500, 2500, 2500)) dtype=float64>>, (3, 1), dtype('float64'), (2500, 2500))")
```

PR #349, opened 2022-10-21
Debug cleanup on Python 3.9.

danielballan:
The CI has been often hanging at the end of the `pytest` run on Python 3.9. I suspect it's related to the changes in the client.

### DIFF
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -95,8 +95,6 @@ class Context:
                 follow_redirects=True,
             )
         else:
-            import atexit
-
             from ._testclient import TestClient
 
             # verify parameter is dropped, as there is no SSL in ASGI mode
@@ -108,7 +106,7 @@ class Context:
             )
             client.follow_redirects = True
             client.__enter__()
-            atexit.register(client.__exit__)
+            # TODO Is there a reasonable way to clean this up?
 
         self.http_client = client
         self._cache = cache


danielballan:
I can reproduce this hanging on Python 3.8.5 locally:

```
c
In [2]: c = from_profile("bmm_direct")
OBJECT CACHE: Will use up to 6_290_642_534 bytes (15% of total physical RAM)

In [3]: c
Out[3]: <Catalog {22521, 22524, 22525, 22526, 22528, 22529, 22534, ...} ~123 entries>

In [4]: del c

In [5]: del _3

In [6]:                                                                                                                                                                                       
Do you really want to exit ([y]/n)? y
<HANGS HERE UNTIL I INTERVENE WITH ^C>
^CError in atexit._run_exitfuncs:
  File "/home/dallan/miniconda3/envs/py38/lib/python3.8/concurrent/futures/thread.py", line 40, in _python_exit
    t.join()
  File "/home/dallan/miniconda3/envs/py38/lib/python3.8/threading.py", line 1011, in join
    self._wait_for_tstate_lock()
  File "/home/dallan/miniconda3/envs/py38/lib/python3.8/threading.py", line 1027, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt
```

danielballan:
To be clear the above is on this branch.

Very likely the problem is with the `portal` failing to shut down: https://github.com/bluesky/tiled/blob/main/tiled/client/_testclient.py

danielballan:
The best fix may be to refactor all the tests to use context managers. Not a small task, but probably a good idea.

PR #350, opened 2022-10-24
Front End Authentication

dylanmcreynolds:
This PR establishes the framework for the tiled web-frontend to be able to authenticate. 

This is highly experimental at this point. We want the ability to support a variety of authentication methods like HTTP Basic and Oath2. A configuration mechanism will have to be developed that lets users indicate which methods they want and, if OAuth2, information about the IdP servers being connected to.

At a high level, we want to support the following functionality:

- [ ] Store the current refresh token in localStorage, adding it to each message by creating an axios interceptor and plugging it into the react app.
- [ ] Create a user interface for authentication
- [ ] Create a react context provider for updating the currently logged in user in the UI, and presenting the user with the opportunity to authenticate when appropriate.

A few details about the initial commit:
- Changes the server to return all information about tokens (refresh, access and their expiration times) and removes the ability to redirect POST, which probably doesn't work anyway
- Adds `auth.tsx`, a UI for authentication. Currently only supports username/password (HTTP Basic) but will hopefully grow the ability to support multiple OAuth2 providers through configuration
- Adds a `user.tsx` which holds code related to a context provider for logged in user. This context provider will update when the user logs in/out.
- Adds a route for login.
- Adds an axios interceptor to `client.ts`. This interceptor stores token information in localStorage when the user successfully logs in, and add the current refresh token from local storage and adds to the headers of each message.


danielballan:
I rebased this on `main` and force-pushed.

danielballan:
Superceded by #1214

PR #356, opened 2022-11-02
Enrich specs.

### LINKED ISSUE #323: Enrich Specs
Specs are currently given as a list of strings, like

```json
["XDI", "HasBatteryChargeData"]
```

I propose to promote them to a list of objects with one required field (the name, as now) and some optional ones. A minimal spec might look like

```json
[{"spec": "XDI"}, {"spec": "HasBatteryChargeData"}]
```

Below are some additional optional fields we may want.

## Version

This would refer to the version of the standard or schema. Validators would have access to the version and be able to switch their behavior on it.

```json
{"spec": "XDI", "version": 1}
```

Alternatively, we could do this in-band, as in `XDI_1` or some such thing, but doing it out-of-band has advantages:

* straightforward version comparisons (e.g. `version > N`)
* simple dispatch to validators and custom clients based on the just the `spec`, leaving version concerns as an internal detail

@maffettone has emphasizes that versioning schemas is an important lesson learned. I'm 50/50 on whether we should _require_ version but I think we should at least support it.

## Location

In BMM's Bluesky metadata, XDI is stored under a subkey (I think it's just `"XDI"`). I could imagine various data providers placing the same standard metadata in various subsidiary keys. It may be useful to indicate where the specified metadata "starts".

```json
{"spec": "XDI", "loc": "start.XDI"}
```

where the value of `"loc"` could be standardized to be a [JMESPath](https://jmespath.org/), which we already use for optionally filtering metadata in Tiled.

## Others?

Of course, once we take the leap from string to object, it's easy to add other fields in the future.

## Downsides

This makes specs more complex and verbose. In the Python API we can ameliorate this by using a dataclass `Spec` with optional arguments:

```py
Adapter(
    ...,
    specs=[Spec("BlueskyRun"), Spec("XDI", version=1, loc="start.XDI")],
)
```

danielballan:
Closes #323 

This change a spec from a simple `str` to an object with a `name` (a `str`) and an optional `version` (`str` or `None`).

```python
In [3]: c
Out[3]: <Catalog {22521, 22524, 22525, 22526, 22528, 22529, 22534, ...} ~123 entries>

In [4]: c.specs
Out[4]: [Spec('CatalogOfBlueskyRuns')]

In [5]: c[22521].specs
Out[5]: [Spec('BlueskyRun')]

In [6]: c[22521].specs[0]
Out[6]: Spec('BlueskyRun')

In [7]: c[22521].specs[0].name
Out[7]: 'BlueskyRun'

In [8]: c[22521].specs[0].version
```


maffettone:
Big love for this from a sustainability and compatibility standpoint.

PR #382, opened 2023-01-31
Switch to single-process (uvicorn) container

### LINKED ISSUE #282: Revisit Dockerfile
We cannot simply _use_ the FastAPI image because it lacks Tiled and Tiled’s dependencies, but would it make sense to build on it instead of building the gunicorn stack from scratch?

https://github.com/bluesky/tiled/blob/18f94198eb6f7cf77da636b7152c25e5d47b3ad7/docker/Dockerfile#L1

danielballan:
Closes #282 

* Remove `gunicorn` from the containerized deployment
* Turns Prometheus metrics on by default
* Makes the multiprocess mechanism in Prometheus (the memory-mapped files) optional, off by default
* Simplifies `Dockerfile`
* Moves the code that checks for scalable config settings out of a one-off script and into the CLI as `--scalable`


dylanmcreynolds:
I think this is a great simplification. Gunicorn was seen as a necessary step to acheive scaling, but the realization that even docker-compose can easily perform round-robin load balancing lets us take a lot of complexity out of containers and into the deployment, where it belongs. 

It is truly sad that podman-compose doe not achieve this out of the box. In that case, one could imagine a gunicorn container and multiple tiled containers to achieve horizontal scaling.

PR #383, opened 2023-01-31
Fix specs configuration and add 'reject_undeclared_specs'.

danielballan:
* Fixes bug in the code that merges one or more configuration files that entirely dropped the section that declares specs
* Make `reject_undeclared_specs` configurable, and off by default, for easy experimentation without getting the server admin to add specs (for @cjtitus)


danielballan:
I ran this by @dylanmcreynolds for conceptual approval. I will self-merge shortly because @cjtitus has been patiently waiting for this fix for awhile!

PR #385, opened 2023-02-01
Fix image to report library_version correctly.

danielballan:
This fixes a regression in #382.

Before:

```
$ http :8000/api/v1/ | jq .library_version
"0+unknown"
```

After:

```
$ http :8000/api/v1/ | jq .library_version
"0.1.0a82.post0.dev0+gf0ac3e8"
```


danielballan:
I now have a databroker image derived from this, working locally. Fingers crossed this is the last fix we need for now.

PR #388, opened 2023-02-04
Bundle example dashboard configuration

### LINKED ISSUE #83: Add prometheus metrics
Draw inspiration from https://docs.dask.org/en/latest/setup/prometheus.html. Exposing `gc` metrics in particular seems a good idea.

### LINKED ISSUE #108: Create Grafana "dashboards as code"
It's now possible to build Grafana dashboards for Tiled. As is usual with Grafana, the dashboards need to be designed by hand using Grafana's web UI. It would be convenient if Tiled included some dashboards as a starting point.

The bad news is that this is not a totally straightforward thing to do. The good news is that Jupyter has a solution to the same problem in a BSD3-licensed repository at https://github.com/jupyterhub/grafana-dashboards.

### LINKED ISSUE #109: Document prometheus deployment
https://gist.github.com/danielballan/1bfb0576781c5671e3b97af2ebafcfa1

danielballan:
Tiled publishes [Prometheus](https://prometheus.io/) metrics that can be used to monitor many aspects of its performance. (We were inspired by JupyterHub and the Dask Distributed Scheduler, which also use Prometheus.) Tiled has had this feature for awhile now, but to _use_ it you first have to deploy Prometheus to consume the metrics and then design a dashboard to view them using a tool like [Grafana](https://grafana.com/). This takes some effort.

In this PR, we bundle example configuration for Prometheus and Grafana --- including a dashboard, ready to go! --- along with a `docker-compose` file that can deploy it all together in one line.

```
TILED_SINGLE_USER_API_KEY=secret docker-compose up
```

(Note: `podman-compose` also works.)

Once that's up, navigate to http://localhost:3000/d/Hnvd_TA4z/tiled-dashboard to see live plots:

localhost_3000_d_Hnvd_TA4z_tiled-dashboard_orgId=1 refresh=5s

Before merging, we should add some more plots to the example dashboard. But this is a promising start. All the pieces fit together.

TO DO:

- [ ] Add more plots.
- [x] Add tutorial page on how to use this, with screenshots.
- [ ] Add reference page listing metrics.

Closes #83 
Closes #108 
Closes #109


danielballan:
Notes:

```
sum(rate(tiled_request_duration_seconds_count{code!~"2.."}[1m]))
```

danielballan:
I added some documentation.

There is more to do here, but we what have works and would be easy for others to build on. Instead of letting this sit and get stale, I have merged it.

PR #393, opened 2023-02-07
Bugfix import_object

matthewcarbone:
The optional argument accept_live_object was never used in the tiled/utils:import_object function. Change the function so that if a live object is passed but `accept_live_object==False`, a ValueError is thrown.

Based on a simple search, it appears that `import_object` is used in quite a few places with different values for `accept_live_object`, so hopefully this tightens up expected behavior.

### DIFF
--- a/tiled/utils.py
+++ b/tiled/utils.py
@@ -397,7 +397,7 @@ class Sentinel:
 UNCHANGED = Sentinel("UNCHANGED")
 
 
-def import_object(colon_separated_string, accept_live_object=False):
+def import_object(colon_separated_string, accept_live_object=True):
     if not isinstance(colon_separated_string, str):
         if not accept_live_object:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


danielballan (approved):
Thanks!

matthewcarbone:
@danielballan haha I knew it would fail somewhere. Looks like at some stage `accept_live_object` was `False`, but a live object was passed and the expected behavior by the rest of the code was that it shouldn't matter.

Thoughts on removing `accept_live_object` entirely?

danielballan:
I would prefer to keep a strict mode, as an option, but I like flipping the default to permissive (`accept_live_object=True`) is probably right. Thoughts?

matthewcarbone:
Yeah I like that suggestion. I'll flip the default and see what happens.

danielballan:
Great. To summarize for Future Us: This PR does not change the behavior, because the default "strict" mode was not actually being enforced. This flips the default to permissive but makes it possible to switch on strict mode for real.

PR #394, opened 2023-02-07
Add env var for in-lined contents for HDF5 adapter.

danielballan:
Awhile back we made it possible for the server to send nested information about generations of tiled nodes "in-lined" into a single response, for large reductions in latency. The motivating use case for xarray datasets.

A comment from @SimonHeybrock on Slack reminded me that we never applied this to our HDF5 reader. In this PR, the internal structure and metadata (attrs) of an HDF5 file are in-lined up to 7 generations by default.

Take the NeXus example. 

```sh
tiled serve pyobject tiled.examples.nexus:tree --public
```

Traversing this tree (downloading all group/dataset names and attrs, but not data) issues a single request for all that information and takes about 80 ms to run, including connection time.

```python
from tiled.utils import tree
%timeit tree(from_uri("http://localhost:8000"), None)
```

Reverting to the old behavior, like so:

```sh
TILED_HDF5_INLINED_CONTENTS_MAX_DEPTH=0 tiled serve pyobject tiled.examples.nexus:tree --public
```

the same client-side command issues 150 HTTP requests and take over 800 ms (10X).

### DIFF
--- a/tiled/adapters/hdf5.py
+++ b/tiled/adapters/hdf5.py
@@ -11,6 +11,7 @@ from ..utils import DictView, node_repr
 from .array import ArrayAdapter
 
 SWMR_DEFAULT = bool(int(os.getenv("TILED_HDF5_SWMR_DEFAULT", "0")))
+INLINED_DEPTH = int(os.getenv("TILED_HDF5_INLINED_CONTENTS_MAX_DEPTH", "7"))
 
 
 class HDF5DatasetAdapter(ArrayAdapter):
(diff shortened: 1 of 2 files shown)


tiled/adapters/hdf5.py
```diff
@@ -150,3 +151,6 @@ def _items_slice(self, start, stop, direction):
         if direction < 0:
             items = reversed(items)
         return items[start:stop]
+
+    def inlined_contents_enabled(self, depth):
+        return depth < INLINED_DEPTH
```
dylanmcreynolds, line 156: Should this be `<=`? If the depth limit is 7 and this node is 7 deep, enable inlining?

danielballan: Good point.

tacaswell (approved):

PR #397, opened 2023-02-09
Tolerate stray files in config dir.

danielballan:
Tiled accepts a _directory_ of config files to enable config to be split arbitrarily across separate files. This can help keep things tidy.

In `main`, tiled has a specific list of file types that it _ignores_. In practice, this was erroring out on a `README.md` file and some other things that we failed to account for in that list. I think it would be better to reverse the logic and use an allow-list containing only `.yml` and `.yaml`.

### DIFF
--- a/tiled/_tests/test_config.py
+++ b/tiled/_tests/test_config.py
@@ -1,6 +1,9 @@
+import yaml
+
 from ..adapters.array import ArrayAdapter
 from ..adapters.mapping import MapAdapter
 from ..client import from_config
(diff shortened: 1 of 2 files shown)


mrakitin (approved):
This addition looks great to me! I like the examples with the "Before" and "After" behavior and the test covering the feature.

I did not know `pre-commit` was used as a part of CI, that's cool! I started adopting it in various projects, and it definitely pays back quickly.

PR #398, opened 2023-02-09
Target /deploy *directory* not /deploy/config.yml

danielballan:
If a config file is not specified, Tiled falls back to `./config.yml`. Therefore, this works:

```
docker run -v /tmp/tiled_config/one_config.yml:/deploy/config.yml -p 8000:8000 ghcr.io/bluesky/tiled:v0.1.0a83
```

But if you try to use a configuration _directory_ it will look only `/deploy/config.yml`. If a file with them name happens to existing the directory it will use that alone and silently ignore all others. If a file with that name does not exist, it will fail.

```
docker run -v ./config_dir:/deploy -p 8000:8000 ghcr.io/bluesky/tiled
```

It's possible to work around this:

```
docker run -e TILED_CONFIG=/deploy -v /tmp/tiled_config:/deploy -p 8000:8000 ghcr.io/bluesky/tiled:v0.1.0a83
```

but I think it would be better if the default behavior was to look at `/deploy` instead of `/deploy/config.yml` specifically.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,4 +60,4 @@ WORKDIR /deploy
 
 EXPOSE 8000
 
-CMD ["tiled", "serve", "config", "--host", "0.0.0.0", "--port", "8000", "--scalable"]
+CMD ["tiled", "serve", "config", "--host", "0.0.0.0", "--port", "8000", "--scalable", "/deploy"]


danielballan:
With a locally-built image, I confirmed that the changes in this PR have the expected effect:

```
$ docker run -v /tmp/tiled_config:/deploy -p 8000:8000 tiled-dev:latest
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Using configuration from /deploy

    Navigate a web browser or connect a Tiled client to:

    http://0.0.0.0:8000?api_key=SECRET


INFO:     Started server process [1]
INFO:     Waiting for application startup.
OBJECT CACHE: Will use up to 10_095_076_147 bytes (15% of total physical RAM)
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
```

mrakitin (approved):
Tested locally, worked for me too.

PR #401, opened 2023-02-15
Adapt to change in FastAPI, Starlette

danielballan:
In `main`, we add ASGI middleware for [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) enforcement during server startup. This is no longer allowed, as of a recent release of FastAPI/Starlette. (I haven't bothered to trace down which one is responsible for the change.) That seem sensible enough.

In this PR, the CORS middleware is added before server startup, and only the state affecting its _configuration_ is mutated during server startup. A unit test was added to ensure that:

* We block cross-origin requests by default.
* Setting `allow_origins` mutates the configuration during server startup and has the intended effect.

### DIFF
--- a/tiled/_tests/test_allow_origins.py
+++ b/tiled/_tests/test_allow_origins.py
@@ -1,0 +1,44 @@
+from tiled.client import from_config
+
+strict_config = {
+    "authentication": {
+        "allow_anonymous_access": True,
+    },
+    "trees": [
+        {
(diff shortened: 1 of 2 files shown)


danielballan:
I also verified that our [tutorial using Plotly Chart Studio](https://blueskyproject.io/tiled/tutorials/plotly-integration.html) still works against this PR branch.

PR #409, opened 2023-02-20
Facilitate JSON logs

danielballan:
When errors are logged as multi-line text, it plays badly with system like Splunk that consume each line from the logs as a separate item. Specifically, it becomes difficult to access tracebacks.

Uvicorn supports log configuration involving custom Python objects. It may be convenient for Tiled to ship with a JSON logger to cover this important use case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,6 @@ share/tiled/ui
 config.yaml
 config.yml
 *.sqlite
+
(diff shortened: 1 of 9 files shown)


danielballan:
We have discussed creating a repository of deployment examples. I think this would be better placed there than in Tiled core. Closing.

PR #413, opened 2023-03-02
Test the Grafana branch from danielballan.grafana

jmaruland:


danielballan:
I incorporated this in #388.

PR #416, opened 2023-03-08
Use async interface to auth database and clean up thread management

danielballan:
This PR does several coupled things:

* SQLAlchemy introduced an [async interface](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html) somewhat recently. This refactors Tiled to use it. This avoids issues with threading and SQLite, which is helpful. It is also expected to be easier to debug and have better runtime performance in general.
* When Tiled is used over ASGI-in-process, such as in the test suite, it runs the server's event loop on a background thread. In `main`, there was no public API for stopping these event loops. The test suite leaked ~100 threads, and then waited a lot time at the end of the test run while [the threading module's atexit hooks](https://github.com/python/cpython/blob/b097925858c6975c73e989226cf278cc382c0416/Lib/threading.py#L1569-L1572) to cleans up before the program can exit. Now, we have `Context.close()`, `Context.__enter__` and `Context.__exit__`. Though the test suite, we shut down the application before leaving the test function. This is clean and avoid a big pile-up of cleanup work at the end.

  Before:
  ```py
  client = from_tree(tree)
  ...
  # leaves thread hanging after test runs
  ```

  After:
  ```py
  with Context.from_app(build_app(tree)) as context:
      ...
  # cleans up event loop thread here, via `Context.__exit__`
  ```
* With this PR, Contexts can be constructed via `Context.from_app` (HTTP/ASGI in process) and `Context.from_uri` (HTTP/TCP over network). In `main` we have a zoo of other constructor functions: `from_app`, `from_tree`, `from_config`, `direct_access`, `direct_access_from_profile`. There are several problems with them:
   * Many of them mash together server-side config and client-side config, which is confusing.
   * It's hard to know which one to use.
   * As @dylanmcreynolds once pointed out long ago, that many entrypoints with that many parameters are hard to thoroughly test.
  
  I think an explicit multi-step construction is much better:
   1. Use `build_app(tree, ...)` or `build_app_from_config(config)` to make the app from live objects from a configuration dictionary, respectively.
   2. Use `Context.from_app(...)` to make the context and start the application's event loop in a background thread.
   3. When a user-facing client object (e.g. `Node`) is desired, then finally use `from_context(context)`.

It's more verbose, but the lifecycle is more clear, the server vs. client aspects are well delineated, and the API surface area is reduced.

## Illustration of tread lifecycle with app and Context

We import everything and begin with just 1 thread, the main thread.
```py
>>> import threading
>>> from tiled.client import Context, from_context
>>> from tiled.server.app import build_app
>>> from tiled.adapters.mapping import MapAdapter
>>> len(threading.enumerate())
1
```

We contruct a FastAPI app. Still 1 thread.
```py
>>> app = build_app(MapAdapter({}))
>>> len(threading.enumerate())
1
```

Create a `Context` from an `app` internally creates a `starlette.testclient.TestClient`, which starts the app's event loop on a background thread. It also starts a threadpool.

```py
>>> context = Context.from_app(app)
OBJECT CACHE: Will use up to 10_095_076_147 bytes (15% of total physical RAM)
>>> len(threading.enumerate())
3
>>> threading.enumerate()
[<_MainThread(MainThread, started 139785339070272)>, <Thread(ThreadPoolExecutor-0_0, started 139784379954880)>, <WorkerThread(AnyIO worker thread, started 139784369465024)>]
```

Closing the `Context` stops the event loop and the threadpool.

```py
>>> context.close()
>>> threading.enumerate()
[<_MainThread(MainThread, started 139785339070272)>]
```

This can also be done using a context manager:

```py
>>> with Context.from_app(app) as context: pass
... 
OBJECT CACHE: Will use up to 10_095_076_147 bytes (15% of total physical RAM)
>>> threading.enumerate()
[<_MainThread(MainThread, started 139785339070272)>]
```


tiled/_tests/test_structured_array.py
```diff
@@ -33,7 +34,6 @@ def test_read():
         dtype=[("name", "U10"), ("age", "i4"), ("weight", "f4")],
     )
     tree = MapAdapter({"A": ArrayAdapter.from_array(data)})
-
-    client = from_tree(tree)
-
+    with Context.from_app(build_app(tree)) as context:
```
dylanmcreynolds, line 37: This pattern is used a lot, and this is somewhat picky, but could you make it a little more DRY but defining a function to be called each time:

```python
def client_app(tree):
    with Context.from_app(tree) as context:
    return client
```

Then each test that obtains a client could:

```python
client = cleint_app(tree)
```

This wouldn't save all that much code, but if your be a nice future proofing if you realize that your strategy needs to change slightly, it could be changed in one place.

danielballan: That doesn't quite work because:

```py
def client_app(tree):
    with Context.from_app(build_app(tree)) as context:
    # Server event loop stops here.
    return client  # dead on arrival
```

It could be done like:

```py
import contextlib

@contextmanager
def client_app(tree):
    with Context.from_app(build_app(tree)) as context:
        yield client
```

If it were always that simple, I think I would agree we want a function to stay DRY. But it gets complicated.

* We sometimes build the app based on `config` instead of a `tree`, so we'd need a `client_app_from_config` variant. Stepping back, this is exactly what `from_tree` and `from_config` were (minus the context manager aspect).
* in many of the tests we need to pass in additional parameters to either `Context.from_app` or `from_context` or both, such as
  ```py
  with Context.from_app(tree, authentication={...}) as context:
      client = from_context(context, structure_clients="dask")
  ```
  so `client_app` would have to handle a mixture of server-related and client-related parameters---kind of confusing. The `from_tree` and `from_config` functions also had this downside.

Is it better to be DRY with `Context.from_app`, `from_context`, `from_tree`, and `from_config` in the mix, or is it better to be repetitive with just `Context.from_app` and `from_context`?

For test modules that need a consistent `tree` or `config` across many tests, I did the equivalent of `client_app` as a `pytest.fixture`:

```py
@pytest.fixture(scope="module")
def client():
    app = build_app(tree)
    with Context.from_app(app) as context:
        yield client
```

such as in `test_search.py`.

danielballan:
Thanks @dylanmcreynolds for the review. This is such a big PR I don't think I can expect much more review, so I'm going to move this one along.

PR #419, opened 2023-03-08
Add guide on custom Python client objects.

danielballan:


danielballan:
This is a guide on how to register custom Python objects that Tiled should use for certain types of data.

Of interest to @mattewdcarbone. Possibly also of interest to @pbeaucage.

docs/source/how-to/custom-clients.md
```diff
@@ -0,0 +1,192 @@
...
+1
+````
+
+where the string representation is the exactly the code you would run to
+reproduce that object.
+
+You may want to override `__str__` in addition, which determines what
+is returned by `str(...)` and displayed by `print(...)`. If you override
+`__str__` but not `__repr__`, then `__repr__` falls back to `__str__`.
+```
```
matthewcarbone, line 81: Something weird might be going on here with the triple quotes.

matthewcarbone:
@danielballan this is really great, thank you. I think it'll be really helpful.

Quick question, for the custom clients, how does that play with `from_uri`? I understand the pieces about specs, but it's not clear how I get e.g.

```
class CustomClient(Node):  # <- note, `Node`?
     @property
     def element(self):
         return self.metadata["XDI"]["Element"]["Symbol"]
```

from `from_uri`.

thanks!

danielballan:
If I understand right, you want to customize the client object that represents the root of the tree, the object returned directly by `from_uri(...)`. Client-side, this runs through the very same code path and works in the same way. Server-side, you need to give the root node a spec that your client to hook into.

This is not supported by the mongomock-backed adapter [you are currently using](https://gist.github.com/danielballan/cd787e98bd0a2821b5f71932e22e460f), but it's a very easy patch. Edit `databroker.experimental.server_ext` to accept a `specs` parameter, and set it as an attribute. It should default to empty list.

```py
class MongoAdapter(collections.abc.Mapping, IndexersMixin):
    structure_family = "node"
    query_registry = QueryTranslationRegistry()
    register_query = query_registry.register

    def __init__(
        self,
        *,
        database,
        directory,
        queries=None,
        sorting=None,
        metadata=None,
        access_policy=None,
        specs=None
    ):
        self.specs = specs or []
        ...
```

And then, in your config, add a spec:

```yaml
trees:
# Create a Mongo "database" in memory.
# When the server is stopped the database will be forgotten.
- tree: databroker.experimental.server_ext:MongoAdapter.from_mongomock
  path: /
  args:
    directory: ./data_files
    specs:
      - CatalogOfXAS  # or whatever...
```

That spec will affect the root node. It will not affect any of the child nodes (e.g. the uploaded datasets).

matthewcarbone:
@danielballan Mainly, I'm just wondering how `from_uri` knows that you've defined this new client. Easier to explain via example. I guess what I want to do lazily is this:

```python
class CustomClient:
    def __init__(self, client):
        self._client = client
    
    def special_method(self, ...):
        do_something_with(self._client)
```

and of course it would just be easier to inherit the base `Node` object so that all of the other methods on the `Node` are defined and I don't need to hack the `__getattr__` method on my above `CustomClient`.

Basically, how do you tell Tiled about `CustomClient`? It's not obvious where that's registered (unlike the `XDIDatasetClient`, where it's clear how to manually register it using the Tiled API and also through entrypoints). Does this make sense? Thanks!

danielballan:
Yes, you should inherit from `Node`. It is not necessary to override `__getitem__` because base `__getitem__` in Node checks the `structure_clients` you passed in to `from_uri` and the entry points you register when deciding what type to return. It propagates this in its child, so that configuration set that at the top of the tree affects the whole tree.

danielballan:
Discovery of entrypoints happens here:

https://github.com/bluesky/tiled/blob/96ac213b7521f7706975d92a540415b835ad1831/tiled/client/constructors.py#L213-L218

The function `client_for_item` determines which type of object to use:

https://github.com/bluesky/tiled/blob/96ac213b7521f7706975d92a540415b835ad1831/tiled/client/utils.py#L266-L294

It's called by `from_uri`:

https://github.com/bluesky/tiled/blob/96ac213b7521f7706975d92a540415b835ad1831/tiled/client/constructors.py#L249

and `Node.__getitem__`:

https://github.com/bluesky/tiled/blob/96ac213b7521f7706975d92a540415b835ad1831/tiled/client/node.py#L321

and well as some other methods on `Node`.

matthewcarbone:
Ok let me play with this a bit. I'm still a bit confused, I think, but at this stage some hands-on work is what's required. I'll probably ask more questions later. Thanks! 👍

matthewcarbone:
Small bug with your proposed client. You used `x` where it should've been `self`. Here's a fix for you:

```python
from tiled.client.dataframe import DataFrameClient

class XASDatasetClient(DataFrameClient):
    """A custom DataFrame client for DataFrames that have XDI metadata."""

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # Optional sanity check to ensure this cannot be accidentally
        # registered for use with data that is not a dataframe.
        assert self.item["attributes"]["structure_family"] == "dataframe"

    def __repr__(self):
        md = self.metadata
        element = md["sample_metadata"]["element"]
        edge = md["sample_metadata"]["edge"]
        return f'<{self.item["id"]} {element} {edge}>'
```

matthewcarbone:
Ok so this is really excellent. I think I know how to frame my question. It's now clear how to point certain specs to certain `client`-like objects. Here's what I did, and it works perfectly fine:

```python
custom = dict(DEFAULT_STRUCTURE_CLIENT_DISPATCH["numpy"])
custom["experimental_XAS"] = XASDatasetClient
api_key = environ["TILED_SINGLE_USER_API_KEY"]
client = from_uri(f"http://127.0.0.1:8000?api_key={api_key}", structure_clients=custom)

# Upload some dummy data
df = pd.DataFrame({
    "energy": [1, 2, 3],
    "mu": [4, 5, 6]
})

metadata = {
    "dataset": "test_dataset",
    "sample_metadata": {"element": "Co", "edge": "K"},
    "experiment_metadata": {"facility": "NSLSII", "beamline": "ISS", "sample_id": "123"}
}

client.write_dataframe(df, metadata=metadata, specs=["experimental_XAS"])

first = client.values().first()
first
# <86e8d81b-54f7-4e6a-b0d5-c54db81662fe Co K>
# (perfect!)
```

Now, what I want to do now is to do exactly the same thing, but instead of overriding the behavior to a certain spec, I want to override behavior to the "base" client _itself_. In other words, what I want to do is this:

```python
something_perhaps = ...
client = from_uri(f"http://127.0.0.1:8000?api_key={api_key}", some_arg=something_perhaps)

# new client has a new __repr__ which simply prints "foo" before everything
client
# <foo Node {'f94aafa9-032c-4a64-baa4-a86ed6612607'}>
```

How do I register a change to the "base" node like this? Thanks again!

danielballan:
You can do:

```py
custom["node"] = CustomNodeClient
```

Tiled does not enable `node` to be overridden by an entry point because if two libraries in a given software environment both did that it would be unclear what the precedence should be. Specs, on the other hand, are expected to be rather specific, so collisions should be rare and manageable on a case by case basis. (Someday, if we need some kind of official registry or something we could conceivably go there.)

Given that limitation, I think I'd suggest following what [Joe did](https://github.com/AI-multimodal/aimmdb/blob/7d88dda3cd9c2fbe3882d1f9917901d84cfe6654/setup.py#L57-L64). Give the root node (the thing returned by `from_uri` a spec [using the approach I outlined above](https://github.com/bluesky/tiled/pull/419#issuecomment-1461146780) use that to create a customized user experience from the top.

The alternative would be to have your own `from_uri` where you can override `node`.

```py
def matt_from_uri(*args, **kwargs):
    return tiled.client.from_uri(*args, structure_clients=custom, **kwargs)
```

But I see downsides there because if you try to connect to, say, `tiled.nsls2.bnl.gov` with `matt_from_uri` and `node` is overwritten there could be unexpected behavior. That's why I'd recommend starting with Joe's approach.

matthewcarbone:
@danielballan perfect, thank you. It's funny I was already in the process of implementing something just like Joe's solution. What do you think of this?

```python
# aimm2.client

from warnings import warn

from tiled.client.dataframe import DataFrameClient


class OperatorHelperMixin:
    """A mixin class that defines how operators can act on the classes."""


class XASDatasetClient(DataFrameClient):
    """A custom DataFrame client for DataFrames that have XDI metadata."""

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # Optional sanity check to ensure this cannot be accidentally
        # registered for use with data that is not a dataframe.
        assert self.item["attributes"]["structure_family"] == "dataframe"

    def __repr__(self):
        md = self.metadata
        element = md["sample_metadata"]["element"]
        edge = md["sample_metadata"]["edge"]
        return f'<{self.item["id"]} {element} {edge}>'

    def apply_QAQC_(self):
        """Performs a quality assurance/quality control check on the data.
        """

        ...


def from_uri(*args, **kwargs):
    """Manually registers all clients. Lightweight wrapper for
    `tiled.client:from_uri`. Note that this is not a substitute for registering
    the clients via Python entrypoints, as that adds additional functionality
    that this function cannot. Hence, this is mostly useful for developemnt or
    for when you know exactly what you're doing."""

    from tiled.client.node import DEFAULT_STRUCTURE_CLIENT_DISPATCH
    from tiled.client import from_uri as _from_uri

    custom = dict(DEFAULT_STRUCTURE_CLIENT_DISPATCH["numpy"])
    custom["experimental_XAS"] = XASDatasetClient

    if "structure_clients" in kwargs.keys():
        warn(
            "structure_clients key found in kwargs. This will attempt to be "
            "merged with the existing clients"
        )
        custom = {**kwargs["structure_clients"], **custom}

    return _from_uri(*args, **kwargs, structure_clients=custom)
```

Obviously still a work in progress.

danielballan:
Yes, looks on track to me.

danielballan:
Merging to publish these docs. If anyone runs into errors of points of confusion while using them, we can improve them in future PRs.

PR #421, opened 2023-03-17
Use a more robust ALL sentinel.

danielballan:


tacaswell (approved):
can self-merge on green

maffettone (approved):

PR #425, opened 2023-03-21
Fix import object

danielballan:
Tiled temporarily adds the directory containing the YAML config file(s) to `sys.path` during server setup, so that _ad hoc_ Python functions and classes can be defined there. It has worked out well to use statically-analyzable YAML for _most_ things but provide an easy escape hatch / escalation pathway for extensions.

In #416 I accidentally broke this feature and no tests were covering it. This PR fixes it and covers it with a test.


danielballan:
Self-merging as hot fix to deploy before the X-ray beam comes back. :zap:

PR #427, opened 2023-03-22
Handle case where __stdin__ is closed.

danielballan:
This addresses an issue we saw "in the wild" with queueserver at BMM. Needs unit tests with mock.


danielballan:
Rebased, and added unrelated fix to tests that I hit during the rebase.

PR #428, opened 2023-03-27
Test profile with 'direct' and fix bug.

danielballan:


danielballan:
Unit test reproduces bug hit by @prjemian.

```
E           ImportError: cannot import name 'build_app_from_config' from 'tiled.server' (/home/dallan/Repos/bnl/tiled/tiled/server/__init__.py)
```

The [direct profile feature](), designed to facilitate a gradual transition from "local process databroker" to the service-based one, was under tested. We would have hit this bug ourselves if we had deployed this to the floor.

danielballan:
I also improved the error reporting. Previously, we did not validate the server-side configuration in-lined in the profile under `direct:`. A broken config like this:

```yaml
test:
  direct:
    # Intentional mistake! Value of `trees:` must be a _list_.
    trees:
      path: /
      tree: tiled.examples.generated_minimal:tree
```

produced unhelpful errors downstream like:

```py

In [2]: c = from_profile("test")
...
AttributeError: 'str' object has no attribute 'get'
```

Now, the configuration is validated same as it would be in normal HTTP/TCP server mode, and we get a clearer error:

```py
ConfigError: ValidationError while parsing configuration file /home/dallan/.config/tiled/profiles/test.yml: {'path': '/', 'tree': 'tiled.examples.generated_minimal:tree'} is not of type 'array'
```

mrakitin:
Oh, I think @thomaswmorris encountered that one last week. We weren't sure whether it was a local config issue or a bug.

mrakitin (approved):
Looks good to me, thanks for the fix.

prjemian (approved):
I'm good with this.  Thanks for the quick action!

PR #429, opened 2023-03-28
Refactor signature-patching function to handle search and distinct for FastAPI

jmaruland:
In Progress. Created one method that helps FastAPI generate documentation for search() and distinct. Currently, the server is registering the methods but it's not finding any of them when they are called from the client side

### DIFF
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -7,6 +7,7 @@ import sys
 import urllib.parse
 from functools import lru_cache, partial
 from pathlib import Path
+from typing import List
 
 import anyio
 import packaging.version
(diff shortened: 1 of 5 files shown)


jmaruland:
Adding `distinct()` at the end of a sequence of search queries, like
`client.search(query1).search(query2)....distinct(key)`
applies the `distinct` request to the search result. Before, the `distinct()` request was being run from the perspective of the root node. This last commit gives more freedom to the use of `distinct()`

tiled/server/app.py
```diff
@@ -324,6 +326,21 @@ async def index(
         # And add this authentication_router itself to the app.
         app.include_router(authentication_router, prefix="/api/v1/auth")
 
+    # The /search route is defined at server startup so that the user has the
```
danielballan, line 329:
```suggestion
    # The /search route is defined after import time so that the user has the
```

danielballan (commented):
Two minor comments to keep comments up to date and avoid cruft.

danielballan (approved):

danielballan:
I took this for a test drive locally. Works as advertised.

PR #430, opened 2023-03-28
Validate server config in 'direct'.

danielballan:
This is a commit I meant to push to #428. Below is the original comment explaining this aspect.

---

I also improved the error reporting. Previously, we did not validate the server-side configuration in-lined in the profile under `direct:`. A broken config like this:

```yaml
test:
  direct:
    # Intentional mistake! Value of `trees:` must be a _list_.
    trees:
      path: /
      tree: tiled.examples.generated_minimal:tree
```

produced unhelpful errors downstream like:

```py

In [2]: c = from_profile("test")
...
AttributeError: 'str' object has no attribute 'get'
```

Now, the configuration is validated same as it would be in normal HTTP/TCP server mode, and we get a clearer error:

```py
ConfigError: ValidationError while parsing configuration file /home/dallan/.config/tiled/profiles/test.yml: {'path': '/', 'tree': 'tiled.examples.generated_minimal:tree'} is not of type 'array'
```

### DIFF
--- a/tiled/_tests/test_client.py
+++ b/tiled/_tests/test_client.py
@@ -1,10 +1,12 @@
 from pathlib import Path
 
 import httpx
+import pytest
 import yaml
 
 from ..adapters.mapping import MapAdapter
 from ..client import Context, from_context, from_profile
+from ..config import ConfigError
 from ..profiles import load_profiles, paths
 from ..server.app import build_app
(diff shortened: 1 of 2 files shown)


mrakitin (approved):
Looks good to me!

prjemian:
@danielballan : Thanks for the work on this!

PR #431, opened 2023-03-31
Fix importable object in direct profile

danielballan:
This regression was seen in the wild by @mrakitin. It was introduced recently, in #416. The issue related to a previously fixed regression, #425. I believe this is the last instance of this bug.

The test reproduces the issue:

```
FAILED tiled/_tests/test_import_object.py::test_direct_profile - ModuleNotFoundError: No module named 'custom_module'
```

And it is then resolved by the next commit.

### DIFF
--- a/tiled/_tests/test_import_object.py
+++ b/tiled/_tests/test_import_object.py
@@ -1,6 +1,10 @@
 from pathlib import Path
 
+import yaml
+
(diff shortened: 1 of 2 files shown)


mrakitin (approved):
That solved the issue at TES for us! Thanks for the quick fix, @danielballan!

PR #433, opened 2023-04-07
Elide getitem requests

danielballan:
For a long time we have supported

```py
c['x', 'y', 'z']
```

as an alias for

```py
c['x']['y']['z']
```

However, we have not been taking advantage of the opportunity to elide the former into a single HTTP request that "jumps" down the tree, instead of making separate HTTP requests at each jump. (Why? We deferred this work because of the subtlety of the implementation and some lingering questions about the meaning of paths in Tiled which have since been settled.)

This is expected to give good savings on latency.

### DIFF
--- a/tiled/client/node.py
+++ b/tiled/client/node.py
@@ -287,7 +287,9 @@ class Node(BaseClient, collections.abc.Mapping, IndexersMixin):
                 len(data) == 1
             ), "The key lookup query must never result more than one result."
             (item,) = data
-            return client_for_item(self.context, self.structure_clients, item)[tail]
+            result = client_for_item(self.context, self.structure_clients, item)
+            if tail:
+                result = result[tail]
         else:
             # Straightforwardly look up the key under this node.
             # There is no search filter in place, so if it is there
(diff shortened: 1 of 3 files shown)


danielballan:
@bruceravel confirmed nice speed-ups on this branch.

tiled/client/node.py
```diff
@@ -300,24 +298,31 @@ def __getitem__(self, key, _ignore_inlined_contents=False):
...
+                    except ClientError as err:
+                        if err.response.status_code == 404:
+                            # If this is a scalar lookup, raise KeyError("X") not KeyError(("X",)).
+                            err_arg = keys[i:]
+                            if len(err_arg) == 1:
+                                (err_arg,) = err_arg
+                            raise KeyError(err_arg)
+                        raise
+                    item = content["data"]
+                    break
```
dylanmcreynolds, line 323: Doesn't this `break` mean we'll only the first key? Maybe I misunderstand what the loop is for?

danielballan: The server may greedily send nested information about children ("inlined contents"). Logic is:

1. In a loop, walk nested information that we have already downloaded about the children of this node (if any).
2. If we reach the bottom of what we already downloaded, send out one request for the rest (and break the walking loop).

danielballan: I'll add this as a code comment.

dylanmcreynolds (approved):

danielballan:
Note: codecov was removed from PyPI. We haven't been using it anyway; it was just carried in from the cookiecutter. Removed in unrelated commit here to un-break CI.

PR #436, opened 2023-05-11
Added test case for multiple specs vs one similar spec for distinct

jmaruland:
This test case show that using distinct on `spec=[a]` and `spec=[a, b]` should count them as unique sets


danielballan:
I think it's still an open question what the right behavior is here: tallying distinct specs or tallying distinct _groups_ of specs. We may want to support both in the end.

But, in the initial work on this in AIMMDB we want with "groups of specs" and I think it makes sense to add this test to codify the current behavior so that we can be sure that any changes in the future are intentional.

PR #439, opened 2023-06-01
Fix up OIDC and add a locally-runnable exampmle

danielballan:
It turns out that #416 broke our OIDC authentication pathways _badly_, with multiple bugs. (N.B. There are no _security_ issues here; it just fails to run and shows the client status code `500`, as it should in such an event.) This is not used in production at NSLS-II---only as an option on tiled-demo.blueskyproject.io, and it went unnoticed for quite awhile.

This PR adds an example that is fairly easy to run locally, to lower the barrier to interactive testing. More work is needed to turn this into proper unit tests.

### DIFF
--- a/example_configs/simple_oidc/README.md
+++ b/example_configs/simple_oidc/README.md
@@ -1,0 +1,46 @@
+# Run a local OIDC provider
+
+This example runs a [toy OIDC provider][] (the same authentication mechanism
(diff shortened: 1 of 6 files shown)


danielballan:
Merging this for the "hot fix" aspects, needed up for an upcoming talk, but very happy to get post-merge review.

PR #442, opened 2023-06-05
Fix from profile error

danielballan:
Caught in the wild at CHX. Includes test that reproduces the error.

### DIFF
--- a/tiled/client/constructors.py
+++ b/tiled/client/constructors.py
@@ -270,6 +270,6 @@ def from_profile(name, structure_clients=None, **kwargs):
         context = Context.from_app(
             build_app_from_config(config, source_filepath=filepath),
         )
-        return from_context(context)
+        return from_context(context, **merged)
     else:
         return from_uri(**merged)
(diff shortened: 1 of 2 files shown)


tiled/client/constructors.py
```diff
@@ -268,7 +268,7 @@ def from_profile(name, structure_clients=None, **kwargs):
                 f"ValidationError while parsing configuration file {filepath}: {msg}"
             ) from err
         context = Context.from_app(
-            build_app_from_config(config, source_filepath=filepath), **merged
```
tacaswell, line 271: where do the things that used to come in via `**merged` get threaded through instead now?

danielballan: It's _just_ below the fold shown by GitHub, Line 275.

tacaswell: That is is the "else" case, in this branch it appear to be silently ignored.

tacaswell (approved):

Issue #423, opened 2023-03-20
Version string from development installation on Windows is odd


danielballan:
On Windows @jmaruland gets version strings like `0.post1850+g933526a` which do not account for the tags, as opposed to `0.1.0aX.postY`. This matters because the server cannot tell whether this version is newer than or older than the minimum version it supports.

We have checked that he does have all the tags fetched locally.

I suspect we have run into some Windows-specific versioneer issue. If there is not a well-known fix, the best way forward here may be to just move on from versioneer to whatever the modern packaging solution is.

PR #459, opened 2023-06-14
Switching from setup.py+setuptools to pyproject.toml with hatchling

danielballan:
The motivation for the change is two-fold:

* The general impulse to modernize, moving away from `setup.py`, and to follow the recommendation in https://learn.scientific-python.org/development/
* A specific issue we hit with versioneer and Windows, seemingly related to https://github.com/python-versioneer/python-versioneer/issues/273, which this fixes

Closes https://github.com/bluesky/tiled/issues/423


danielballan:
I expect this will take some iteration to get quite right, but I think now is the time to try it.

PR #461, opened 2023-06-15
Update docker-compose to use catalog

danielballan:
We have an awesome `docker-compose.yml` with tiled+prometheus+grafana, documented here https://blueskyproject.io/tiled/how-to/metrics.html.

This has had an annoying extra step: we need the user to provide some data to serve, or thread a toy config file in, which is not a very realistic scenario or useful in production. But now, with the writable catalog in Tiled, we can launch empty instance and just tell the user to write some data in. This is something that could be useful to just deploy as is.

Question: How to deal with the fact that the database needs to be initialized at some point? How do people normally handle this? We could make it a two-step process:

1. Run the tiled container with the `CMD` `tiled catalog init ...` to initialize the database.
    ```
    mkdir storage
    podman run -v /storage:./storage:rw tiled tiled catalog init /storage/catalog.db
    ```

2. `podman-compose up`

Is that the best we can do?

Secondary question: Should we ship the `docker-compose.yml` such that it _builds_ the current repo or should we hard-code a tagged version of pre-built tiled? Or should we ship _both_ and if so what should we call them?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,8 +57,13 @@ ENV VIRTUAL_ENV=/opt/venv
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 COPY --from=builder $VIRTUAL_ENV $VIRTUAL_ENV
 
+COPY ./entrypoint /usr/local/bin/entrypoint
+ENTRYPOINT ["/usr/local/bin/entrypoint"]
+ENV TILED_CATALOG_DB="/storage/catalog.db"
+
 WORKDIR /deploy
 
 EXPOSE 8000
 
-CMD ["tiled", "serve", "catalog", "/storage/catalog.db", "--write", "/storage/data", "--host", "0.0.0.0", "--port", "8000", "--scalable"]
+# Invoke the shell (bash -c ...) to enable variable expansion of ${TILED_CATALOG_DB}.
+CMD ["bash", "-c", "tiled serve catalog --write /storage/data --host 0.0.0.0 --port 8000 --scalable ${TILED_CATALOG_DB}"]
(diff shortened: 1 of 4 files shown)


danielballan:
Just to state an alternative: another possible answer to the first question is to add some kind of `--init-if-empty` flag. But this kind of conditional initialization makes me itchy. I can imagine it going very wrong in a _scaled_ deployment where N containers are launching simultaneously. That seems like a pretty realistic scenario.

dylanmcreynolds:
For the first question., I think the official mongo image handles this with external files that you can mount. I don't know if we want to go that far but see [Initializing a fresh instance](https://hub.docker.com/_/mongo). Postgres has a [similar mechanism](https://levelup.gitconnected.com/creating-and-filling-a-postgres-db-with-docker-compose-e1607f6f882f). These both seem cool but a lot of work to maintain for us though.

For the second question, I think a tag to the prebuilt image is sufficient. Providing a docker-compose file is a favor for people who want kick the tires. If you're building from current source, you probably know how to volume mount the source into the repo.  Some projects provide a `docker-compose-dev.yml` or the like, but that seems excessive to me.

danielballan:
Oh, that Mongo link was interesting. I think we could apply that easily enough. Imagine docs like...

<hr />

For the first use, run:

```bash
TILED_INIT_DB=./storage/catalog.db TILED_SINGLE_USER_SECRET_KEY=secret podman-compose up
```

The `TILED_INIT_DB` parameter creates and initializes a database that Tiled uses to store metadata.

If you restart the container later, omit `TILED_INIT_DB`; the database is already initialized.

Above we are using a simple file as the database (SQLite). Alternatively, a PostgreSQL URI may be used

```bash
TILED_CATALOG_URI="postgresql+asyncpg://username:password@localhost/database_name:5432"
TILED_INIT_DB=$CATALOG_URI
podman-compose up
```

<hr />

In terms of maintanance, I _think_ that just requires adding a small bash script to the image, using it as the default `CMD`. I wouldn't expect it evolve much from there.

danielballan:
I implemented something that is a bit simpler than one I outlined above. I like this.

danielballan:
Superseded by #507

PR #465, opened 2023-06-16
Support Adapters that Interact with Services

### LINKED ISSUE #453: Support Adapters that interact with authenticated services
We have Tiled Adapters that interact with filesystems and databases. We have long planned Adapters that interact with services, but we have not yet implemented one.

To implement one, we seem to require two things:

1. The addition of "session state" to stash information that can be used to access the service. https://github.com/bluesky/tiled/blob/2862817d4e36e84ed8807344a95390712eb90cf3/tiled/database/orm.py#L184-L205
2. A way to pass session state to an Adapter

For (2) I propose this:

* Implement a new interface, which implements the method `with_session(session)` (name negotiable). That method should return an Adapter to an external service, and said Adapter should be passed the session info, including session state.
* In the server, [when we are traversing the URL path to find the entry of interest](https://github.com/bluesky/tiled/blob/2862817d4e36e84ed8807344a95390712eb90cf3/tiled/server/dependencies.py#L83), check for the method `with_session` and, if it's there, call it with the current session. This kind of thing is currently implement with `hasattr`, which is probably fine for now, but we should use protocols later: #449.

dylanmcreynolds:
Resolves #453

An issue with having an Adapter that interacts with third  party services is a third party service may have its own authentication scheme. That scheme might involve passing a token through Tiled from the client as a password, and then when the Adapter interacts with the third party service, the it somehow needs access to that token from within the same user session. 

This PR implements two parts of this. 

1. It adds the ability for a username/password Authenticator to return a state object back to Tiled.  Tiled stores that dictionary in the session row of the auth database.

2. We will provide a mechanism for Adapters to have access to this state.

The initial commits address the first case but not yet the second, so this PR is preliminary.

### DIFF
--- a/example_configs/simple_oidc/README.md
+++ b/example_configs/simple_oidc/README.md
@@ -8,11 +8,12 @@ use in production.**
    follows.
 
 ```
-docker run --rm -p 9000:9000 -v $(pwd):/config -e CONFIG_FILE=/config/oidc_provider_config.json -e USERS_FILE=/config/users.json qlik/simple-oidc-provider:0.2.4
+docker run --rm -p 9000:9000 -v $(pwd):/config -e CONFIG_FILE=/config/oidc_provider_config.json -e USERS_FILE=/config/users.json docker.io/qlik/simple-oidc-provider:0.2.4
 ```
 
-2. From a web browser or commandline, access `http://localhost:9000`. These
-   contain public keys which _reset every time the container is (re)started_.
+2. From a web browser or commandline, access
+   `http://localhost:9000/keys`. This contains
+   public keys which _reset every time the container is (re)started_.
    Fill them into the Tiled configuration in this directory, `config.yml`,
    under `public_keys:`.
 
--- a/example_configs/simple_oidc/README.md
+++ b/example_configs/simple_oidc/README.md
@@ -12,7 +12,7 @@ docker run --rm -p 9000:9000 -v $(pwd):/config -e CONFIG_FILE=/config/oidc_provi
 ```
 
 2. From a web browser or commandline, access
-   `http://localhost:9000/keys`. This contains
+   `http://localhost:9000/certs`. This contains
    public keys which _reset every time the container is (re)started_.
    Fill them into the Tiled configuration in this directory, `config.yml`,
    under `public_keys:`.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-        - "3.7"
         - "3.8"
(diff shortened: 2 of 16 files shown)


danielballan:
I like this.

We do need a database revision for this, to support migration.

https://github.com/bluesky/tiled/blob/3127d98cfaae36d62588feca25e3fae017d3f009/tiled/database/alembic_utils.py#L62-L74

dylanmcreynolds:
> I like this.
> 
> We do need a database revision for this, to support migration.
> 
> https://github.com/bluesky/tiled/blob/3127d98cfaae36d62588feca25e3fae017d3f009/tiled/database/alembic_utils.py#L62-L74

Interesting. I would have thought that SQLAlchemy would gracefully handle a new column with no data changes.

danielballan:
- Fixed errors in LDAP test on CI
-  Manually this against the OIDC example, which we do not run on CI, and resolved some breakage. (It would be nice if we _could_ test OIDC on CI. I'm not sure how that would work, with a web browser involved in the flow, but people must have some way to do this.)
- Placed session state in JWT, where `get_session_state` looks for it.

Finally, I changed:

```py
    entry.with_session_state(state)
```

to

```py
    entry = entry.with_session_state(state)
```

The adapter _may_ do:

```py
class Adapter:
    ...

    def with_session_state(self, state):
       self.state = state  # mutate!
       return self
```

but now it may also do:

```py
class Adapter:
    ...

    def with_session_state(self, state):
        return NewInstanceOfAdapterOrSomething(...)
```

which seems safer and better to me offhand. Anyway, we have both options now.

The only remaining issue is that `typing.Protocol` is not available in Python 3.7. This may be a signal to just drop 3.7, as it is end-of-life anyway. (Security support ended two weeks ago.)

example_configs/simple_oidc/README.md
```diff
@@ -8,11 +8,12 @@ use in production.**
...
 
 ```
-docker run --rm -p 9000:9000 -v $(pwd):/config -e CONFIG_FILE=/config/oidc_provider_config.json -e USERS_FILE=/config/users.json qlik/simple-oidc-provider:0.2.4
+docker run --rm -p 9000:9000 -v $(pwd):/config -e CONFIG_FILE=/config/oidc_provider_config.json -e USERS_FILE=/config/users.json docker.io/qlik/simple-oidc-provider:0.2.4
 ```
 
-2. From a web browser or commandline, access `http://localhost:9000`. These
-   contain public keys which _reset every time the container is (re)started_.
+2. From a web browser or commandline, access
+   `http://localhost:9000/keys`. This contains
```
danielballan, line 15: This should be `/certs`.

PR #466, opened 2023-06-21
Update to deal with upstream Python packaging changes.

danielballan:


danielballan:
This CI is not relevant for this change.

PR #467, opened 2023-06-21
Fix docker compose

danielballan:
While we iterate on #461, a small fix:

* `context` was misspelled, so this was straight broken
* While we're at it, for this use case it's better to just point to an image rather than build it. Developers can always edit this file to build it if they want.


danielballan:
The CI is not relevant for this change.

PR #474, opened 2023-06-24
Build React UI as part of Python build process

danielballan:
Tiled has a React-based web UI. The source, under `web-frontend/`, must be built and placed above the Python package at the standard location for shared data: `{...}/share/tiled`.

Up till now, I've been doing this manually has part of the release process so that the wheel typically installed by `pip install tiled` includes the pre-built web UI, ready to go. A setuptools feature placed it in `{...}/share/tiled` at install time.

In this PR, I configured `pyproject.toml` so that `pip install .` (and also "editable" installations `pip install -e .`) perform the build as part of the build process.

It requires an `npm` executable on the `PATH` in order to do this. If it cannot find one, it logs a message to stderr and skips over it. The user can also opt out of this explicitly, whether `npm` is available or not, with the env var `TILED_BUILD_SKIP_UI=1`. This is documented in the installation docs.

This work required some changes to the `Dockerfile`. It seems like a good moment to set up [container canary](https://github.com/NVIDIA/container-canary), which runs a container image and performs checks against it, as specified in a YAML file (`canary-validator.yml`, in this case).

Here it's running against the latest release:

```bash
$ canary validate --file canary-validator.yml ghcr.io/bluesky/tiled:latest
Validating ghcr.io/bluesky/tiled:latest against tiled
 Serves a UI at /ui/browse                          [passed]
 Serves a REST API at /api/v1                       [passed]
 Exposes an HTTP interface on port 8000             [passed]
validation passed
```

And here it's running against this branch, built locally:

```bash
 canary validate --file canary-validator.yml localhost/tiled_dev:latest
Validating localhost/tiled_dev:latest against tiled
 Exposes an HTTP interface on port 8000             [passed]
 Serves a UI at /ui/browse                          [passed]
 Serves a REST API at /api/v1                       [passed]
validation passed
```

I have added a GH Action that runs this on every PR, based on [an NVIDIA blog post](https://developer.nvidia.com/blog/testing-container-images-against-multiple-platforms-with-container-canary/).


danielballan:
CI failed because:

```
Treating warnings as errors because process.env.CI = true.
```

danielballan:
The warnings are revolved.

Next I'm seeing a failure from `python -m build`. (As an aside, I don't understand why `pip install .` works but `python -m build` fails.) I think the issue comes down to the fact that we are using a deprecated stack:

> While create-react-app is still functional, developers are now encouraged to use alternative tools like Next. js and Remix or Vite.

I intend to refactor to use vite.

danielballan:
This is good shape now. The build and release process and standard and automated. All of my clever "get it working..." hacks are gone, and I think anyone familiar with npm, vite, and modern Python packaging would make their way through this without surprises. I added a GH action to publish releases to PyPI, closely following https://learn.scientific-python.org/development/guides/gha-pure/.

Notes for #283...

* The UI defaults to being served at the base path `/ui/`. This can only be changed by (re)building with `vite build --base=/my/custom/path`. Ref: https://vitejs.dev/guide/build.html#public-base-path
* The UI makes an API call `GET /tiled-ui-settings` to learn the base URL of the API (default: `/api/v1`). The API _can_ be changed without rebuilding.

PR #478, opened 2023-07-04
Rename 'node' to 'container' and shorten URLs.

### LINKED ISSUE #472: Give a more distinct name to the structure family currently called "node"
In Tiled we have defined structure families:

https://github.com/bluesky/tiled/blob/6b1ee862e571a6e6493f5446652996600ec98638/tiled/structures/core.py#L12-L16

One them is `"node"`. It is a container that has no structure or data in itself, but it has children. It may correspond to a directory, a dictionary-like object, an HDF5 group, or any similar structure.

Confusingly, we also use the word "node" to refer to any object in tiled, whatever its structure. To give just one example, the endpoint `/node/metadata/x/y/z` accesses the metadata for the node at `/x/y/z` whether it's an "array" or "dataframe" or "node". We have received feedback from several directions that this is very confusing.

I propose to **change**
* `StructureFamily.node` -> `StructureFamily.container`
* `tiled.client.Node` -> `tiled.client.Container`

This is an important change but it is a fairly targeted one that I think will not affect very much code. Some aspects can be made backward-compatible, but not all.

Other names considered but rejected in favor of "container":
* tree
* mapping
* group
* collection

Discussed with @dylanmcreynolds

danielballan:
Closes #472.

To be determined: Should we update the client module and class `tiled.client.node.Node` to `tiled.client.container.Container`? I lean "yes" but I do not feel 100% sure. It is technically still a Node, but it is _specifically_ a Container node, so it should be updated, right?


danielballan:
This also seemed like the right time to rename the queries on specs and structure families to `SpecQuery` and `StructureFamilyQuery`, to avoid confusion with the `Spec` dataclass and the `StructureFamily` enum. They are in different modules, of course, so they don't technically _collide_ but this seemed likely to be a source of confusion for users.

pshafer-als:
> To be determined: Should we update the client module and class `tiled.client.node.Node` to `tiled.client.container.Container`?

Is there an existing or likely usage where you could meaningfully replace `def do_something_with(node: Node): …` with e.g. `def do_something_with(node: DataFrame): …` or `def do_something_with(node: Array): …`?  

If not, then you can comfortably replace `…Node` with `…Container`.

pshafer-als:
I suppose a nice way to transition the node module and its classes would be to create aliases (`Node = Container`) and a module-level deprecation warning.

danielballan:
Done!

To make this change _fully_ backward-compatible would get messy very quickly. At this early stage, I think it's better to avoid loading up the codebase with that mess. I am opting to make only a basic effort here---raising clear error message / warnings at least.

There is a corresponding update in Databroker, ready to go once Tiled is released.

PR #492, opened 2023-07-07
CI: Use pip cache.

danielballan:
The Windows build takes more than 6 minutes to install the depdencies---longer than a Linux makes takes the run end-to-end sometimes. Apparently `v4` of the `setup-python` GH Action supports pip caching. I hope that will help.


danielballan:
The seems to have worked. The Windows build now takes only 4 minutes.

PR #495, opened 2023-07-08
Include formats in server deps.

### LINKED ISSUE #493: tifffile docker requirements
With the newer changes to the a97 and a98, the docker container is not picking up tifffile as an installed package. I'm guessing that with the recent change to using pyproject.toml, that tifffile was in the `[server]` install before and is not now. https://github.com/bluesky/tiled/blob/80f3a0c7c1fac3a9588c7a5eaf66a9d236db33f7/pyproject.toml#L252

danielballan:
Closes #493

Indeed, it looks I neglected to merge the `formats` section into the aggregate `server` section. Here, they are merged and then sorted.


danielballan:
Latest release:

```
$ podman run ghcr.io/bluesky/tiled:v0.1.0a98 python -c "import tifffile"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'tifffile'
```

This PR:

```
$ podman run localhost/tiled python -c "import tifffile"
$
```

Issue #473, opened 2023-06-24
Run client against catalog in-process


danielballan:
Consider the following use case.

User connects to a remote Tiled server.
```py
from tiled.client import from_uri

client = from_uri("https://tiled.nsls2.bnl.gov")
...
```

Perhaps a client-side cache is engaged---maybe by default---and automatically stashes locally any data or metadata they browse, up to a certain size, just as a web browser cache does.

Then, they decide they want to download a swath of the data for local, offline use. Something like:

```py
# This does not exist---just a proposal.
from tiled.client import download

download(client.search(...), "stuff/")
```

That could also be available as a CLI (`tiled download ...`) or a button in a web app. However it happens, suppose that this creates a zip archive or directory with contents like:

```
stuff/
    catalog.db
    data/
        ...
```

where `data/` contains files. These would be the same files backing Tiled on the server side, perhaps exactly the files the detector wrote. (Notice that if there were a client-side cache engaged, `download(...)` would naturally use it, so the user would not be downloading anything twice.)

The user can then use this local archive in three different ways.

## Option 1: Just the files, please

Open the files in `data/`, and just ignore `catalog.db` and Tiled.

## Option 2: Local Tiled Server

Run a local tiled server against this data

```
tiled serve catalog --public stuff/
```

Then navigate to `http://localhost:8000` in a web browser, or connect to it from any other program, or use the Tiled Python client:

```py
from tiled.client import from_uri

client = from_uri("http://localhost:8000")
```

## Option 3: In-process access

But if we want to access the local data _from Python specifically_ it's not even necessary to start a server in a separate process with `tiled serve catalog ...`. We can skip that step and do everything from one Python process.

```py
from tiled.client import from_catalog

client = from_catalog("stuff/catalog.db", readable_storage=["stuff/data"])
```

The above uses ASGI to run a "server" and the client in the user's Python process, passing HTTP messages via Python function calls within one process instead of TCP packets between separate server and client processes.

Because this runs in a single process, it can easily be wrapped up in third-party convenience libraries which can be as "magical" or explicit as one wants

```py
from nsls2_data_thingie import remote_access, download, local_access

client = remote_access()

download()  # perhaps downloads to some default location, like ~/.cache/nsls2_data_thingie
client = local_access()
```

danielballan:
[...]

PR #497, opened 2023-07-08
Rewrite client-side HTTP response cache

### LINKED ISSUE #281: Refactor client cache to use httpx transport

danielballan:
Closes #281 

- **Massively** simplify `tiled.client.Context` by moving the caching and decompression logic from the context down into the `httpx.Transport` layer. Basically, it now sits _inside_ the `httpx.Client` instead of wrapping _around_ it. I have learned that this is the kind of thing httpx intends us to do.
- Use a client HTTP response cache by default. It has always been opt-in; now it's opt-out.
- Use a SQLite for cache storage, by default located at `~/.cache/tiled/http_response_cache.db` (or wherever` XDG_CACHE_HOME` says).
- Expose minimal configuration with sensible defaults. Users should never have to notice that this exists.
- Remove "offline mode" and "download", which were misfeatures. They demoed well, but they were not useful in practice. A better approach to this use case is coming in #473.

Needs unit tests.


danielballan:
Moving ahead with this, as I've discussed the design with @tacaswell and @dylanmcreynolds _ad nauseum_ over the last year. There are sure to be bugs, but I think less bugs than the current implementation.

danielballan:
Also, check out that negative diff! Wooo deleting code....

PR #504, opened 2023-07-10
Make decoders work for chunked responses.

danielballan:
Also, add a basic unit test for the custom decoders, zstd and blosc.


danielballan:
Windows CI hit a transient failure on an unrelated test:

test_collision_after_startup

PR #517, opened 2023-07-13
Add Python 3.11 to CI matrix.

danielballan:
This was held up by the delayed numba release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
         - "3.8"
         - "3.9"
         - "3.10"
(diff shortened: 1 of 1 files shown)


danielballan:
The 3.11 CI passed. The documentation build failed because the matplotlib website is down so the intersphinx could not operate:

```
intersphinx inventory 'https://matplotlib.org/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 522 Server Error:  for url: https://matplotlib.org/objects.inv
make: *** [Makefile:20: html] Error 2
```

PR #533, opened 2023-07-28
More docker examples

danielballan:
I have not yet tested this.

### DIFF
--- a/docs/source/how-to/docker.md
+++ b/docs/source/how-to/docker.md
@@ -133,7 +133,7 @@ docker run --name tiled-test-postgres -p 5432:5432 -e POSTGRES_PASSWORD=secret -
 mkdir config/
 ```
 
-Place a copy of `example_config/single_user_single_catalog.yml`, from the Tiled
+Place a copy of `example_configs/single_catalog_single_user.yml`, from the Tiled
 repository root, in this `config/` directory.
 
 Replace the line:
@@ -167,7 +167,7 @@ Register the files in the directory `files/` with this catalog.
 docker run \
   -e TILED_SINGLE_USER_API_KEY=secret \
   -e TILED_DATABASE_PASSWORD=secret \
-  -v ./config:/deploy/config:ro
+  -v ./config:/deploy/config:ro \
   -v ./files:/files:ro \
   ghcr.io/bluesky/tiled:latest \
   tiled catalog register /storage/catalog.db /files --verbose
--- a/docs/source/how-to/docker.md
+++ b/docs/source/how-to/docker.md
@@ -64,7 +64,7 @@ quickly generate some with:
 
 ```
 # Optional: Generate sample files... TIFF, Excel, HDF5, etc.
-python -m tiled.examples.generate_files files/
+python -m tiled.examples.generate_files data/
 ```
 
 ### Quick Start (Not Scalable)
@@ -76,9 +76,9 @@ at server startup.
 docker run \
   -p 8000:8000 \
   -e TILED_SINGLE_USER_API_KEY=secret \
-  -v ./files:/files:ro \
+  -v ./data:/data:ro \
   ghcr.io/bluesky/tiled:latest \
-  tiled serve directory --host 0.0.0.0 /files
+  tiled serve directory --host 0.0.0.0 /data
 ```
 
 Two problems with this one-line approach:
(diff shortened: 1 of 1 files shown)


danielballan:
I tested the first two sections. Now blocked on

```
ModuleNotFoundError: No module named 'asyncpg'
```

We do include `asyncpg` as a server dep now, so we will be able to make progress after the next release.

danielballan:
Now stuck on some podman networking thing.

```
$ podman run -it   -p 8000:8000   -e TILED_SINGLE_USER_API_KEY=secret   -e TILED_DATABASE_PASSWORD=secret   -v ./config:/deploy/config:ro   -v ./files:/files:ro   ghcr.io/bluesky/tiled:latest /bin/bash
```

```
root@1a9322ee3886:/deploy# tiled catalog init --if-not-exists postgresql+asyncpg://postgres:secret@localhost:5432
<snipped>
ConnectionRefusedError: [Errno 111] Connect call failed ('127.0.0.1', 5432)
```

clintonroy:
in a docker compose setup we would have postgres as a dns entry inside the container, as a wild guess, inside the container 127.0.0.1 refers to the container itself, rather than your host?

if your development machine has a real ip, try that. maybe try..host network mode

(I'm not super confident of my analysis here :)

danielballan:
Good idea, thanks @clintonroy. Will try that next time I pick this up.

danielballan:
Thanks @clintonroy---`--net=host` did the trick.

danielballan:
The version of this page in `main` was originally written before Tiled supported writing. Thus, it focused on server a (read-only) directory of files, rather than accepting write requests backed by a SQL database and a writable volume.

I wonder if we should split the guide in some way to separate those two use cases. As written, it may be a bit confusing.

dylanmcreynolds (commented):
This looks good. One note I have is that "files" and "storage" are pretty close in meaning. How about:
- data and storage
- data and database
- files and database

docs/source/how-to/docker.md
```diff
@@ -67,17 +67,128 @@ quickly generate some with:
...
   -e TILED_SINGLE_USER_API_KEY=secret \
-  -v ./files:/files:ro ghcr.io/bluesky/tiled:latest \
+  -v ./files:/files:ro \
+  ghcr.io/bluesky/tiled:latest \
   tiled serve directory --host 0.0.0.0 /files
 ```
 
+Two problems with this one-line approach:
+
+* If you restart the server, all the indexing work is done from scratch.
```
padraic-shafer, line 86:
```suggestion
* If you restart the server, all the indexing work is re-done from scratch.
```

docs/source/how-to/docker.md
```diff
@@ -67,17 +67,128 @@ quickly generate some with:
...
+Initialize the database. (This creates the tables, indexes, and so on used by Tiled.)
+
+```
+export TILED_DATABASE_PASSWORD=secret
+export TILED_DATABASE_URI=postgresql+asyncpg://postgres:${TILED_DATABASE_PASSWORD}@localhost:5432
+
+docker run --net=host ghcr.io/bluesky/tiled:latest tiled catalog init $TILED_DATABASE_URI
+```
+
+Create a directory for TIled configuration, e.g. `config/`.
```
padraic-shafer, line 141:
```suggestion
Create a directory for Tiled configuration, e.g. `config/`.
```

docs/source/how-to/docker.md
```diff
@@ -67,17 +67,128 @@ quickly generate some with:
...
+uri: "sqlite+aiosqlite:////storage/catalog.db"
+```
+
+with a PostgreSQL database URI, such as:
+
+```yaml
+uri: "postgresql+asyncpg://postgres:${TILED_DATABASE_PASSWORD}@localhost:5432"
+```
+
+Start the server, potentially multiple across many hosts.
```
padraic-shafer, line 163:
```suggestion
Start the server, potentially multiple servers across many hosts.
```

padraic-shafer: Eventually, it would be useful to include in a cookbook the example of setting up multiple servers with a shared db.

E.g., does each server need to have the same config settings?

padraic-shafer (approved):
This looks good.

I agree with @dylanmcreynolds 's comments about "files"/"storage". 

In a similar vein, using 'secret' as both the Tiled key and the db password could give the impression that these services MUST use the same secret.  Consider distinguishing them here (`secret` and `db_secret`) as a clarification for general learners.

PR #534, opened 2023-07-28
Use arrow_schema in place of meta, divisions.

### LINKED ISSUE #520: Use Arrow schema in structure, in place of meta
The structure of a `dataframe` is provided as:

- `meta`, an Arrow encoded empty DataFrame that has the same column names and data types as the original
- `divisions` an Arrow encoding DataFrame with a column containing the index values at the boundaries between partitions

This was prompted very directly by what is needed to construct a dask DataFrame. Today I confirmed what I have suspected for awhile: we can and should use the Arrow schema; we should replace `meta` with `schema` and use that information to construct what dask needs.

The schema has no JSON representation. For now, at least, we will have to continue to base64-encode it. But maybe in the future we could revive this work, which seems viable: https://github.com/apache/arrow/pull/7110

I believe `divisions` is optional for dask, and I think `divisions` may not actually be being used by any current workflows. I propose to just drop it for now and put it back if and when we have a use case.

danielballan:
Closes #520

### DIFF
--- a/docs/source/explanations/structures.md
+++ b/docs/source/explanations/structures.md
@@ -256,31 +256,17 @@ $ http :8000/metadata/long_table | jq .data.attributes.structure
     "resizable": false
   },
   "micro": {
-    "meta": "data:application/vnd.apache.arrow.file;base64,...",
-    "divisions": "data:application/vnd.apache.arrow.file;base64,...",
+    "schema": "data:application/vnd.apache.arrow.file;base64,...",
   }
 }
 ```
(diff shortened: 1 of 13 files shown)


danielballan:
This removes a bunch of special-case conditionals and generally simplifies things a lot. I am happy with this change.

danielballan:
I got a conceptual 👍 from @tacaswell and @dylanmcreynolds when I pitched this, so it is going in.

PR #547, opened 2023-08-08
Simplify some APIs

danielballan:
Some nice minor simplifications here

* Remove `BaseStructureClient`; all we need is `BaseClient`.
* Simplify `init_storage` API to take the argument `structure` rather than various individual attributes of structure, which may change over time, or be different across different adapters.


danielballan:
Merging, as this is just cleaning up my own extraneous complexity. :-)

danielballan:
Got a quick conceptual review from @tacaswell

PR #548, opened 2023-08-08
Fix bug where sparse writing assumed 2D array.

danielballan:


danielballan:
Includes unit test that reproduces the bug and then a fix

Issue #450, opened 2023-06-13
Add support for AwkwardArray structures


danielballan:
@tacaswell and I have wanted to add support for AwkwardArrays from the start, but I think we do not have a GH Issue for it yet.

Notes from chat with @jpivarski a couple weeks ago...

Requirements:
* We would like to support upload and download of AwkwardArray structures.
* In the Python client we would like the option to access the data with or without `dask-awkward`.
* In normal Tiled fashion, we would like to be able to download a specific _slice_ of interest, and we would like the Tiled server to be able to only read, serializing, and transmit the specific slice of interest.

Proposed Approach:
* We considered using Arrow to transport AwkwardArrays between client and server. However, representing Awkward in Arrow blurs out detailed form information. Specifically, it loses the `form_key` that can be used to address specific buffers.
* Instead, we will operate directly on AwkwardArray's own representation, which comprises JSON-encodable `form`, outer `length` (an integer), and a dict-like `container` whose keys are referenced in the `form` and whose values are buffers.
* By reusing the typetracer machinery in awkward (which was developed to support dask-awkward) we can project a slice into a form and get a "projected form". The example below illustrates this, and uses only one piece of internal awkward API (`_touch_data`). This could conceivably be made into a public method.
* HTTP endpoints may look roughly like:
  ```
  /api/v1/awkward/full/{path_to_dataset}?slice=...
  /api/v1/awkward/buffers/{path_to_dataset}?form_key=...&form_key=...&form_key=...&slice=...
  ```
* Multiple buffers may be encoded in a container format like TAR or ZIP (not necessarily compressed, just used as a container). @jakirkham pointed out an advantage of ZIP: web browsers understand it.

Code snippet:

```py
import numpy as np
import awkward as ak

# The array we want to talk about.
array = ak.Array(
    [[{"x": 1.1, "y": [1]}, {"x": 2.2, "y": [1, 2]}], [], [{"x": 3.3, "y": [1, 2, 3]}]]
)

# On the server, you separately store form, length, and all the named buffers.
form, length, container = ak.to_buffers(array)

# When a client wants a lazy (slice-only) object, send them the form+length
# and keep a type-tracer array in your TiledObject's metadata.
meta_step1 = ak.Array(
    form.length_zero_array(highlevel=False).to_typetracer(forget_length=True)
)

# Second type-tracer array will tell us the set of buffers that the result of
# the slice will need, so that the client can make multiple requests.
[...]

PR #549, opened 2023-08-09
Sketch support for writing, reading sliced AwkwardArrays.

danielballan:
First crack at #450. This uses the `project_form` approach that @jpivarski sketched back in May.

To take this PR branch for a spin...

```
pip install -e .[server,client]
```

Start a Tiled server. Using a stable API key (instead of random cryptographic) is nice for development.

```
TILED_SINGLE_USER_API_KEY=secret tiled serve catalog --temp
```

Use it from the Python client.

```py
import awkward as ak

c = from_uri("http://localhost:8000", api_key="secret")
array = ak.Array(
    [[{"x": 1.1, "y": [1]}, {"x": 2.2, "y": [1, 2]}], [], [{"x": 3.3, "y": [1, 2, 3]}]]
)

# This sends two requests: (1) metadata and structure as JSON (2) zipped buffers
# The buffers could (later) be chunked across multiple requests / parallelized as we do for plain arrays.
c.write_awkward(array, key="test")

# This could be in a separate process to fully illustrate sharing the data between processes...
aac = c["test"]
aac.read()
aac[:]  # equivalent to read()
aac[0, "y", 1:]  # fetches only the form keys it needs, fewer bytes transferred
```

It may be more interesting to use `metadata` to make the data findable, instead of or in addition to giving it a `key`. (If the `key` is unspecified, a UUID4 key is provided by the server.)

```py
c.write_awkward(array, metadata={"color": "blue"})
from tiled.queries import Key
c.search(Key("color") == "blue")
c.search(Key("color") == "blue").values().first().read()
```

The messages written to stderr by the server at startup include the location of the temp dir being used. There, we can find the buffers:

```
$ ls /tmp/tmphxumip5h/data
total 27
drwxrwxr-x 3 dallan dallan 3 Aug  8 23:06 ./
drwxrwxr-x 3 dallan dallan 4 Aug  8 23:06 ../
drwxrwxr-x 2 dallan dallan 6 Aug  8 23:06 test/
$ ls /tmp/tmphxumip5h/data/test/
total 38
drwxrwxr-x 2 dallan dallan  6 Aug  8 23:06 ./
drwxrwxr-x 3 dallan dallan  3 Aug  8 23:06 ../
-rw-rw-r-- 1 dallan dallan 32 Aug  8 23:06 node0-offsets
-rw-rw-r-- 1 dallan dallan 24 Aug  8 23:06 node2-data
-rw-rw-r-- 1 dallan dallan 32 Aug  8 23:06 node3-offsets
-rw-rw-r-- 1 dallan dallan 48 Aug  8 23:06 node4-data
```

and the SQLite database:

```
$ sqlite3 /tmp/tmphxumip5h/catalog.db 
SQLite version 3.37.2 2022-01-06 13:25:41
Enter ".help" for usage hints.
sqlite> .mode columns
sqlite> select * from nodes;
id  key                                   ancestors  structure_family  metadata          specs  time_created         time_updated       
--  ------------------------------------  ---------  ----------------  ----------------  -----  -------------------  -------------------
1   test                                  []         awkward           {}                []     2023-08-09 03:06:50  2023-08-09 03:06:50
2   14c2ccde-3b0b-4f31-b5b2-b772b7932a26  []         awkward           {"color":"blue"}  []     2023-08-09 03:09:49  2023-08-09 03:09:49
sqlite> select * from data_sources;
id  node_id  structure                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          mimetype         parameters  management  time_created         time_updated       
--  -------  ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------  ---------------  ----------  ----------  -------------------  -------------------
1   1        {"length":3,"form":{"class":"ListOffsetArray","offsets":"i64","content":{"class":"RecordArray","fields":["x","y"],"contents":[{"class":"NumpyArray","primitive":"float64","inner_shape":[],"parameters":{},"form_key":"node2"},{"class":"ListOffsetArray","offsets":"i64","content":{"class":"NumpyArray","primitive":"int64","inner_shape":[],"parameters":{},"form_key":"node4"},"parameters":{},"form_key":"node3"}],"parameters":{},"form_key":"node1"},"parameters":{},"form_key":"node0"},"suffixed_form_keys":["node0-offsets","node2-data","node3-offsets","node4-data"]}  application/zip  {}          writable    2023-08-09 03:06:50  2023-08-09 03:06:50
2   2        {"length":3,"form":{"class":"ListOffsetArray","offsets":"i64","content":{"class":"RecordArray","fields":["x","y"],"contents":[{"class":"NumpyArray","primitive":"float64","inner_shape":[],"parameters":{},"form_key":"node2"},{"class":"ListOffsetArray","offsets":"i64","content":{"class":"NumpyArray","primitive":"int64","inner_shape":[],"parameters":{},"form_key":"node4"},"parameters":{},"form_key":"node3"}],"parameters":{},"form_key":"node1"},"parameters":{},"form_key":"node0"},"suffixed_form_keys":["node0-offsets","node2-data","node3-offsets","node4-data"]}  application/zip  {}          writable    2023-08-09 03:09:49  2023-08-09 03:09:49
sqlite> select * from assets;
id  data_uri                                                                    is_directory  hash_type  hash_content  size  time_created         time_updated       
--  --------------------------------------------------------------------------  ------------  ---------  ------------  ----  -------------------  -------------------
1   file://localhost/tmp/tmphxumip5h/data/test                                  1                                            2023-08-09 03:06:50  2023-08-09 03:06:50
2   file://localhost/tmp/tmphxumip5h/data/14c2ccde-3b0b-4f31-b5b2-b772b7932a26  1                                            2023-08-09 03:09:49  2023-08-09 03:09:49
```

The writing feature in general has only been around since early summer, so there are sure to be bugs, but it's a start. More context, including how to start a server with persistent storage, is [in the docs](https://blueskyproject.io/tiled/tutorials/writing.html).

One question: Is there a clever way to associate the "form keys" like `node0` with their suffixed variants like `node0-offsets`? To get something working, I opted to store `suffixed_form_keys` (list of strings) alongside the JSON `form` and the `length`. That works fine, but there may be a better way.

### DIFF
--- a/tiled/client/awkward.py
+++ b/tiled/client/awkward.py
@@ -29,7 +29,7 @@ class AwkwardArrayClient(BaseClient):
         )
         proxy_array = awkward.Array(typetracer)
         # TODO Ask awkward to promote _touch_data to a public method.
-        ak.typetracer.touch_data(proxy_array[slice])
+        awkward.typetracer.touch_data(proxy_array[slice])
         form_keys_touched = set(report.data_touched)
         projected_form = project_form(form, form_keys_touched)
         # The order is not important, but sort so that the request is deterministic.
--- a/tiled/client/awkward.py
+++ b/tiled/client/awkward.py
@@ -3,7 +3,7 @@ import awkward
 from ..serialization.awkward import from_zipped_buffers, to_zipped_buffers
 from ..structures.awkward import project_form
 from .base import BaseClient
-from .utils import handle_error
+from .utils import export_util, handle_error, params_from_slice
 
 
 class AwkwardArrayClient(BaseClient):
@@ -13,10 +13,12 @@ class AwkwardArrayClient(BaseClient):
         return f"<{type(self).__name__}>"
 
     def write(self, container):
+        structure = self.structure()
+        components = (structure.form, structure.length, container)
         handle_error(
             self.context.http_client.put(
                 self.item["links"]["full"],
-                content=bytes(to_zipped_buffers(container, {})),
+                content=bytes(to_zipped_buffers(components, {})),
                 headers={"Content-Type": "application/zip"},
             )
         )
@@ -41,7 +43,7 @@ class AwkwardArrayClient(BaseClient):
                 params=params,
             )
         ).read()
-        container = from_zipped_buffers(content)
+        container = from_zipped_buffers(content, projected_form, structure.length)
         projected_array = awkward.from_buffers(
             projected_form, structure.length, container
         )
@@ -49,3 +51,43 @@ class AwkwardArrayClient(BaseClient):
 
     def __getitem__(self, slice):
         return self.read(slice=slice)
+
+    def export(self, filepath, *, format=None, slice=None, template_vars=None):
+        """
+        Download data in some format and write to a file.
+
+        Parameters
+        ----------
+        file: str or buffer
+            Filepath or writeable buffer.
+        format : str, optional
+            If format is None and `file` is a filepath, the format is inferred
+            from the name, like 'table.csv' implies format="text/csv". The format
+            may be given as a file extension ("csv") or a media type ("text/csv").
+        slice : List[slice], optional
+            List of slice objects. A convenient way to generate these is shown
+            in the examples.
+        template_vars: dict, optional
+            Used internally.
+
+        Examples
+        --------
+
+        Export all.
+
+        >>> a.export("numbers.csv")
+
+        Export an N-dimensional slice.
+
+        >>> import numpy
+        >>> a.export("numbers.csv", slice=numpy.s_[:10, 50:100])
+        """
+        template_vars = template_vars or {}
+        params = params_from_slice(slice)
+        return export_util(
+            filepath,
+            format,
+            self.context.http_client.get,
+            self.item["links"]["full"].format(**template_vars),
+            params=params,
+        )
--- a/tiled/client/awkward.py
+++ b/tiled/client/awkward.py
@@ -36,7 +36,7 @@ class AwkwardArrayClient(BaseClient):
         params = {"form_key": sorted(list(form_keys_touched))}
         content = handle_error(
             self.context.http_client.get(
-                self.item["links"]["full"],
+                self.item["links"]["buffers"],
                 headers={"Accept": "application/zip"},
                 params=params,
             )
--- a/tiled/client/awkward.py
+++ b/tiled/client/awkward.py
@@ -44,7 +44,10 @@ class AwkwardClient(BaseClient):
         ).read()
         container = from_zipped_buffers(content, projected_form, structure.length)
         projected_array = awkward.from_buffers(
-            projected_form, structure.length, container
+            projected_form,
+            structure.length,
+            container,
+            allow_noncanonical_form=True,
         )
         return projected_array[slice]
 
--- a/tiled/client/awkward.py
+++ b/tiled/client/awkward.py
@@ -30,7 +30,6 @@ class AwkwardArrayClient(BaseClient):
             form,
         )
         proxy_array = awkward.Array(typetracer)
-        # TODO Ask awkward to promote _touch_data to a public method.
         awkward.typetracer.touch_data(proxy_array[slice])
         form_keys_touched = set(report.data_touched)
         projected_form = project_form(form, form_keys_touched)
--- a/tiled/client/awkward.py
+++ b/tiled/client/awkward.py
@@ -6,7 +6,7 @@ from .base import BaseClient
 from .utils import export_util, handle_error
 
 
-class AwkwardArrayClient(BaseClient):
+class AwkwardClient(BaseClient):
     def __repr__(self):
         # TODO Include some summary of the structure. Probably
         # lift __repr__ code from awkward itself here.
--- a/tiled/client/awkward.py
+++ b/tiled/client/awkward.py
@@ -1,0 +1,52 @@
+import awkward
+
+from ..serialization.awkward import from_zipped_buffers, to_zipped_buffers
+from ..structures.awkward import project_form
+from .base import BaseClient
+from .utils import handle_error
+
+
+class AwkwardArrayClient(BaseClient):
+    def __repr__(self):
+        # TODO Include some summary of the structure. Probably
+        # lift __repr__ code from awkward itself here.
+        return f"<{type(self).__name__}>"
+
+    def write(self, container):
+        handle_error(
+            self.context.http_client.put(
+                self.item["links"]["full"],
+                content=bytes(to_zipped_buffers(container, {})),
+                headers={"Content-Type": "application/zip"},
+            )
+        )
+
+    def read(self, slice=...):
+        structure = self.structure()
+        form = awkward.forms.from_dict(structure.form)
+        typetracer, report = awkward.typetracer.typetracer_with_report(
+            form,
+            forget_length=True,
+        )
+        proxy_array = awkward.Array(typetracer)
+        # TODO Ask awkward to promote _touch_data to a public method.
+        proxy_array[slice].layout._touch_data(recursive=True)
+        form_keys_touched = set(report.data_touched)
+        projected_form = project_form(form, form_keys_touched)
+        # The order is not important, but sort so that the request is deterministic.
+        params = {"form_key": sorted(list(form_keys_touched))}
+        content = handle_error(
+            self.context.http_client.get(
+                self.item["links"]["full"],
+                headers={"Accept": "application/zip"},
+                params=params,
+            )
+        ).read()
+        container = from_zipped_buffers(content)
+        projected_array = awkward.from_buffers(
+            projected_form, structure.length, container
+        )
+        return projected_array[slice]
+
+    def __getitem__(self, slice):
+        return self.read(slice=slice)
--- a/tiled/client/awkward.py
+++ b/tiled/client/awkward.py
@@ -29,7 +29,7 @@ class AwkwardArrayClient(BaseClient):
         )
         proxy_array = awkward.Array(typetracer)
         # TODO Ask awkward to promote _touch_data to a public method.
-        proxy_array[slice].layout._touch_data(recursive=True)
+        ak.typetracer.touch_data(proxy_array[slice])
         form_keys_touched = set(report.data_touched)
         projected_form = project_form(form, form_keys_touched)
         # The order is not important, but sort so that the request is deterministic.
--- a/tiled/client/awkward.py
+++ b/tiled/client/awkward.py
@@ -3,7 +3,7 @@ import awkward
 from ..serialization.awkward import from_zipped_buffers, to_zipped_buffers
 from ..structures.awkward import project_form
 from .base import BaseClient
-from .utils import export_util, handle_error, params_from_slice
+from .utils import export_util, handle_error
 
 
 class AwkwardArrayClient(BaseClient):
@@ -52,7 +52,7 @@ class AwkwardArrayClient(BaseClient):
     def __getitem__(self, slice):
         return self.read(slice=slice)
 
-    def export(self, filepath, *, format=None, slice=None, template_vars=None):
+    def export(self, filepath, *, format=None):
         """
         Download data in some format and write to a file.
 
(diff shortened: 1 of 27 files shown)


danielballan:
The test failure on Python 3.8 may be related to [an issue with backing `ZipFile` objects with in-memory buffers](https://stackoverflow.com/q/5624669/1221924). Needs investigation.

jpivarski:
This looks great! I recognize that it's still a work in progress, but can I share it on the IRIS-HEP Slack so that more people (including some likely to want to use it) will know about it?

> One question: Is there a clever way to associate the "form keys" like `node0` with their suffixed variants like `node0-offsets`? To get something working, I opted to store `suffixed_form_keys` (list of strings) alongside the JSON `form` and the `length`. That works fine, but there may be a better way.

The mapping from a node's `form_key` to the buffer names used by [ak.from_buffers](https://awkward-array.org/doc/main/reference/generated/ak.from_buffers.html) is controlled by the `buffer_key` argument, which can be a format string or a whole Python function. The issue is that individual Form nodes have `form_keys`, but some types of Form nodes have multiple buffers. the "`{attribute}`" part of that mapping completes it.

These are all of the Form node types and all of their attributes:

| Form node type[^1] | attributes |
|:--|:--|
| [NumpyForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.NumpyForm.html) | `data` |
| [EmptyForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.EmptyForm.html) | _(none)_ |
| [ListOffsetForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.ListOffsetForm.html) | `offsets` |
| [ListForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.ListForm.html) | `starts`, `stops` |
| [RegularForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.RegularForm.html) | _(none)_ |
| [IndexedForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.IndexedForm.html) | `index` |
| [IndexedOptionForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.IndexedOptionForm.html) | `index` |
| [ByteMaskedForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.ByteMaskedForm.html) | `mask` |
| [BitMaskedForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.BitMaskedForm.html) | `mask` |
| [UnmaskedForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.UnmaskedForm.html) | _(none)_ |
| [RecordForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.RecordForm.html) | _(none)_ |
| [UnionForm](https://awkward-array.org/doc/main/reference/generated/ak.forms.UnionForm.html) | `tags`, `index` |

If you never have `ListArray` data or `UnionArray` data, you could set `buffer_key="{form-key}"` and the `form_key` would be one-to-one with the buffer names. It's pretty easy to end up with `ListArray` data:

```python
>>> original = ak.Array([[0, 1, 2], [], [3, 4], [5], [6, 7, 8, 9]])
>>> type(original.layout)
<class 'awkward.contents.listoffsetarray.ListOffsetArray'>
>>> rearranged = original[[3, 4, 0, 1, 1, 3]]
>>> rearranged
<Array [[5], [6, 7, 8, 9], [0, 1, 2], [], [], [5]] type='6 * var * int64'>
>>> type(rearranged.layout)
<class 'awkward.contents.listarray.ListArray'>
```

but it's also easy to avoid it with [ak.to_packed](https://awkward-array.org/doc/main/reference/generated/ak.to_packed.html), a function we would likely apply to any data that will be saved or transmitted somewhere, since it removes hidden values (useful for zero-copy, not useful when we know the array will be copied for other reasons).

```python
>>> packed = ak.to_packed(rearranged)
>>> packed
<Array [[5], [6, 7, 8, 9], [0, 1, 2], [], [], [5]] type='6 * var * int64'>
>>> type(packed.layout)
<class 'awkward.contents.listoffsetarray.ListOffsetArray'>
```

`UnionArray`, on the other hand, is unavoidable: it's the only way to express heterogeneous data, and it needs both `tags` and `index`. If it helps, maybe this could be the only exception to the one-to-one mapping from `form_key` to buffer names:

```python
ak.from_buffers(
    ...
    buffer_key="{form-key}{'-tags' if attribute == 'tags' else ''}",
)
```

[^1]: Same as the value of the `"class"` key in the Form's JSON format by replacing `s/Form/Array/`.

danielballan:
Oh, thanks, this is super helpful. Makes sense.

Yes, I’d be happy for it to be shared. Naturally, I would not advise relying on it for anything important at least until it is in a release, but it is great if people are game to play with it.

danielballan:
Two follow-up questions, @jpivarski, if you have a moment:

1. Does https://github.com/bluesky/tiled/pull/549/commits/1531002b3bfd7e7e032f09e57770d1fba17f97f4 look sensible? The goal is, "Given a form, extract what the keys in `container` would be. Those are the file names of the buffers we want to read." Glancing through the awkward source, I could not find a ready-made utility for this. If one exists, I'll use it, but it was easy enough to write one.
2. Would you be up for implementing support for `UnionForm` in `project_form`? In Seattle, [we left it as a `NotImplementedError`](https://github.com/bluesky/tiled/pull/549/files#diff-de1b0af7951a759d0aa37300fb76a92bbad0f923020842affab65c00c2b98c13R37-R38), waiting to prove out the concept before diving into those details. I think you mentioned some complexity arose related to re-instating some forms that were removed in previous iterations.

jpivarski:
> 1. Does [1531002](https://github.com/bluesky/tiled/commit/1531002b3bfd7e7e032f09e57770d1fba17f97f4) look sensible?

Yes, it does, although it also sounds like the sort of utility function we ought to provide. scikit-hep/awkward#2660 adds that as a method on `Form` (not too visible, since only a few developers will need it), with full generality, allowing for different `buffer_key` functions and providing the dtypes that will be assumed of the buffers, even though you don't need them for your case. You can use

```python
form.expected_from_buffers().keys()
```

(The only additional information one might want to know is the minimum expected lengths of 
these buffers, but you do need to read some of the buffers to know the expected lengths of other buffers, so there's no advantage over running `ak.from_buffers` directly.)

---------------

As for the `UnionForm` in `project_form`, I'll remind myself what was needed and think about whether something can be done about it.

tiled/structures/awkward.py
```diff
@@ -0,0 +1,80 @@
...
+                fields.append(field)
+                contents.append(content)
+
+        if form.fields is None:
+            fields = None
+
+        return form.copy(fields=fields, contents=contents)
+
+    elif isinstance(form, awkward.forms.UnionForm):
+        raise NotImplementedError
```
jpivarski, line 38: Eventually, you'll be able to

```suggestion
    elif isinstance(form, awkward.forms.UnionForm):
        step1 = [project_form(x, form_keys_touched) for x in form.contents]
        step2 = [x for x in step1 if x is not None]
        if len(step2) == 0:
            return None
        elif len(step2) == 1:
            return step2[0]
        else:
            return awkward.forms.UnionForm.simplified(
                form.tags,
                form.index,
                step2,
                parameters=form.parameters,
                form_key=form.form_key,
            )
```

but the `awkward.forms.UnionForm.simplified` will drop the `form_keys` of all the `Content` instances in `step2`. We've been living with that problem for too long: I'm going to open a PR on Awkward to fix that, and when that's merged and in a new release, the above implementation will be 100% correct. For now, it will only work if `len(step2) < 2`.

jpivarski: When/if scikit-hep/awkward#2661 goes into a release of Awkward Array, the above code will be correct.

tiled/client/awkward.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def read(self, slice=...):
+        structure = self.structure()
+        form = awkward.forms.from_dict(structure.form)
+        typetracer, report = awkward.typetracer.typetracer_with_report(
+            form,
+            forget_length=True,
+        )
+        proxy_array = awkward.Array(typetracer)
+        # TODO Ask awkward to promote _touch_data to a public method.
```
danielballan, line 32: xref https://github.com/scikit-hep/awkward/issues/2668

tiled/client/awkward.py
```diff
@@ -0,0 +1,52 @@
...
+                headers={"Content-Type": "application/zip"},
+            )
+        )
+
+    def read(self, slice=...):
+        structure = self.structure()
+        form = awkward.forms.from_dict(structure.form)
+        typetracer, report = awkward.typetracer.typetracer_with_report(
+            form,
+            forget_length=True,
```
agoose77, line 29: This argument is default `True`, and seeing it used here made me realise that we [probably want to deprecate it](https://github.com/scikit-hep/awkward/pull/2671). Could you remove the usage?
```suggestion
```

tiled/client/awkward.py
danielballan: https://github.com/scikit-hep/awkward/pull/2672

tiled/client/awkward.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def read(self, slice=...):
+        structure = self.structure()
+        form = awkward.forms.from_dict(structure.form)
+        typetracer, report = awkward.typetracer.typetracer_with_report(
+            form,
+        )
+        proxy_array = awkward.Array(typetracer)
+        # TODO Ask awkward to promote _touch_data to a public method.
+        proxy_array[slice].layout._touch_data(recursive=True)
```
agoose77, line 32: This will become
```suggestion
        ak.typetracer.touch_data(proxy_array[slice])
```
once a new release has been made :)

danielballan:
All of the enhancements to Awkward discussed above have been implemented and released as of awkward v2.4.0, with the exception of https://github.com/scikit-hep/awkward/issues/2666 (which, per a comment above, will enable support `UnionForm` where `len(step2) > 1`.

I have updated this PR to pin a minimum version of Awkward and to use those enhancements. It now relies on only public APIs.

I intend to do a little more refinement and then merge.

danielballan:
One more follow-up question for you, @jpivarski. While reading through the documentation on [saving to HDF5](https://awkward-array.org/doc/main/user-guide/how-to-convert-buffers.html#saving-awkward-arrays-to-hdf5) I bumped on this:

```py
group.attrs["length"] = json.dumps(length)  # JSON-encode it because it might be a list
```

This PR assumes `length` will always be an integer. Should it be a list of integers? Is the list ever nested?

I spent just a moment trying to contrive a scenario with a non-integer length and found the following:

```py
array = ak.Array([{"x": [1,2,3], "y": [4,5,6]}])
ak.to_buffers(array)  # works as expected
ak.to_buffers(array[0])  # TypeError: ak.Record objects are not allowed in this function
```

which makes me wonder if maybe `length` used to be more flexible but is now rigidly always an integer.

jpivarski:
The `length` used to be more flexible, but now it's always an integer, although it has never been possible to pass a scalar record (`ak.Record`) into `ak.to_buffers`. If we want to save scalar records, then we'd have to pass a length-1 RecordArray and carry an additional `bool` of metadata to know that it's supposed to be a Record and not a RecordArray. (We do that in `ak.to_parquet`/`ak.from_parquet`.)

The extra flexibility that `length` used to have is that Awkward 1.x allowed for partitioned arrays; the partitioning could only be applied to the top level of structure and would fill the `container` with different buffers for each partition. In that case, we need a length for every partition, and so `length` could be a list of integers in Awkward 1.x. That's no longer true, though, because dask-awkward does partitioning instead.

danielballan:
Oh, that makes perfect sense. Thanks.

Then I think this has converged to a reasonable first iteration. I will update the docs to include Awkward and then put this in.

danielballan:
TO DO:

- [ ] Add example to demo
- [ ] Update `structures.md`
- [ ] Update `README.md`
- [ ] Update API reference

tiled/structures/awkward.py
```diff
@@ -0,0 +1,67 @@
...
+                "Certain UnionForms are not yet supported. "
+                "See https://github.com/scikit-hep/awkward/issues/2666"
+            )
+            return awkward.forms.UnionForm.simplified(
+                form.tags,
+                form.index,
+                step2,
+                parameters=form.parameters,
+                form_key=form.form_key,
+            )
```
agoose77, line 55: @jpivarski @danielballan my understanding of the purpose of this routine is to build the form that reflects the array produced by a given slice. The benefit of doing this is to minimise reading.

This will be a problem for unions, because at present we want them to be erased if they're not actually required. This is once example of what we call "canonicalisation rules".

Consider a union of two records:
```
union[
   {x: int64, y: int32},
   {x: int64}
]
```
The following slice would erase the union
```python
array["x"]
```
In doing so, we'd copy and merge the buffers, i.e. we can't directly load the resulting array using `from_buffers`.

I think this presents a problem with the approach taken here.

In this instance, I have an idea; don't simplify the form *at all*, and use the private `simplify` argument to `ak.from_buffers`. This would change the form of the result, but it *would* do what we want here (despite concatenating buffers if unions with >1 content end up being erased).

We need to discuss this - @jpivarski I'll ping you in Slack.

danielballan: Applied in 109592c8c868210f76f5cb389551711fd7f6f922

PR #553, opened 2023-08-21
Fixed search error in the documentation frontend

jmaruland:
This PR solve a search error in the documentation frontend, where any time a query was requested from the search bar, the website stayed hanging and never showed the results; it showed the searching message indefinitely and the error message in the browser was "Uncaught ReferenceError: jQuery is not defined".

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -45,6 +45,7 @@ extensions = [
     "sphinx_click",
     "sphinx_copybutton",
     "myst_parser",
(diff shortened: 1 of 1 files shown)


padraic-shafer:
@jmaruland Confirming that building the docs with your change worked for me locally.

Adding [this link](https://github.com/readthedocs/sphinx_rtd_theme/issues/1452#issuecomment-1490504991) for posterity, in case we wonder 2-years later why this configuration setting is present. :-)

PR #554, opened 2023-08-22
Use double quotes in installation instructions

danielballan:
Usage like:

```
pip intsall "tiled[client]"
```

works in the Linux default shell (bash), Mac default shell (`zsh`), and Windows command prompt. Using no quotes breaks in `zsh` and using single quotes breaks on Windows.

### DIFF
--- a/docs/source/tutorials/installation.rst
+++ b/docs/source/tutorials/installation.rst
@@ -27,42 +27,21 @@ Install Tiled from PyPI using pip.
 
 .. code:: bash
 
-   python3 -m pip install 'tiled[all]'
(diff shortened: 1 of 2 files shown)


danielballan:
@jmaruland Can you confirm again that double quotes works on Windows? Just double checking.

jmaruland:
@danielballan Quotes work on Windows, too. I tried it on a clean new environment, just in case.

jmaruland (approved):
@danielballan Quotes work on Windows, too. I tried it on a clean new environment, just in case.

PR #555, opened 2023-08-23
Allow digits in tiff-sequence file names

Wiebke:
Expands possible file name patters recognized as a tiff sequence to those including numbers, e.g.  `run2_20230823_0001.tiff`,`run2_20230823_0002.tiff`, ... to be registered as a node `run2_20230823_`

Sequences not containing any prefix (`0000.tif`, `0001.tif`, ...) or with a prefix containing just non-digits are still treated as before.

### DIFF
--- a/tiled/_tests/test_tiff.py
+++ b/tiled/_tests/test_tiff.py
@@ -64,6 +64,10 @@ def test_tiff_sequence_block(client, block_input, correct_shape):
 async def test_tiff_sequence_with_directory_walker(tmpdir):
     """
     directory/
+      00001.tif
+      00002.tif
+      ...
+      00010.tif
       single_image.tif
       image00001.tif
       image00002.tif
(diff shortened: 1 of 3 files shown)


Wiebke:
The bug mentioned in #551 for ordering assets also affects Tiff sequences and the test added in [bb1640f](https://github.com/bluesky/tiled/pull/555/commits/bb1640f69127c4dc2c48bb7b4a907fe1536d4f2c) fails most of the time. 

[6ea2a67](https://github.com/bluesky/tiled/pull/555/commits/6ea2a67aea19d7a78b5d172685cd4c64e2c91927) temporarily fixes this, by sorting the files names when creating the `TiffSequence`. This would be no longer needed if ordering is tracked elsewhere.  

Some additional thoughts/questions:

- Should we guard against non-zero padded indices by sorting on parsed sequence indices?
- If we parse numbers, should we keep track of min/max and warn about missing indices? e.g. if there are files `0000.tif` `0002.tif`, but no `0001.tif` these would be grouped nonetheless, but we would have the option to warn that there are indices between `min_index` (0) and `max_index` (2), but only `len(sequence)` instead of `max_index-min_index+1` (2 instead of 3) files.

jmaruland (approved):

danielballan:
OK. This is a significant improvement, so I agree we should go ahead with this. We can revisit when the Asset <--> DataSource association information is made richer.

PR #560, opened 2023-08-25
Fixes "await" order of execution typos in serialization.xarray

### LINKED ISSUE #559: Typo in xarray.serialize_json
There is a minor typo on line 100 of `tiled.serialization.xarray.serialize_json` in tiled version 0.1.0a104

This

```py
 df = await as_dataset(node).to_dataframe()
```

needs to be

```py
 df = (await as_dataset(node)).to_dataframe()
```

I can submit this as a PR but that seems like a lot for a typo. :)

martintb:
Fixes #559 

Minor typo in coroutine call causes json serialization to fail.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.2"
 services:
   tiled:
-    image: ghcr.io/bluesky/tiled:v0.1.0a104
+    image: ghcr.io/bluesky/tiled:v0.1.0a105
     environment:
       - TILED_SINGLE_USER_API_KEY=${TILED_SINGLE_USER_API_KEY}
     ports:
--- a/docs/source/tutorials/installation.rst
+++ b/docs/source/tutorials/installation.rst
@@ -27,42 +27,21 @@ Install Tiled from PyPI using pip.
 
 .. code:: bash
 
-   python3 -m pip install 'tiled[all]'
+   python3 -m pip install "tiled[all]"
 
 If you are connecting to an existing to a tiled server as a client, there
 is not need to install all the server-related dependencies.
 
 .. code:: bash
 
-   python3 -m pip install 'tiled[client]'  # client only
+   python3 -m pip install "tiled[client]"  # client only
 
 Likewise, if you are deploying a tiled server but not using the client, you can
 skip a couple client-related dependencies.
 
 .. code:: bash
 
-   python3 -m pip install 'tiled[server]'  # server only
-
-.. warning::
-
-   We use single quotes in the installation commands, as in:
-
-   .. code:: bash
-
-      python3 -m pip install 'tiled[all]'
-
-   On MacOS, you MUST include the single quotes.
-   On Windows, you MUST NOT include them; you should instead do:
-
-   .. code:: bash
-
-      python3 -m pip install tiled[all]
-
-   On Linux, you may include them or omit them; it works the same either way.
-
-   (This difference between Mac and Linux is not due to the operating system *per
-   se*. It just happens that modern MacOS systems ship with zsh as the default
-   shell, whereas Linux typically ships with bash.)
+   python3 -m pip install "tiled[server]"  # server only
 
 Minimal Installation (fewer dependencies)
 -----------------------------------------
@@ -75,8 +54,8 @@ exploring metadata.
 
 .. code:: bash
 
-   python3 -m pip install 'tiled[minimal-client]'
-   python3 -m pip install 'tiled[minimal-server]'
+   python3 -m pip install "tiled[minimal-client]"
+   python3 -m pip install "tiled[minimal-server]"
 
 See lists of dependencies in `pyproject.toml` in the repository root for
 details.
@@ -90,7 +69,7 @@ To install an editable installation for local development:
 
    git clone https://github.com/bluesky/tiled
    cd tiled
-   pip install -e '.[all]'
+   pip install -e ".[all]"
 
 Web UI
 ------
--- a/share/tiled/templates/index.html
+++ b/share/tiled/templates/index.html
@@ -21,7 +21,7 @@
     <h2 class="subtitle">
       Install and use Tiled's Python client from Python, IPython, or Jupyter.
     </h2>
-    <pre><code>$ pip install tiled[client]
+    <pre><code>$ pip install "tiled[client]"
 $ python
 >>> from tiled.client import from_uri
 >>> client = from_uri("{{ root_url }}")</code></pre>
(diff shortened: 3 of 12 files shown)


martintb:
I just noticed that [line 56](https://github.com/bluesky/tiled/pull/560/files#diff-2f12432d5b396133d4e213d0d9be146c707c3f25067d78d2bf1c0a25e811db61L59) will also likely have this same error.

I'm going to revamp this PR to include that fix too (and maybe attempt to add the associated test at the same time)

padraic-shafer:
Good catch!

[This section](https://docs.xarray.dev/en/stable/user-guide/io.html#netcdf) might be helpful toward a test that reads the data buffer and compares contents to the original dataset.  Scroll down to "We can load netCDF files to create a new Dataset using [open_dataset()](https://docs.xarray.dev/en/stable/generated/xarray.open_dataset.html#xarray.open_dataset):"

martintb:
Thanks Padraic! 

So I have a test that is a modification of your json test but I'm running into two issues that are keeping me from finalizing the changes. I'm having trouble getting `xarray.open_dataset` to read the memoryview returned by the `serialize_netcdf` method (i.e., `BytesIO.getbuffer()`).

```python
# simulate serialize_netcdf
file      = _BytesIOThatIgnoresClose()
out      = ds.to_netcdf(file,engine="scipy")
buffer = file.getbuffer()

# simulate deserialization/reconstruction of dataset
file2 = _BytesIOThatIgnoresClose(buffer)
xarray.open_dataset(file2, engine='scipy') # Raises KeyError in scipy.io._netcdf

xarray.open_dataset(buffer.tobytes(), engine='scipy') # Raises (same) KeyError in scipy.io._netcdf

```
If I write the file to disk and then read it everything works fine

```python
ds.to_netcdf('test.nc', engine="scipy")
with open('test.nc', 'rb') as f:
    bytes_ = f.read()
xarray.open_dataset(_BytesIOThatIgnoresClose(bytes_)) #returns correct dataset
```

weirdly, using the bytestring returned directly from xarrays to_netcdf function also works

```python
file      = _BytesIOThatIgnoresClose()
xarray_bytes = ds.to_netcdf(file,engine="scipy")

file2 = _BytesIOThatIgnoresClose(xarray_bytes)
xarray.open_dataset(file2, engine='scipy') # returns correct dataset
```
To further test this, I changed the `serialize_netcdf` method in `tiled.serialize` to return the bytes from `xarray.to_netcdf` and all tests pass including my new one.  If there is a strong reason for that method returning a memoryview, this is not a solution though.

padraic-shafer:
@martintb Thanks for working on this.  Your troubleshooting inspired me to find another approach that I believe does what we want.  I've submitted a [PR](https://github.com/martintb/tiled/pull/2) to you for this branch.

padraic-shafer:
> Edit: Well, I've installed pre-commit locally now to avoid the excess formatting commits below...

That was a good call to use `pre-commit` locally.  This and other developer tools will get installed when you run `python -m pip install -e '.[all,dev]'` in the root directory of your local repository.  'dev' gets you the development packages; 'all' gets you the tiled client and server packages ([and maybe a few extra goodies](https://github.com/bluesky/tiled/blob/ca9acdc1384dba3672ac83ea8a845f280dce5fb5/pyproject.toml#L44-L45)).

martintb:
> > Edit: Well, I've installed pre-commit locally now to avoid the excess formatting commits below...
> 
> 
> 
> That was a good call to use `pre-commit` locally.  This and other developer tools will get installed when you run `python -m pip install -e '.[all,dev]'` in the root directory of your local repository.  'dev' gets you the development packages; 'all' gets you the tiled client and server packages ([and maybe a few extra goodies](https://github.com/bluesky/tiled/blob/ca9acdc1384dba3672ac83ea8a845f280dce5fb5/pyproject.toml#L44-L45)).

Weirdly I had already done the editable pip install of the [dev] variant but pre-commit didn't show up as an executable until I manually pip installed it.

martintb:
> @martintb Thanks for working on this.  Your troubleshooting inspired me to find another approach that I believe does what we want.  I've submitted a [PR](https://github.com/martintb/tiled/pull/2) to you for this branch.

Merged! I'll point out that your solution still relies on my change to serialization.xarray.serialize_netcdf which changes the output from a memoryview to a bytestring. It's unclear if any other parts of the code rely on the memoryview but if they do it's not covered in the tests.

padraic-shafer:
> your solution still relies on my change to serialization.xarray.serialize_netcdf which changes the output from a memoryview to a bytestring

Oh, I missed that earlier.  Thanks for reminding me.  I'll take another look at it now.

padraic-shafer:
@danielballan In `serialize_netcdf()` why does `Dataset.to_netcdf()` write to a freshly created byte stream and then return a view into its buffer?  Per the `xarray.Dataset.to_netcdf()` documentation, we could instead have that call directly return a byte string by omitting the `path` parameter.

https://github.com/bluesky/tiled/blob/ca9acdc1384dba3672ac83ea8a845f280dce5fb5/tiled/serialization/xarray.py#L55-L60

I ask because:
1. A memory view would be efficient if the file/stream existed before the call to `serialize_netcdf()`, but that is not the case here.
2. It's not clear to me how to use the buffer (the result of `getbuffer()`) to re-create the dataset. [See the workaround](https://github.com/martintb/tiled/blob/cdffb266bf48cdfc3ab96c03a21d34906c7646f9/tiled/serialization/xarray.py#L55-L60) proposed by @martintb.
3. The byte string returned by `Dataset.to_netcdf()` is not quite equal to the value in the stream buffer -- the assertion below fails.  Most bytes are the same, but you can see where they differ by pasting this loop into `serialize_netcdf()`.

```python3
        for i,(a,b) in enumerate(zip(bytes_, bytes(file.getbuffer()))):
            if a != b:
                print(f'<{i}> [{a}]_[{b}]')

        assert bytes_ == bytes(file.getbuffer())  # fails
```

padraic-shafer:
@martintb @danielballan 
With my encouragement, this PR has grown more complicated than the original typo fix that was submitted.  I recommend that we retreat a bit to get the typos fixed quickly, and defer the netCDF resolution to a subsequent PR.  **Does this sound good?**

Let's keep:
* The bug fixes: mandatory parentheses around the `await` statements in `serialize_json()` and `serialize_netcdf()`.
* The working test for `serialize_json()`: `test_serialize_json()`.

Let's remove:
* Other changes to `serialize_netcdf()`.
* The test for `serialize_netcdf()`: `test_serialize_netcdf()`.
* --> I've started a [new issue to track these](https://github.com/bluesky/tiled/issues/568).

danielballan (approved):

padraic-shafer:
@martintb  Thank you for this contribution :)

martintb:
Thanks for your help and input @padraic-shafer!

Issue #515, opened 2023-07-13
Change `metadata` from property to method.


danielballan:
This has been on the table for awhile, and I think we have enough good reason to justify the change:

* It enables `metadata()` to be optionally async, as `read()` now is, if it needs to do blocking I/O.
* It leaves open a path to (optionally) accepting filter parameters for predicate push-down of something like JSONPath.

danielballan:
Discussed on a call with @tacaswell. Agreed that a path to predicate push-down is important.

padraic-shafer:
There seems to still be a mix of `metadata` sometimes as a "property" and sometimes as a "method".  This leads to e.g. being able to call either `read()` or `serialize_json()` on an xarray node(container) but not both.

I _think_ the update strategy from this PR simply need to be propagated elsewhere, but I haven't tested what might break if that happens.
* Remove `@property` from `def metadata(...`
* `X.metadata[...]` becomes `X.metadata()[...]`

See:
https://github.com/bluesky/tiled/blob/b3564569db8b80672a8e75583817e5b9f8e3a927/tiled/client/base.py#L157-L163
https://github.com/bluesky/tiled/blob/b3564569db8b80672a8e75583817e5b9f8e3a927/tiled/client/xarray.py#L99-L103

padraic-shafer:
> I think the update strategy from this PR simply need to be propagated elsewhere, but I haven't tested what might break if that happens.

[UPDATE]  I tried this out locally and all tests pass.  Let me know if I should submit a PR for this.

PR #561, opened 2023-08-29
Consistently access metadata as a method across the project

padraic-shafer:
See post-close discussion in #515.

I could not discern whether it was an intentional decision to keep `metadata` as a property in the python client -- perhaps to avoid breaking beamline scripts during the previous operations cycle?

One consequence of the current state is that calling `serialize_json()` for an xarray causes errors that cannot be resolved by simple changes in one module.  Some of the underlying calls expect `metadata` to be a method, while other calls require `metadata` to be a property.

I'm proposing here to consistently treat `metadata` as a method throughout--effectively taking #527 a step farther.

---
This suggestion arose from testing the bug fix in #560.  My recommendation is to apply this PR, followed immediately by #560.

### DIFF
--- a/docs/source/how-to/custom-clients.md
+++ b/docs/source/how-to/custom-clients.md
@@ -26,8 +26,8 @@ Suppose data labeled with the `xdi` spec is guaranteed to have a metadata
 dictionary containing the following two entries:
 
 ```py
-x.metadata["XDI"]["Element"]["Symbol"]
-x.metadata["XDI"]["Element"]["Edge"]
+x.metadata()["XDI"]["Element"]["Symbol"]
+x.metadata()["XDI"]["Element"]["Edge"]
 ```
 
 When the Tiled client encounters this type of data, we would like to hand
@@ -48,7 +48,7 @@ class XDIDatasetClient(tiled.client.dataframe.DataFrameClient):
         assert self.item["attributes"]["structure_family"] == "table"
 
     def __repr__(self):
-        md = self.metadata["XDI"]
+        md = self.metadata()["XDI"]
         return f'<{x.item["id"]} {md["Element"]["Symbol"]} {md["Element"]["Edge"]}>'
 ```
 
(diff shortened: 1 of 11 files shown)


danielballan:
I'll back up and explain the intentions.

On the _server_ side...

* We want the option to be async, which properties cannot be.
* We want the option, in the future, to inject arguments that filter particular metadata of interest, and do so at the lowest level possible ("predicate push-down") for performance.
* Brevity is not a concern; no one is interactively typing this code.

All of this motivated the change from property to method.

On the _client_ side...

* Brevity is a consideration. All else being equal, a property is nice and succinct, where it makes sense.
* The I/O has already happened; `.metadata` is accessing information we already have in memory, so there is no thought of async (no I/O...) or filtering (too late, we already downloaded it).
* We are at the point where we consider backward-incompatible API changes to the client with care, wanting to avoid breaking user code unless we have a strong reason.

So on the client `metadata` remained a property.

This does introduce a kind of inconsistency between the Adapter API and Client API that could give developers a moment of confusion. However, I look at these as very different APIs---even though they "rhyme" in places---that serve distinct roles in Tiled. In the future, I could even imagine splitting the server and client codebases into separate repos/packages. It has been convenient to co-develop them, and that may remain the case, but it's useful to think of them as distinct things.

Is this compelling?

padraic-shafer:
OK, I follow your reasoning.

It's likely that I have misunderstood the type of "node" that is expected [here in `as_dataset(node)`](https://github.com/bluesky/tiled/blob/b3564569db8b80672a8e75583817e5b9f8e3a927/tiled/serialization/xarray.py#L17).  In the [proposed test](https://github.com/martintb/tiled/pull/1) that I wrote, I passed it a client node(container); but maybe I was supposed a different type of object (ArrayAdaptor, DatasetAdapter, MapAdaptor, ...)?

https://github.com/bluesky/tiled/blob/b3564569db8b80672a8e75583817e5b9f8e3a927/tiled/serialization/xarray.py#L17

danielballan:
Oh I see. I will the PR more carefully. Yes, serialization takes data from the adapter and translates it into bytes to send to the client.

padraic-shafer:
> Yes, serialization takes data from the adapter and translates it into bytes to send to the client.

That's what I was missing.

In that case, I will:
1. Withdraw this PR
2. Rewrite the test correctly for #560 
3. Go back and clean up the conversations as needed.

Issue #434, opened 2023-04-18
Support a fallback structure family for "opaque bytes"


danielballan:
This is another idea that arose during the NIST visit.

Tiled's data model constrains everything to be one of its recognized structure families (array, dataframe, sparse, node) or JSON-encodable metadata sitting alongside one of those types. There will be cases where there is binary (not JSON-encodable) information that is relevant and that some clients programs will know what to do with.

Our line on this so far has been, "If you have a files, use a static file server or Globus or another file-based solution, and _link_ to that from the metadata in Tiled." And for cases where you have a lot of un-structured data (directory of PowerPoint documents or PDFs) I think think that's the right call. But John Henry at NIST articulated a compelling argument that it is useful to enable Tiled to carry binary data in-line when it's useful.

I think this would take the shape of a new structure family, perhaps `opaque_bytes`. Its structure would simply be a length, and it would be sliceable by byte range. Tiled would not be able to transcode it, only send it in its original representation as bytes. Any context necessary to interpret the bytes would have been either known _a priori_ to a client. The (JSON-encodable) `metadata` attached to the `opaque_bytes` node may provide helpful information in this regard for a client, but it would be "opaque" to Tiled itself.

We are all in agreement that if you have _mostly_ unstructured / opaque data, then Tiled is not adding value and you should just use a static file server. But if you have a _little_ unstructured / opaque data and you want to place it logically alongside structure data, there is an argument that Tiled should enable this.

danielballan:
One possible name for this proposed new structure family is `unstructured`.

danielballan:
We have discussed various bars that data can clear:

0. We technically have the bytes, but everything is totally unlabeled.
1. We have the bytes and some metadata.
2. We know how to open the file and interpret the bytes as numbers.
3. We have a schema that tells us in a "machine-actionable" way significance of the numbers.

Tiled currently insist that you start at (2). When @jmaruland and I first visited NIST, John Henry made the case that we should actually start at (1) --- that we should accept files that we cannot open. That leaves a lot of Tiled's capability on the table. Interfaces like these rely on Tiled being able to open the file and provide it in a known format:

https://tiled-demo.blueskyproject.io/ui/browse/generated/short_table
https://tiled-demo.blueskyproject.io/ui/browse/fxi/raw/1b0b4d73-6d87-43ab-8d62-ed035c51b9b4/primary/data/Andor_image

[...]

Issue #90, opened 2021-09-28
Add /assets routes for direct (unstructured) access to underlying data


danielballan:
* `GET /assets/manifest/{path}` --- List of underling resources and, if applicable and enabled by server config, their local paths on the local filesystem (which the client may or may not be able to directly access)
* `GET /assets/content/{something}` --- Direct download of an asset, spooled from disk

danielballan:
An update on current thinking:

The “manifest” of asset information is (optionally) included in the description of a node returned by `/metadata` and `search` routes. This seems better than a separate route, as originally proposed above, for latency.

The route for content can just be `/asset/{id}`.

danielballan:
Closed by #649

PR #570, opened 2023-09-05
Support for reading data as simple "bytes"

padraic-shafer:
## Background

_From Issue #434_: Tiled's data model constrains everything to be one of its recognized structure families (array, table, sparse, awkward, container, composite) or JSON-encodable metadata sitting alongside one of those types. There are cases where there is binary (not JSON-encodable) information that is relevant and that some client programs will know what to do with: PDFs, firmware blobs, proprietary binary formats, raw detector dumps, etc.

This PR introduces a new `bytes` structure family for cataloging such files without requiring them to be interpreted by Tiled.

## Design

The `bytes` family is **registration-only**: it exists so externally produced files with no useful logical structure can be cataloged alongside the other structure families, and downloaded as-is. Concretely:

- `BytesStructure` carries no fields of its own. The interpretation (MIME type) is recorded on the `DataSource`; the per-asset byte length is recorded on each `Asset` (new `Asset.size` field).
- `BytesAdapter` is minimal and read-only — it has no `read()` / `read_stream()` methods.
- There is **no `/bytes`-family content endpoint and no specialized client class**. Bytes nodes are returned as the generic `BaseClient`.
- Downloads go through the existing `/api/v1/asset/bytes/{path}?id=N` route — one HTTP request per underlying asset, gated by `settings.expose_raw_assets`. This is the same route used to retrieve raw files backing arrays and tables, so bytes piggybacks on infrastructure that already exists, supports HTTP `Range` requests, and is permission-aware.
- The Python entry point is `BaseClient.raw_export(...)`, which is the existing high-level "download the raw files behind this node" API. It works for any node; for bytes nodes it is the primary way to retrieve content. It now also accepts an in-memory `MutableMapping` (e.g. a `dict`) as destination, streaming each asset into a `BytesIO` buffer with no filesystem I/O.

This design deliberately avoids:
- A dedicated `/bytes` route or specialized client (would duplicate `/asset/bytes` and `raw_export`).
- Storing per-chunk size/offset on `BytesStructure` (those belong on `Asset.size`).
- HTTP Range semantics specific to bytes (already supported uniformly by `/asset/bytes`; see #521).

### Bytes family
- New `StructureFamily.bytes` value, `BytesStructure` (empty), `BytesAdapter` (minimal, read-only).
- Catalog and schema wiring; Alembic migration.
- Bytes children are allowed in `composite` containers.
- Bytes is the fallback adapter for any unknown-structure MIME type (e.g. mapping `application/pdf` to `BytesAdapter` is enough to register PDFs).

### `Asset.size`
- The existing `size` field surfaced on the `Asset` API model (the ORM column has existed since the catalog schema was created but was previously never written or read).
- `tiled register` populates it via `os.stat()` for non-directory `file://` assets walked from the local filesystem.
- New `tiled.utils.size_from_uri(data_uri)` helper dispatches to `os.stat()` for `file://` and `obstore.head()` for `s3://` / `az://` / `gs://`, so callers building `Asset` objects for object-store-backed files can populate `size` uniformly.
- Server-side `_put_asset` opportunistically refreshes `size` on re-registration (without clobbering a known value with `None`).

### `BaseClient.raw_export`
- Now accepts a directory `Path` **or** a `MutableMapping` for in-memory downloads into `BytesIO` buffers. Single-asset nodes produce one entry keyed by the server-supplied filename; multi-asset nodes produce `<asset_id>/<filename>` entries, mirroring the on-disk layout.
- Parameter renamed `destination_directory` → `destination`; the old name is accepted with a `DeprecationWarning`.
- Tolerates responses with no `Content-Length` header (e.g. streamed compressed bodies — fixes a real bug surfaced by the project's zstd middleware).

### Tests
- Integration tests covering register/download round-trip (single- and multi-asset), `expose_raw_assets` gating, arbitrary MIME types, `Asset.size` upsert, the deprecation warning, and the missing-`Content-Length` regression.
- `raw_export` tests parametrized over disk and in-memory destinations.
- New `size_from_uri` tests.
- Bytes-in-composite tests in `test_composite.py`.

## References

- Issue #434 (original request).
- Related discussions: #473, #90, #521.
- _From Issue #434_, _@danielballan_, _@jmaruland_, _@padraic-shafer_ (original PR author).


padraic-shafer:
In terms of where the code needs to be updated, there are many analogs between this current PR and #549.  For convenience [here are the diffs](https://github.com/bluesky/tiled/pull/549/files).

tiled/structures/bytes.py
```diff
@@ -0,0 +1,56 @@
...
+    underlying asset. The invariant `size == sum(chunks)` is enforced.
+
+    The interpretation of the payload (MIME type, filename, encoding)
+    is delegated to the `DataSource` / `Asset`. Instead, `BytesStructure`
+    intentionally describes _only_ the layout that the server needs to
+    resolve a byte-range slice into a set of per-chunk reads.
+    """
+
+    size: int  # total logical byte length
+    chunks: Tuple[int, ...]  # per-chunk byte lengths; sum(chunks) == size
```
danielballan: Since the content is opaque, I'm not sure that server-provided chunks add value, and I worry that this JSON could become large.

tiled/server/router.py
```diff
@@ -1766,6 +1810,100 @@ async def awkward_full(
         except UnsupportedMediaTypes as err:
             raise HTTPException(status_code=HTTP_406_NOT_ACCEPTABLE, detail=err.args[0])
 
+    @router.get("/bytes/full/{path:path}", name="Full Bytes")
```
danielballan: I wonder if we use the existing asset endpoints and avoid introducing a new endpoint.

tiled/structures/bytes.py
genematx: yeah, I think, I agree; it probably doesn't do much. The idea was to record the size of each constituent file (i.e. chunk), but the problem is those files themselves can be huge, to download a "chunk", it still would need to be split into smaller pieces. Having these chunks in the structure would help the server to split the response, but it probably is not very critical.

genematx: I don't think JSON would be any larger than what we have in some cases with datasets that contain thousands of individual tiff files.

tiled/server/router.py
genematx: I thought about that too! yes, I will have a look if it's possible without any major changes.

genematx: fixed. BytesStructure nodes fully rely on `/asset/bytes` endpoints now.

tiled/structures/bytes.py
genematx: removed any attributes from `BytesStructure`. The sizes of individual assets can be determined from their own `size` attributes (which are populated in the DB now).

docs/source/explanations/structures.md
```diff
@@ -11,6 +11,8 @@ The structure families are:
 
 * array --- a strided array, like a [numpy](https://numpy.org) array
 * awkward --- nested, variable-sized data (as implemented by [AwkwardArray](https://awkward-array.org/))
+* bytes --- an opaque sequence of bytes (a fallback for externally registered
```
danielballan: I wonder whether this should be placed last because it is special---especially in the context of documentation on "structures" because it has none.

docs/source/explanations/structures.md
```diff
@@ -276,6 +278,101 @@ $ http :8000/api/v1/metadata/nested/ragged_array | jq .data.attributes.structure
...
+}
+```
+
+Because Tiled gains no leverage from treating this as opaque, `bytes` is
+intentionally **read-only**. Use it for external files (PDFs, firmware blobs,
+proprietary binary formats) that lack a useful logical structure. To *store*
+new data in Tiled, prefer a structured family (`array`, `table`, `awkward`, …)
+so the server can chunk, slice, and serialize intelligently.
+
+#### Auto-registration with `tiled register`
```
danielballan: This page is focused on explaining each structure. Everything from here reads like "how-to guide" content to me, maybe better on a new document (which could be cross-referenced from here).

docs/source/getting-started/10-minutes-to-tiled.md
```diff
@@ -367,6 +368,22 @@ this efficiently by launching parallel downloads.
 c['examples/xraydb/C/edges'].raw_export('downloads/')
 ```
 
+This works for any node with assets (arrays backed by HDF5, tables backed by
```
danielballan: The goal of this document is to present a flavor of what is possible. This introduces a couple concepts that may be unfamiliar to a casual programmer: `MutableMapping`, `BytesIO`, "asset". Consider explaining less or cross-referencing to a how-to guide for this kind of depth.

tiled/utils.py
```diff
@@ -793,6 +793,33 @@ def ensure_uri(uri_or_path) -> str:
...
+    Dispatches on URI scheme: `file://` URIs are stat'd on the local
+    filesystem; object-store URIs (those in `SUPPORTED_OBJECT_URI_SCHEMES`)
+    issue a HEAD via obstore. Raises `ValueError` for any other scheme.
+    Underlying I/O errors (`FileNotFoundError`, network errors, etc.)
+    propagate; callers that want best-effort behavior should catch them at
+    the call site.
+    """
+    # Deferred import: `tiled.storage` imports from this module, so importing
+    # it at module top would create a cycle.
+    from .storage import SUPPORTED_OBJECT_URI_SCHEMES, ObjectStorage, get_storage
```
danielballan: Should we keep this utility in `tiled.storage`? Too late, I read about a growing consensus to stop making catch-up `utils` modules. When in doubt, avoiding _growing_ the ones we have might be a good direction.

tiled/client/container.py
```diff
@@ -1415,6 +1416,7 @@ def _write_partition(x, partition_info, client):
             "array": _LazyLoad(("..array", Container.__module__), "DaskArrayClient"),
             # TODO Create DaskAwkwardClient
             # "awkward": _LazyLoad(("..awkward", Container.__module__), "DaskAwkwardClient"),
+            "bytes": _Wrap(BaseClient),
```
danielballan: What do you think of defining

```py
BytesClient = BaseClient
```

I see some small advantages:
1. It has a clearer name.
2. It might be confusing/surprising to instantiate a `Base...` class on its own.
3. If, in the future, `bytes` calls for special convenience methods, we'll have a place to put them without breaking back-compat.

For me number (3)---i.e., future-proofing---is the strongest argument.

danielballan (commented):
Awesome! I left some comments, mostly about docs.

tiled/client/container.py
genematx: Agree! Seeing `BaseClient` there bothered me too. Fixed.

tiled/utils.py
genematx: Moved.

docs/source/getting-started/10-minutes-to-tiled.md
genematx: Yes, agree. Moved these details to a separate doc.

docs/source/explanations/structures.md
genematx: Reordered.

docs/source/explanations/structures.md
genematx: Moved to how-to's.

danielballan (approved):

PR #572, opened 2023-10-02
Improve test coverage on compression, and fix bugs

danielballan:
In particular, `gzip` is intended to be supported for certain data formats (e.g. plain text) but it was not due to a small bug.

### DIFF
--- a/tiled/_tests/test_compression.py
+++ b/tiled/_tests/test_compression.py
@@ -1,0 +1,50 @@
+import numpy
+import pytest
+from starlette.testclient import TestClient
+
+from ..adapters.array import ArrayAdapter
+from ..adapters.mapping import MapAdapter
+from ..server.app import build_app
+
+
+@pytest.fixture
+def app():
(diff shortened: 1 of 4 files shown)


tiled/_tests/test_compression.py
```diff
@@ -0,0 +1,50 @@
...
+    tree = MapAdapter(
+        {
+            "compresses_well": ArrayAdapter.from_array(
+                numpy.zeros((1000, 1000)), metadata=metadata
+            )
+        },
+    )
+    return build_app(tree, authentication={"single_user_api_key": "secret"})
+    # In this module we use a raw TestClient instead of tiled.client to omit
+    # tiled.client's default headers and other configuration.
```
padraic-shafer, line 22: I don't understand which headers/config are being bypassed, but I accept that you've thought about this.


I'd be interested in hearing about this later.

padraic-shafer (approved):
This looks good.  Thanks for keep the size of the PR digestible. :)

I've included one optional formatting suggestion.  I'll be resolving the rest, which are simply commentary.

PR #573, opened 2023-10-02
Satisfy expected API of decoders.

danielballan:
httpx expects `Decoder.decode(data)` to always return `bytes`, not `None`.


padraic-shafer (approved):
This makes perfect sense.

PR #582, opened 2023-10-17
Fix `project_form` for projections within a RecordArray field and for node types without indexes.

### LINKED ISSUE #578: Can't slice (with predicate push-down) a field within a field of an Awkward Array
We get `KeyError: 'nodeX-offsets'` issues (for different `X`, depending on the attempted slice).

I should make a suite of examples to test and go through them with a persistent database that I can reboot, make changes to the code, and try again, to narrow down on this error. Maybe that can become a unit test.

I don't know why, but it's only showing me the part of the stack trace in which Awkward fails to find the key in the `container`. It's not showing me the Tiled server code that calls Awkward, where I want to look and see which keys it _does_ have.

The types of slices that are failing:

```python
client["dataset_name"]["file", "filename"]
```

where `"filename"` is a field of a record that is inside the `"file"` field of an outer record.

jpivarski:
Issue #578 was because of a typo in the code that I sent you. The second and third fixes are more fundamental/conceptual.

Fixes #578.

### DIFF
--- a/tiled/_tests/test_awkward.py
+++ b/tiled/_tests/test_awkward.py
@@ -144,3 +144,21 @@ def test_large_number_of_form_keys(client):
     assert form_key_length > URL_LENGTH_LIMIT
     url_length = len(str(request.url))
     assert url_length <= URL_LENGTH_LIMIT
+
+
+def test_more_slicing_1(client):
+    array = awkward.Array([
+        {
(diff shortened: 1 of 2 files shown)


danielballan (approved):
Looks sensible. I played with these test cases interactively and did not hit any problems.

PR #586, opened 2023-10-23
Fix indexing on results after applying a search query

### LINKED ISSUE #585: Bug in keying into search results on a Catalog
To reproduce, start a server:

```
tiled serve catalog --temp --public
```

and try to key into a search result. (There doesn't need to be any data in this to reproduce the error.)

```

c = from_uri("http://127.0.0.1:8000")
from tiled.queries import Key
c.search(Key("test") == "whatever")["stuff"]
```

This raises on the client side:

```
HTTPStatusError: Server error '500 Internal Server Error' for url 'http://127.0.0.1:8000/api/v1/search/?filter%5Blookup%5D%5Bcondition%5D%5Bkey%5D=stuff&filter%5Beq%5D%5Bcondition%5D%5Bkey%5D=test&filter%5Beq%5D%5Bcondition%5D%5Bvalue%5D=%22whatever%22&sort='
For more information check: https://httpstatuses.com/500
```

Server logs show:

```
  File "/home/dallan/Repos/bnl/tiled/tiled/server/core.py", line 151, in apply_search
    tree = MapAdapter({key_lookup: tree[key_lookup]}, must_revalidate=False)
                                   ~~~~^^^^^^^^^^^^
TypeError: 'CatalogContainerAdapter' object is not subscriptable
```

which is this line here:

https://github.com/bluesky/tiled/blob/7c122f9307b52c24cfba4c3b7e23d2ebe720a694/tiled/server/core.py#L150

I believe the fix is to make the `apply_search` function async, and then to use `await tree.lookup_adapter(key_lookup)` if a `lookup_adapter` method exists. Similar to this pattern:

https://github.com/bluesky/tiled/blob/7c122f9307b52c24cfba4c3b7e23d2ebe720a694/tiled/server/dependencies.py#L84-L94

jmaruland:
This PR fixes issue #585 where a bug was found that causes a TypeError while trying to access the results after performing a search query.
This happens because indexing of a search result (c.search(Key(test)==value)[index]) reaches the node through a different process that the basic indexing (c[index])

Closes #585


jmaruland:
Next step: adding a test case in test_search.py to cover this case.

danielballan:
Thanks for the clear PR title and description. For extra credit, if you edit the original post the add the text "Closes #585" it will automatically close that issue when merged. To work, it has to be in the top post, not a comment.

About the test: it currently defined a pytest fixture which is (re)used by every test in the `test_search.py` module:

https://github.com/bluesky/tiled/blob/7c122f9307b52c24cfba4c3b7e23d2ebe720a694/tiled/_tests/test_search.py#L25-L30

If we turn that fixture into a [parameterized fixture](https://docs.pytest.org/en/7.3.x/how-to/fixtures.html#fixture-parametrize) we can automatically get each test to run twice, once against a `MapAdapter` and once against a `Catalog`. Here is a sketch, leaving one aspect as an exercise for the reader. :-) 

```py
@pytest.fixture(scope="module", params=["map", "catalog"]) 
def client(tmpdir): 
    if request.param == "map":
        app = build_app(tree) 
        with Context.from_app(app) as context: 
            yield client
    elif request.param == "catalog":
        catalog = in_memory(writable_storage=str(tmpdir))
        app = build_app(catalog) 
        with Context.from_app(app) as context: 
            # Upload arrays and metadata from tree into client so that it has matching content.
            ...
            yield client
    else:
        assert False 
```

padraic-shafer:
Pytest has a feature called `getfixturevalue` that helps with this pattern.  Then you can use a specialized fixtures as a parameter to a more general fixture. 

It has the benefit of naming the specialization (which can be used independently) and avoids conditional/branch logic.  It's easy to add or remove the params as needed.  Also no need for an "else" branch because the test will fail with "no fixture named ..." or something like that.

```python3
@pytest.fixture
def map_client():
    app = build_app(tree) 
    with Context.from_app(app) as context: 
        yield client


@pytest.fixture
def catalog_client(tmpdir):
    catalog = in_memory(writable_storage=str(tmpdir))
    app = build_app(catalog) 
    with Context.from_app(app) as context: 
        # Upload arrays and metadata from tree into client so that it has matching content.
        ...
        yield client


@pytest.fixture(
    params=("map_client", "catalog_client"),
)
def client(request: pytest.FixtureRequest):
    yield request.getfixturevalue(request.param)
```

danielballan (approved):

PR #587, opened 2023-10-24
Support POST for /node/full/

### LINKED ISSUE #579: Add `POST` option for long `GET` URL queries
In #577, we added `POST /awkward/buffers/{path}` as an alternative to `GET /awkward/buffers/{path}` that supports large number of buffers to be requested, placing the names of the buffers in the request body, as a JSON array, rather than in the URL query parameters.

The benefit of `POST` is that it scales; the benefit of `GET` is that HTTP response caches---be they web browser caches, the cache in Tiled's own Python client, or server-side caches---generally work on `GET` but not on `POST`.

We previously encountered the same problem with fetching wide xarray Datasets: we're selecting all the columns of scalar data, distinct from those with N-dimensional data, and it can be a large list. There, we took a different approach. Instead of adding a `POST` route, we split the `GET` request into multiple batches, with each `GET` fetching a subset of the keys.

https://github.com/bluesky/tiled/blob/1536fc3df93015c593270583cbb917afa3200fad/tiled/client/xarray.py#L144-L167

In the wild, we don't see _large_ number of batches needed...just a couple more than 1, sometimes.

I wonder if `POST` would be better here. The existing solution feels a bit fragile (hard-coded URL limit) and complex. For many use cases, Tiled's performance is latency-bound, and reducing the _number_ of requests has often been shown to be the best way to improve performance. Going from N `GET`s to one `POST` may be a win in this way.

We could potentially be benefiting from parallelism from the batched `GET`s, but as you can see we are not using parallelism here, just a simple loop. And we could do the same with batched `POST`, for that matter.

I cannot recall why we didn't go for `POST` from the start. I know I was aware of the pattern of using `POST` as an alt for `GET` when the requests parameters are large. Maybe I was highly weighting keeping HTTP response caching, because at the time we relied on it for Tiled's since-removed "offline mode".  As we now have a [more robust notion of how to achieve offline access](https://github.com/bluesky/tiled/issues/473), and it does not involve relying directly on an HTTP response cache, I think we should revisit this.

AbbyGi:
Closes #579.

### DIFF
--- a/tiled/client/container.py
+++ b/tiled/client/container.py
@@ -248,10 +248,10 @@ class Container(BaseClient, collections.abc.Mapping, IndexersMixin):
             key, *tail = keys
             tail = tuple(tail)  # list -> tuple
             content = handle_error(
-                self.context.http_client.get(
+                self.context.http_client.post(
                     self.item["links"]["search"],
                     headers={"Accept": MSGPACK_MIME_TYPE},
-                    params={
+                    json={
                         **_queries_to_params(KeyLookup(key)),
                         **self._queries_as_params,
                         **self._sorting_params,
--- a/tiled/server/router.py
+++ b/tiled/server/router.py
@@ -4,7 +4,7 @@ from datetime import datetime, timedelta
 from functools import partial
 from typing import Any, List, Optional
 
-from fastapi import APIRouter, Depends, HTTPException, Query, Request, Security
+from fastapi import APIRouter, Body, Depends, HTTPException, Query, Request, Security
 from jmespath.exceptions import JMESPathError
 from pydantic import BaseSettings
 
@@ -545,7 +545,7 @@ async def table_partition(
     response_model=schemas.Response,
     name="full 'container' or 'table'",
 )
-async def node_full(
+async def get_node_full(
     request: Request,
     entry=SecureEntry(scopes=["read:data"]),
     principal: str = Depends(get_current_principal),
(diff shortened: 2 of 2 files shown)


AbbyGi:
This isn't working yet. Running the `test_search.py` tests locally results in the following error:
`tiled.client.utils.ClientError: 405: Method Not Allowed http://local-tiled-app/api/v1/search/`

padraic-shafer:
Here is a hint...

The failing test is receiving a response that indicates only the GET method is allowed for this route.

```
response.__dict__ = {'status_code': 405, 'headers': Headers({'allow': 'GET', 'content-length': '31', 'content-type': 'application/json', ..., '_request': <Request('POST', 'http://local-tiled-app/api/v1/search/')>, ...
```

---
It seems that either this section (below) should revert to a GET request...or the `/search/` route (and presumably others) need to have a corresponding POST route.

https://github.com/AbbyGi/tiled/blob/fee5ee77a788eaa346abeda6a86fd6c43d73a3ec/tiled/client/container.py#L251-L258

I would tend to suggest reverting to the GET version in the client; then instead write a test that verifies the new POST route in the server/router for `/node/full/{path}`.

Subsequently we might still need to add POST routes corresponding to all the GET data routes; similar to `/awkward/buffers/{path}` and illustrated by `/array/*/{path}` in #579.

padraic-shafer:
With fresh eyes on this, I think it would make more sense to start from the bottom up. I believe the general `/node/full/{path}` POST will fail if the POST `/full/{path}` is not already implemented for the other structure families and media types.

E.g., Add a test to `test_xarray.py` that verifies a new (yet to be added) POST call. Then create the POST route that satisfies the test (without breaking other tests). Repeat this for `test_array.py`, `test_dataframe.py`, `test_dataframe.py`, `test_hdf5.py`, `test_sparse.py`, `test_structured_array.py`, `test_tiff.py`, and `test_container_files.py`. By that point, a POST route for `/node/full/{path}`has probably been created. If not, a container-specific test might be warranted.

I welcome input from @danielballan and others on the approach here.

danielballan:
I wonder if we should merge this first, which deprecates `/node/full` and may make things more clear. https://github.com/bluesky/tiled/pull/612

padraic-shafer:
@AbbyGi With the recent helpful updates from #612, I think we should abandon this PR and start anew.  I'll move some of this discussion above to issue #579.

danielballan:
Superseded by #657

PR #592, opened 2023-10-26
add dummy file name for some tiffseq

dylanmcreynolds:
This addresses a bug with the tiff sequence code.

When the files of a tiffsequence don't start with letters, just numbers, a hash was used for the key for that node. 

So:
```
tiff_files
  - 001.tiff
  - 002.tiff
  - a001.tiff
  - a002.tiff
  - b001.tiff
  - b002.tiff
```
would result in 

```
['d3a7b120-91f8-41ef-a1ff-f3e312f23525',
 'a',
 'b']
```

The unit test wasn't expecting a random name for this node, so it didn't test for it.

This PR detects that case and adds an `_` in as the key name. On a call with @taxe10 and @danielballan, we debated this default root name for only a couple seconds. I'm happy to hear better ideas.

### DIFF
--- a/tiled/_tests/test_tiff.py
+++ b/tiled/_tests/test_tiff.py
@@ -7,7 +7,7 @@ import tifffile as tf
 from ..adapters.mapping import MapAdapter
 from ..adapters.tiff import TiffAdapter, TiffSequenceAdapter
 from ..catalog import in_memory
-from ..catalog.register import register
+from ..catalog.register import register, TIFF_SEQUENCE_EMPTY_NAME_ROOT
 from ..client import Context, from_context
 from ..server.app import build_app
 
@@ -129,7 +129,7 @@ a,b,c
         # Single image is its own node.
         assert client["single_image"].shape == (3, 5)
         # Each sequence is grouped into a node.
-        assert client["_"].shape == (10, 3, 5)
+        assert client[TIFF_SEQUENCE_EMPTY_NAME_ROOT].shape == (10, 3, 5)
         assert client["image"].shape == (10, 3, 5)
         assert client["other_image"].shape == (10, 3, 5)
         assert client["other_image2_"].shape == (10, 3, 5)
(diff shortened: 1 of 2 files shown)


padraic-shafer:
This seems ok.  

Other thoughts, just to throw them out there: `''`, `' '`, or `None`.

```python 3
>>> d1 = {'': 0, 'a':1, 'b': 2}
>>> d1['']
0

>>> d2 = {' ': 0, 'a':1, 'b': 2}
>>> d2[' ']
0

>>> d3 = {None: 0, 'a':1, 'b': 2}
>>> d3[None]
0

# But not...
>>> d3[]
  File "<stdin>", line 1
    d3[]
       ^
SyntaxError: invalid syntax
```

danielballan:
Since the keys also end up in path parameters, this unfortunately rules out a couple of those options:

* `''` ends up as `/api/v1/metadata/tif//` (emphasis on `//`) which HTTP elides into `/`.
* `None` has to to coerced to a string of some kind, so its special status as an out-of-band sentinel does not survive.
* Space technically works but may invite confusion.

I think we'll be happy if we stick with "The key must be a non-empty string."

danielballan:
Feedback from @Wiebke, who thought about related issues in her work on #555, would be very welcome as well.

dylanmcreynolds:
I almost feel like something super explicit might be better, even if a bit verbose:

```
['_unnamed_sequence',
 'a',
 'b']

```

padraic-shafer:
Explicit and _brief_ would be even better: "_default", "_unnamed", "_blank", "_na", etc.

I favor including a non-alphanumeric to try to minimize name clashes with the user-defined sequences. Something that does not require url-encoding would be good. I suppose a leading "_" will list this key at the beginning for most sorting routines?

danielballan:
:+1: for brevity. Any of these seem good to me:

> "_default", "_unnamed", "_blank", "_na", etc.

Using a number would get weird in this context:

```py
c['tif']['0'][0]  # first image
```

dylanmcreynolds:
I committed `_unnamed`. I like it because it's pretty explicit.

danielballan (approved):
Looks good to me, but we should wait for @Wiebke to weigh in before we merge.

Wiebke (approved):
I think this is a sensible solution for unnamed sequences and I liked `_unnamed` as well.

Issue #518, opened 2023-07-13
Cover CLI with unit tests


danielballan:
The recent breakage fixed by #512 reminds me that unit tests for the CLI are overdue.

https://typer.tiangolo.com/tutorial/testing/

PR #596, opened 2023-10-31
Add `--api-key` CLI option

### LINKED ISSUE #583: Support single-user API key configuration via CLI arg?
In a development or demo context, it is convenient to start a writable server with a stable single-user API key, such that restarting the server does not require copying a new key over into the client. As such, I type this a lot:

```
TILED_SINGLE_USER_API_KEY=secret tiled catalog serve --temp
```

It might be nicer to type something like

```
tiled catalog serve --temp --api-key=secret
```

## Considerations on whether we should add this at all

I intentionally avoided doing this at first because I worry that it makes it too easy to run a server with an insecure (guessable) API key. But, the `TILED_SINGLE_USER_API_KEY=secret ...` pattern is now all over the docs, so it's not exactly hard to discover, and I think there may not be much value in making it inconvenient to set.

## Considerations on what to name it.

Maybe it should be `--token` instead of `--api-key`? I think `token`:

* May be more memorable to non-devs
* Is easy to _say_ (as opposed to "A P I dash...no, dash, not underscore...KEY") which comes up in a my life...
* Matches what Jupyter uses, for what it's worth

In fact I wonder if everywhere we have `api_key`...

* `$ tiled api_key ...`
* `https://...?api_key=...`

we should instead name it `token`. I chose `api_key` early on because at the time I wasn't sure if "token" had a certain precise technical meaning. Now, years into this, it seems to me that "token" is used broadly to cover both API keys (e.g. GitHub "Personal Access Token") and the more transient bearer tokens used in OAuth2.

Early on, I also wondered if we'd run into confusing name collisions between API key-style tokens and OAuth2 tokens. But, as the latter are managed internally and transparently to the user, this potential name collision does not really end up being user-facing.

Thoughts?

danielballan:
Closes #583 

e.g.

```
```

Since there is only ever one key, I think a flag plus the existing env var (`TILED_SINGLE_USER_API_KEY`) is good enough for now. We can add `--api-key-file` later if we like.

### DIFF
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -409,8 +409,8 @@ or via the environment variable TILED_SINGLE_USER_API_KEY.""",
         ]:
             if authentication.get(item) is not None:
                 setattr(settings, item, authentication[item])
-            if authentication.get("single_user_api_key") is not None:
-                settings.single_user_api_key_generated = False
+        if authentication.get("single_user_api_key") is not None:
+            settings.single_user_api_key_generated = False
         for item in [
             "allow_origins",
             "response_bytesize_limit",
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -409,6 +409,8 @@ or via the environment variable TILED_SINGLE_USER_API_KEY.""",
         ]:
             if authentication.get(item) is not None:
                 setattr(settings, item, authentication[item])
+            if authentication.get("single_user_api_key") is not None:
+                settings.single_user_api_key_generated = False
         for item in [
             "allow_origins",
             "response_bytesize_limit",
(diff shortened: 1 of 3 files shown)


danielballan:
FYI, unit testing these CLIs is a known issue, #518.

tiled/server/app.py
```diff
@@ -409,6 +409,8 @@ def override_get_settings():
         ]:
             if authentication.get(item) is not None:
                 setattr(settings, item, authentication[item])
+            if authentication.get("single_user_api_key") is not None:
+                settings.single_user_api_key_generated = False
```
padraic-shafer, line 413: Move this out of the for-loop above?

```suggestion
        if authentication.get("single_user_api_key") is not None:
            settings.single_user_api_key_generated = False
```

padraic-shafer (changes requested):
Spotted one settings update that gets called multiple times in a for-loop.  I think it should get moved out of that loop.

The rest looks good.

padraic-shafer (approved):
This is a helpful addition to the Tiled CLI

PR #607, opened 2023-11-16
Revoke session on logout.

### LINKED ISSUE #602: Logout should revoke user's session?
The `/logout` route currently clears the cookie that stores the API key.  It seems like this call should also [revoke the current session](https://github.com/bluesky/tiled/blob/4ef171605f28bc3503db9f8ed39cf7b889e68379/tiled/server/authentication.py#L897-L904) to prevent further use of the associated tokens.  Does that sound reasonable?

https://github.com/bluesky/tiled/blob/4ef171605f28bc3503db9f8ed39cf7b889e68379/tiled/server/authentication.py#L1098-L1106

danielballan:
Closes #602 

In `main`, logging out with the Python client discards the relevant tokens, but it leaves the session open. Anyone holding a copy of the refresh token (somehow) can refresh the session and continue to use it, as long as they do so within the refresh token lifetime (days or weeks).

Now, logging out first _revokes_ the session, such that it can not be refreshed again, and then discards the tokens. A new unit test verifies that the old refresh token cannot be reused after logout.

### DIFF
--- a/tiled/_tests/test_authentication.py
+++ b/tiled/_tests/test_authentication.py
@@ -119,7 +119,7 @@ def test_logout(enter_password, config, tmpdir):
         # not been that long.
         client = from_context(context, username="alice")
         # The refresh token refers to a revoked session, so refreshing the
-        # session to new a *new* access and refresh token will fail.
+        # session to generate a *new* access and refresh token will fail.
         with pytest.raises(CannotRefreshAuthentication):
             client.context.force_auth_refresh()
 
--- a/tiled/_tests/test_authentication.py
+++ b/tiled/_tests/test_authentication.py
@@ -92,7 +92,7 @@ def test_password_auth(enter_password, config):
 
 def test_logout(enter_password, config, tmpdir):
     """
-    A password that is wrong, empty, or belonging to a different user fails.
+    Logging out revokes the session, such that it cannot be refreshed.
     """
     with Context.from_app(build_app_from_config(config)) as context:
         # Log in as Alice.
--- a/tiled/_tests/test_authentication.py
+++ b/tiled/_tests/test_authentication.py
@@ -100,18 +100,26 @@ def test_logout(enter_password, config, tmpdir):
             from_context(context, username="alice")
         # Reuse token from cache.
         client = from_context(context, username="alice")
-        # This is set by an autouse fixture in conftest.py.
+        # This was set to a unique (temporary) dir by an autouse fixture in conftest.py.
         tiled_cache_dir = os.environ["TILED_CACHE_DIR"]
+        # Make a backup copy of the cache directory, which contains the auth tokens.
         shutil.copytree(tiled_cache_dir, tmpdir / "backup")
-        # The removes the takes from the cache AND invalidates the session.
+        # Logout does two things:
+        # 1. Revoke the session, so that it cannot be refreshed.
+        # 2. Clear the tokens related to this session from in-memory state
+        #    and on-disk state.
         client.logout()
-        # Put the tokens back.
+        # Restore the tokens from backup.
+        # Our aim is to test, below, that even if you have the tokens they
+        # can't be used anymore.
         shutil.rmtree(tiled_cache_dir)
         shutil.copytree(tmpdir / "backup", tiled_cache_dir)
-        # Implementation detail: The access token JWT is still valid, so this
-        # works, for a limited time.
+        # There is no way to revoke a JWT access token. It expires after a
+        # short time window (minutes) but it will still work here, as it has
+        # not been that long.
         client = from_context(context, username="alice")
-        # The refresh token refers to a revoked session, so this fails.
+        # The refresh token refers to a revoked session, so refreshing the
+        # session to new a *new* access and refresh token will fail.
         with pytest.raises(CannotRefreshAuthentication):
             client.context.force_auth_refresh()
 
--- a/tiled/_tests/test_authentication.py
+++ b/tiled/_tests/test_authentication.py
@@ -1,4 +1,6 @@
 import io
+import os
+import shutil
 import subprocess
 import sys
 import time
@@ -86,6 +88,32 @@ def test_password_auth(enter_password, config):
         with fail_with_status_code(422):
             with enter_password(""):
                 from_context(context, username="alice")
+
+
+def test_logout(enter_password, config, tmpdir):
+    """
+    A password that is wrong, empty, or belonging to a different user fails.
+    """
+    with Context.from_app(build_app_from_config(config)) as context:
+        # Log in as Alice.
+        with enter_password("secret1"):
+            from_context(context, username="alice")
+        # Reuse token from cache.
+        client = from_context(context, username="alice")
+        # This is set by an autouse fixture in conftest.py.
+        tiled_cache_dir = os.environ["TILED_CACHE_DIR"]
+        shutil.copytree(tiled_cache_dir, tmpdir / "backup")
+        # The removes the takes from the cache AND invalidates the session.
+        client.logout()
+        # Put the tokens back.
+        shutil.rmtree(tiled_cache_dir)
+        shutil.copytree(tmpdir / "backup", tiled_cache_dir)
+        # Implementation detail: The access token JWT is still valid, so this
+        # works, for a limited time.
+        client = from_context(context, username="alice")
+        # The refresh token refers to a revoked session, so this fails.
+        with pytest.raises(CannotRefreshAuthentication):
+            client.context.force_auth_refresh()
 
 
 def test_key_rotation(enter_password, config):
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -669,7 +669,13 @@ and enter the code:
             return
 
         # Revoke the current session.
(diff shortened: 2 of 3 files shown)


tiled/_tests/test_authentication.py
```diff
@@ -88,6 +90,40 @@ def test_password_auth(enter_password, config):
...
+        shutil.rmtree(tiled_cache_dir)
+        shutil.copytree(tmpdir / "backup", tiled_cache_dir)
+        # There is no way to revoke a JWT access token. It expires after a
+        # short time window (minutes) but it will still work here, as it has
+        # not been that long.
+        client = from_context(context, username="alice")
+        # The refresh token refers to a revoked session, so refreshing the
+        # session to new a *new* access and refresh token will fail.
+        with pytest.raises(CannotRefreshAuthentication):
+            client.context.force_auth_refresh()
```
danielballan, line 124: This test was written first, and before the changes, this line was able to refresh the session _without_ raising an error. This reproduced #602 and demonstrated that the PR fixed it.

tiled/server/authentication.py
```diff
@@ -894,11 +894,31 @@ async def refresh_session(
...
+    settings: BaseSettings = Depends(get_settings),
+    db=Depends(get_database_session),
+):
+    "Mark a Session as revoked so it cannot be refreshed again."
+    request.state.endpoint = "auth"
+    payload = decode_token(refresh_token.refresh_token, settings.secret_keys)
+    session_id = payload["sid"]
+    # Find this session in the database.
+    session = await lookup_valid_session(db, session_id)
+    if session is None:
```
dylanmcreynolds, line 910: This is a little picky, but...404 is probably not the right error code here. You're posting to something without specifying an ID...a client might interpret the 404 as saything that the path `/session/revoke` is not found. How about [409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409)?

dylanmcreynolds (changes requested):
Just one little picky bit...

tiled/_tests/test_authentication.py
```diff
@@ -88,6 +90,40 @@ def test_password_auth(enter_password, config):
...
+        # Our aim is to test, below, that even if you have the tokens they
+        # can't be used anymore.
+        shutil.rmtree(tiled_cache_dir)
+        shutil.copytree(tmpdir / "backup", tiled_cache_dir)
+        # There is no way to revoke a JWT access token. It expires after a
+        # short time window (minutes) but it will still work here, as it has
+        # not been that long.
+        client = from_context(context, username="alice")
+        # The refresh token refers to a revoked session, so refreshing the
+        # session to new a *new* access and refresh token will fail.
```
padraic-shafer, line 122:
```suggestion
        # session to generate a *new* access and refresh token will fail.
```

???

tiled/client/context.py
```diff
@@ -669,7 +669,19 @@ def logout(self):
             return
 
         # Revoke the current session.
-        handle_error(self.http_client.post(f"{self.api_uri}auth/logout"))
+        refresh_token = self.http_client.auth.sync_get_token("refresh_token")
+        csrf_token = self.http_client.cookies["tiled_csrf"]
+        # Manually build a request, circumventing auth because this request is not authenticated.
+        # The refresh_token is the body is the relevant proof, not the Authentication header.
```
padraic-shafer, line 675:
```suggestion
        # The refresh_token in the body is the relevant proof, not the Authentication header.
```

tiled/server/authentication.py
```diff
@@ -894,8 +894,29 @@ async def refresh_session(
...
+        raise HTTPException(409, detail=f"No session {session_id}")
+    session.revoked = True
+    db.add(session)
+    await db.commit()
+    return Response(status_code=204)
```
padraic-shafer, line 915: Does creating an explicit `Response` object here imply that no cookie is being sent?  In the '/logout' route the cookie was explicitly deleted (expired) -- presumably because there it "intercepts" the response as a function parameter?

In any case, I think we want the browser to clear this cookie -- even though it has already been revoked by the server.

padraic-shafer:
Would it make sense to return 409 for this case also?  Session might exist, but it belongs to a different user.

https://github.com/bluesky/tiled/blob/cfb4cf3a900637be56460004d0dc13bb701a25bd/tiled/server/authentication.py#L931-L936

danielballan:
I think 404 is right in this case. The user is providing an ID, and this follows the principle, "If this isn't yours, you're now allowed to know whether or not it exists."

tiled/server/authentication.py
danielballan: Authentication can happen in two ways:
1. API key (shared secret), backed by an `APIKey` in the database, passed in a header, cookie, or query parameter
2. OAuth2 with JWT access and refresh tokens, backed by a `Session` in the database, passed in a header only

Clients use one or the other, not both. This new route is addressing (2), and no cookies are involved.

The now-deprecated `logout` route was targeting (1), but there aren't any clients that actually use it. The route was created somewhat speculatively in the early development stage.

padraic-shafer:
> I think 404 is right in this case. The user is providing an ID, and this follows the principle, "If this isn't yours, you're now allowed to know whether or not it exists."

Oh, right.  That make sense.

tiled/server/authentication.py
padraic-shafer: In OAuth2, I thought that once the refresh token is received by the requesting server, it should only be sent to the client via an http-only cookie. That way a long-lifetime refresh token is not available to client-side scripts or from other origins.  This is the cookie I was imagining should get deleted/expired for tidiness rather than security (because the session was already revoked) so that it does not interfere with subsequent login attempts or unauthenticated routes.

I suppose for the python client the http-only cookie has less meaning than for a typical web browser.

dylanmcreynolds: Cookies make things easy, but they get complicated in this world of SPAs that can talk to one or more backend services. If the tiled server is a different domain from the SPA talking to it, then the the SPA [won't have access to the cookie.](https://auth0.com/docs/manage-users/cookies/spa-authenticate-with-cookies)

Perhaps not quite on track for this issue, but here's on approach to [refresh tokens and SPAs.](https://auth0.com/blog/securing-single-page-applications-with-refresh-token-rotation/)

dylanmcreynolds (approved):

tiled/server/authentication.py
danielballan: In OAuth2, the refresh and access tokens never go near cookies. They are placed in the response body of the initial authorization response, and in the response body of each refresh. The client then holds the refresh token in internal storage somehow (could be browser local storage, or in the case of the Python client `~/.cache/tiled/tokens/...` with restrictive permissions on that file). The client places the access token in the header of future requests, as in `Authorization: Bearer {access_token}`.

danielballan: You can follow that logic in the [Authentication Details](https://blueskyproject.io/tiled/reference/authentication.html) docs. You'll see the `access_token` and `refresh_token` in a response body. Notice that the only time `set-cookie` shows up is in relation to [CSRF](https://developer.mozilla.org/en-US/docs/Glossary/CSRF) protection, holding a random unique string that is not a token.

tiled/server/authentication.py
```diff
@@ -894,8 +894,29 @@ async def refresh_session(
...
+    "Mark a Session as revoked so it cannot be refreshed again."
+    request.state.endpoint = "auth"
+    payload = decode_token(refresh_token.refresh_token, settings.secret_keys)
+    session_id = payload["sid"]
+    # Find this session in the database.
+    session = await lookup_valid_session(db, session_id)
+    if session is None:
+        raise HTTPException(409, detail=f"No session {session_id}")
+    session.revoked = True
+    db.add(session)
```
padraic-shafer, line 913: In SQLAlchemy, is [`Session.add()`](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.add) an "upsert" operation -- it will perform INSERT or UPDATE as needed?

tiled/server/authentication.py
```diff
@@ -1095,12 +1116,13 @@ async def whoami(
     )
 
 
-@base_authentication_router.post("/logout")
+@base_authentication_router.post("/logout", include_in_schema=False)
```
padraic-shafer, line 1119: This might be a moot point because it's already removed for the schema, but wanted to point out here the `deprecated` keyword as a possibility.

```suggestion
@base_authentication_router.post("/logout", include_in_schema=False, deprecated=True)
```

padraic-shafer (approved):
This looks great.  A couple of minor questions in the review.

tiled/server/authentication.py
danielballan: Moot but good to know about!

tiled/server/authentication.py
danielballan: https://docs.sqlalchemy.org/en/20/orm/session_basics.html#adding-new-or-existing-items

danielballan: (As usual with SQLAlchemy, the documentation is sprawling and specific things can be hard to find.)

PR #609, opened 2023-11-18
Test creation of API keys for other principals

padraic-shafer:
This PR adds a test to verify that an admin can create API keys using the route `/principal/{uuid}/apikey` and that non-admins cannot.  This test also verifies that a user cannot bypass the scopes checks by passing an empty scopes list when creating an API key and/or when requesting an admin resource.

The test is fairly long (in lines of code, not in execution speed).  I could probably break it into several individual tests that are backed by some common fixtures.  Do you have advice on keeping this more concise or clear?


danielballan:
To my taste, a long-test that "tells a story" is more readable that lots of small tests that tell part of a story. This one is borderline too long, I agree, but I don't see a clean place to break it out or any obvious resusable abstraction that would be a natural fixture.

I do wonder if we should take this moment to promote `create_api_key_other_principal` to a method on `Context`. (I mentioned on Slack that I've found reasons to do this, and I usually end up doing it via the Swagger UI or via the `http` CLI, writing URLs by hand.)

I had in the back of my mind the idea of making an `admin` namespace hanging off of context, like `c.context.admin.create_api_key_other_principal`. First, this make the `c.context.<TAB>` namespace less cluttered with methods that will not be applicable to most users. Second, it communicates that not just _any_ user can impersonate another user, which might otherwise prompt confusion or concern.

padraic-shafer:
> I had in the back of my mind the idea of making an admin namespace hanging off of context, like c.context.admin.create_api_key_other_principal. First, this make the c.context.<TAB> namespace less cluttered with methods that will not be applicable to most users. Second, it communicates that not just any user can impersonate another user, which might otherwise prompt confusion or concern.

This sounds like a good idea to me.

padraic-shafer:
> To my taste, a long-ish test that "tells a story" is more readable than lots of small tests that tell part of a story. This one is borderline too long, I agree, but I don't see a clean place to break it out or any obvious resusable abstraction that would be a natural fixture.

Thanks for affirming this.  I have two thoughts on how to simplify this, without too much effort.  I'm out of time for now though.  I'll move this PR to draft and attempt to resume it later this week.

padraic-shafer:
I've split the new long test into several smaller tests, that I think are easier to digest. This PR is ready for review.

dylanmcreynolds (approved):

dylanmcreynolds (approved):

tiled/_tests/test_authentication.py
```diff
@@ -449,3 +449,151 @@ def test_sticky_identity(enter_password, config):
     # Clear the default.
     clear_default_identity(context.api_uri)
     get_default_identity(context.api_uri) is None
+
+
+@pytest.fixture
+def principals_context(enter_password, config):
```
danielballan, line 455: Clever!

tiled/_tests/test_authentication.py
```diff
@@ -449,3 +449,151 @@ def test_sticky_identity(enter_password, config):
...
+        principal_uuid = principals_context["uuid"][username]
+        api_key = _create_api_key_other_principal(
+            context=context, uuid=principal_uuid, scopes=scopes
+        )
+        assert api_key
+        context.logout()
+
+        url_with_key = f"{resource}?api_key={api_key}"
+        context.http_client.get(url_with_key).raise_for_status()
+        context.http_client.cookies = None
```
danielballan, line 514: This is unnecessary and might in the future break CSRF-related things, if the test happens to evolve to include a `POST` or `PUT` example. Better to remove it, I think.

tiled/_tests/test_authentication.py
```diff
@@ -449,3 +449,151 @@ def test_sticky_identity(enter_password, config):
...
+            # Try with/without key, with/without empty scopes
+            for query_params in (
+                f"api_key={api_key}",
+                "scopes=[]",
+                f"api_key={api_key}&scopes=[]",
+            ):
+                url_query = f"{resource}?{query_params}"
+                with fail_with_status_code(401):
+                    context.http_client.get(url_query).raise_for_status()
+                context.http_client.cookies = None
```
danielballan, line 584: Ditto

tiled/_tests/test_authentication.py
```diff
@@ -449,3 +449,151 @@ def test_sticky_identity(enter_password, config):
...
+            context.authenticate(username="alice")
+
+        principal_uuid = principals_context["uuid"][username]
+        api_key = _create_api_key_other_principal(
+            context=context, uuid=principal_uuid, scopes=scopes
+        )
+        assert api_key
+        context.logout()
+
+        url_with_key = f"{resource}?api_key={api_key}"
```
danielballan, line 512: The API key can be placed in three places:

1. Query parameter `?api_key=SECRET`
2. Header as `Authorization:Apikey SECRET`
3. Cookie `tiled_api_key=SECRET`

The query parameter method is convenient because it can be shared by copy/paste, but it's also less secure. (For one, URLs tend to be logged. For another, if it is used in a browser, JavaScript on the page has access to the URL.) There have been some discussions about making support for API keys in query parameters configurable, and maybe even off by default, for security.

With that in mind, for future proofing, use

```py
context.api_key = api_key
```

which will set the `Authorization` header:

https://github.com/bluesky/tiled/blob/204676bbb841d252c1413abfe2646c3b7d27a525/tiled/client/context.py#L335-L341

and then you can simple use `context.http_client.get(resource)` below.

tiled/_tests/test_authentication.py
```diff
@@ -449,3 +449,151 @@ def test_sticky_identity(enter_password, config):
...
+        for resource in ("/api/v1/auth/principal", "/api/v1/array/full/A1"):
+            # Try with/without key, with/without empty scopes
+            for query_params in (
+                f"api_key={api_key}",
+                "scopes=[]",
+                f"api_key={api_key}&scopes=[]",
+            ):
+                url_query = f"{resource}?{query_params}"
+                with fail_with_status_code(401):
+                    context.http_client.get(url_query).raise_for_status()
```
danielballan, line 583: The `.get(...)` method takes a `params` dict, which you can use to avoid doing the string munging yourself here. Also, see comment above about preferring `context.api_key`.

tiled/_tests/utils.py
```diff
@@ -13,7 +13,7 @@
 @contextlib.contextmanager
 def fail_with_status_code(status_code):
     with pytest.raises(httpx.HTTPStatusError) as info:
-        yield
+        yield info
```
danielballan, line 16: :+1:

danielballan (changes requested):
This is good.

I left some suggestions for future-proofing and streamlining a bit.

tiled/_tests/test_authentication.py
padraic-shafer: That's helpful. Thanks!

tiled/_tests/test_authentication.py
padraic-shafer: Previously that was a sledghammer logout.   Now `context.api_key = None` fulfills that purpose.

tiled/_tests/test_authentication.py
padraic-shafer: Nice tip!

tiled/_tests/test_authentication.py
padraic-shafer: Done.  Now using `None` as the default for `context.api_key` (_above_) automatically logs out the key that previously was persisted as a cookie.

padraic-shafer:
@danielballan All recommendations have been incorporated. This is ready for review.

danielballan (approved):

PR #621, opened 2023-12-13
Fix public access

danielballan:
Fix long-standing bug where public access on an otherwise authenticated server became broken. One commit reproduces the error:

```
self = <tiled.access_policies.SimpleAccessPolicy object at 0x7f9898453910>, principal = <SpecialUsers.public: 'public'>

    def _get_id(self, principal):
        # Get the id (i.e. username) of this Principal for the
        # associated authentication provider.
>       for identity in principal.identities:
E       AttributeError: 'SpecialUsers' object has no attribute 'identities'

tiled/access_policies.py:52: AttributeError
```

and the next fixes it.

### DIFF
--- a/tiled/_tests/test_access_control.py
+++ b/tiled/_tests/test_access_control.py
@@ -26,6 +26,7 @@ def tree_b(access_policy=None):
 def context(tmpdir_module):
     config = {
         "authentication": {
+            "allow_anonymous_access": True,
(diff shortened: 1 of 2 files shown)


dylanmcreynolds (approved):

dylanmcreynolds (approved):
Looks great. Thanks!

PR #628, opened 2023-12-29
Create Scientific Metadata Search Engine for PostgreSQL (FullText)

### LINKED ISSUE #457: Support FullText search in PostgreSQL

Kezzsim:
fix #457

This does a number of things to provide basic $FullText search that is idempotent to mongoDB without using any third party indexers. Utilizes Postgres's built in functions [ts_vector & ts_query](https://www.postgresql.org/docs/current/functions-textsearch.html). Both of these methods appear to support `jsonb` columns out of the box.

Progress:
- [x] Create `metadata_search` index during `tiled catalog init` in `orm.py`
- [x] Add fullText as a query method to `adapter.py`
- [ ] Add fullText logic to `adapter.py`
- [ ] Create Alembic migration to add `metadata_search` index to existing Tiled deployments
- [ ] Enable fullText pytests

---
  - https://app.asana.com/0/0/1206267196412362

### DIFF
--- a/docs/source/reference/python-client.md
+++ b/docs/source/reference/python-client.md
@@ -138,6 +138,7 @@ Tiled currently includes two clients for each structure family:
    tiled.client.base.BaseClient.formats
    tiled.client.base.BaseClient.metadata
    tiled.client.base.BaseClient.uri
+   tiled.client.base.BaseClient.structure_family
    tiled.client.base.BaseClient.item
    tiled.client.base.BaseClient.login
    tiled.client.base.BaseClient.logout
--- a/tiled/client/base.py
+++ b/tiled/client/base.py
@@ -172,6 +172,11 @@ class BaseClient:
         "Direct link to this entry"
         return self.item["links"]["self"]
 
+    @property
+    def structure_family(self):
+        "Quick access to this entry"
+        return StructureFamily[self.item["attributes"]["structure_family"]]
+
     def new_variation(self, structure_clients=UNCHANGED, **kwargs):
         """
         This is intended primarily for internal use and use by subclasses.


tiled/catalog/adapter.py
```diff
@@ -975,6 +976,11 @@ def contains(query, tree):
     return tree.new_variation(conditions=tree.conditions + [condition])
 
 
+def fullText(query, tree):
```
danielballan, line 979: [PEP 8](https://peps.python.org/pep-0008/#function-and-variable-names):

> Function names should be lowercase, with words separated by underscores as necessary to improve readability.

> Variable names follow the same convention as function names.

> mixedCase is allowed only in contexts where that’s already the prevailing style (e.g. threading.py), to retain backwards compatibility.

danielballan:
While beginning to try to understand this feature for review, I came across an article that you may have also found: https://amitosh.medium.com/full-text-search-fts-with-postgresql-and-sqlalchemy-edc436330a0c

The approach declares a "computed" column and runs searches against that. I wonder how this compares to the PR’s approach, declaring an index with:

```py
func.to_tsvector(text("'simple'"), func.lower("metadata")),
```

danielballan:
Ah, indeed; I read the Issue and the PR out of order. You found the same article. Worth a close look, I think.

tiled/catalog/adapter.py
Kezzsim: Added in [b713504fddbb7f0c064e27900dbcc97473365ceb](https://github.com/bluesky/tiled/pull/628/commits/b713504fddbb7f0c064e27900dbcc97473365ceb)

PR #633, opened 2024-01-11
Make structure family a property on the client side

jmaruland:
Adding this feature on the client side to have a quick and easy access to structure family.

### DIFF
--- a/tiled/client/base.py
+++ b/tiled/client/base.py
@@ -172,6 +172,11 @@ class BaseClient:
         "Direct link to this entry"
         return self.item["links"]["self"]
 
+    @property
+    def structure_family(self):
+        "Quick access to this entry"
+        return self.item["attributes"]["structure_family"]
+
     def new_variation(self, structure_clients=UNCHANGED, **kwargs):
         """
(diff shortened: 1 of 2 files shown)


danielballan:
Thanks!

Can you add a line to the reference documentation? Around here:

https://github.com/bluesky/tiled/blob/adf5c0a0a33cfc74011f0bc04a4e2fdd0b954e6f/docs/source/reference/python-client.md?plain=1#L139

Then this is good to merge.

tiled/client/base.py
```diff
@@ -172,6 +172,11 @@ def uri(self):
         "Direct link to this entry"
         return self.item["links"]["self"]
 
+    @property
+    def structure_family(self):
+        "Quick access to this entry"
+        return self.item["attributes"]["structure_family"]
```
danielballan, line 178: Sorry, one more note that I didn't think of in the first pass. I think it would be nice if this returned the `StructureFamily` enum, e.g. `StructureFamily.array` instead of the string `'array'`.

danielballan (approved):

PR #635, opened 2024-01-20
Fix bug in slicing into container adapter

danielballan:
This bug was found at [SasView Contributor Camp](https://github.com/SasView/sasview/wiki/ContributorCampXII) while trying to integrate Tiled with the `sasdata` loader.

The first commit reproduces the error by extending an existing test:

```
E           TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

It also adds a new test, which does not exercise the bug, but nonetheless adds useful coverage to this general area of functionality.

The second commit fixes the bug.

### DIFF
--- a/tiled/_tests/test_container_files.py
+++ b/tiled/_tests/test_container_files.py
@@ -1,3 +1,5 @@
+import io
+
 import h5py
 import pandas
 import pytest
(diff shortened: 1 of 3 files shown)


danielballan:
Merging this to un-block some export work, but any post-merge review is still welcome!

padraic-shafer (commented):
This looks quite reasonable.

PR #637, opened 2024-01-23
Normalize `structure` into its own table

### LINKED ISSUE #576: Should we normalize the 'structure' column into its own table?
The tiled catalog database stores the `structure` JSON in a column the `data_sources` table. If many `data_sources` share the same structure, that same information is stored many times. For `array` structures, which are fairly compact, the overhead is not so high. For `table` structures, which include base64-encoded Arrow schemas, the overhead can be significant. For `awkward` structures, they can be even more so.

Should we store `structure` in a separate table, reference by foreign key, perhaps using a hash of the JSON to uniquely key it? The benefit is more compact storage for repeated structures, which does seem like it could be large win for many/most use cases, where duplicate structures are common.

Adding another table does add more to the total complexity. I have a small reservation about making them one step more complicated to explain, inspect, to quickly understand. On the other hand, this change could actually make it _easier_ to look at and explain: with the noisy, not-so-human-readable structure off to the side, the `data_sources` table will be possible to fit on a screen and read.

This came up in https://github.com/jpivarski/tiled-uproot/discussions/1 with @jpivarski.

danielballan:
The goal is avoid storing many duplicate copies of the same structure JSON.

As discussed in https://github.com/bluesky/tiled/issues/576, we uniquely identify a structure by hashing a canonical form of the JSON-encoded string, as defined by [RFC 8785](https://www.rfc-editor.org/rfc/rfc8785). We move the structure JSON into its own table and reference it via this hash.

Design decisions that need review:

- Use [jcs](https://pypi.org/project/jcs/), which distributes the Python 3 implementation from https://github.com/cyberphone/json-canonicalization/tree/master/python3. Of the implementations I reviewed, this was by far the most solid and actively maintained.
- Use MD5 for the hash. We need uniqueness and speed, not cryptographic security.
- Store the hash as HEX (32 characters of Unicode-typed data), trading some compactness to make it easier for developers to look at the tables with eyeballs, as shown below if you scroll right. I question this and I am open to feedback on it.

```
postgres=# select * from data_sources;
 id | node_id |                             mimetype                              | parameters | management |        time_created        |        time_updated        |           structure_id           
----+---------+-------------------------------------------------------------------+------------+------------+----------------------------+----------------------------+----------------------------------
  1 |       1 | application/vnd.openxmlformats-officedocument.spreadsheetml.sheet | {}         | external   | 2024-01-23 12:47:42.582651 | 2024-01-23 12:47:42.582651 | 37a6259cc0c1dae299a7866489dff0bd
  2 |       2 | image/tiff                                                        | {}         | external   | 2024-01-23 12:47:42.607718 | 2024-01-23 12:47:42.607718 | 9f3f87f38ccebf6368e13a1fcbd615d4
  3 |       3 | text/csv                                                          | {}         | external   | 2024-01-23 12:47:42.62236  | 2024-01-23 12:47:42.62236  | f299cdf0bf4f86ea320fe68d7407dfbf
  4 |       4 | image/tiff                                                        | {}         | external   | 2024-01-23 12:47:42.631224 | 2024-01-23 12:47:42.631224 | 9f3f87f38ccebf6368e13a1fcbd615d4
  5 |       5 | image/tiff                                                        | {}         | external   | 2024-01-23 12:47:42.638066 | 2024-01-23 12:47:42.638066 | 9f3f87f38ccebf6368e13a1fcbd615d4
  6 |       7 | multipart/related;type=image/tiff                                 | {}         | external   | 2024-01-23 12:47:42.650546 | 2024-01-23 12:47:42.650546 | 12a69ab45c9247f323dd7dea39b7c136
  7 |       8 | multipart/related;type=image/tiff                                 | {}         | external   | 2024-01-23 12:47:42.658897 | 2024-01-23 12:47:42.658897 | 37e3e82ef7cc8002d2f1eb3f7b2483d1
  8 |      10 | image/tiff                                                        | {}         | external   | 2024-01-23 12:47:42.670582 | 2024-01-23 12:47:42.670582 | 9f3f87f38ccebf6368e13a1fcbd615d4
  9 |      11 | image/tiff                                                        | {}         | external   | 2024-01-23 12:47:42.676989 | 2024-01-23 12:47:42.676989 | 9f3f87f38ccebf6368e13a1fcbd615d4
(9 rows)
```

That's about all there is to it. It came together surprisingly simply. The only complex part is the database migration, and even that was not so bad.

Closes #576 

This is the last planned change to the database, so once this is in, I plan to finally write an Explanation-style documentation page on the Catalog database.


danielballan:
@jpivarski If you have any comments I would welcome them.

padraic-shafer:
> * Use [jcs](https://pypi.org/project/jcs/), which distributes the Python 3 implementation from https://github.com/cyberphone/json-canonicalization/tree/master/python3. Of the implementations I reviewed, this was by far the most solid and actively maintained.

This looks reasonable.  Is Apache 2.0 license ok?

> * Use MD5 for the hash. We need uniqueness and speed, not cryptographic security.

Seems like a good tradeoff.

> * Store the hash as HEX (32 characters of Unicode-typed data), trading some compactness to make it easier for developers to look at the tables with eyeballs, as shown below if you scroll right. I question this and I am open to feedback on it.

Yes, this is useful.

danielballan:
> Is Apache 2.0 license ok?

Good question. Yes, as a dependency. Since we are importing it, not copying and vendoring it, we don't need to add anything to our LICENSE.

padraic-shafer (approved):
This works for me

tiled/catalog/adapter.py
```diff
@@ -574,13 +581,45 @@ async def create_node(
...
+                else:
+                    # Obtain and hash the canonical (RFC 8785) representation of
+                    # the JSON structure.
+                    structure = _prepare_structure(
                         structure_family, data_source.structure
-                    ),
+                    )
+                    structure_id = compute_structure_id(structure)
+                    # The only way to do "insert if does not exist" i.e. ON CONFLICT
+                    # is to invoke dialect-specific insert.
```
dylanmcreynolds, line 594: Can you avoid the dialect-specific clause by doing the search first? You're in the same transaction.

Alternatively, if I'm reading correctly, can you import the `insert` within the dialect select condition, then do the actual work in one piece of code?


```
if self.context.engine.dialect.name == "sqlite":
  from sqlalchemy.dialects.sqlite import insert

elif self.context.engine.dialect.name == "postgresql":
   from sqlalchemy.dialects.postgresql import insert

else:
   assert FALSE

statement = (insert(orm.Structure).values(
    id=structure_id,
    structure=structure,
    )
).on_conflict_do_nothing(index_elements=["id"])
await db.execute(statement)

```

tiled/catalog/utils.py
```diff
@@ -1,7 +1,10 @@
+import hashlib
 import re
 from pathlib import Path
 from urllib.parse import urlparse, urlunparse
 
+import jcs
```
dylanmcreynolds, line 6: I've never used a canonicalizing library in python, but does [conoicaljson](https://pypi.org/project/canonicaljson/) maybe look like a safer bet? It has declared `2.0.0` vs `0.21.x`, and doesn't rely on a different project to publish to PyPi (which is a over a year behind in the case of `jcs`)

dylanmcreynolds (commented):
Left a couple commment in.

tiled/catalog/utils.py
danielballan: Good find, I had missed that one!

OK, I'll switch. Either way, the surface area is a single function call (`compute_structure_json`) so this will be a very easy refactor if we have to move in the future.

jpivarski:
> @jpivarski If you have any comments I would welcome them.

What you chose are all better ideas than I can think of.

* Normalizing JSON by round-tripping through `json.dumps(json.loads(x))`. Pro: no third-party library dependence. Con: depends on the stability of Python's `json.dumps` formatting, which is unlikely to change, but not guaranteed. These third-party libraries canonicalize the JSON to a specified standard.
* Getting the hash by `hash(freeze_dicts(json.loads(x)))` where `freeze_dicts` replaces dicts with frozendicts. It should hash on the AST objects and be free of third-party libraries. But again, this method depends on Python not changing or making the database depend on Python version.
* Displaying the hash with Base64 [or higher](https://en.wikipedia.org/wiki/Binary-to-text_encoding#Encoding_standards) to make it smaller in the table than hexidecimal. But as hexidecimal, `structure_id` is no wider than `time_updated`; it doesn't seem to be a priority and hex is familiar.

tiled/catalog/utils.py
danielballan: Done!

tiled/catalog/adapter.py
danielballan: Good call. I refactored for now. It's possible we can remove the dialect-specific clause too. I need to run some experiments to check my understanding of transactions in SQLAlchemy and how they map onto the `SAVEPOINT` concept in PG and SQlite. Leaving that for a follow-up.

danielballan:
OK, thanks @jpivarski! I'm glad to think through those alternatives. I agree that keeping the SQL free of Python-isms is nice. If anyone wanted to make the case for base64 (or higher) I wouldn't put up resistance but I think agree that hex wins out because it's familiar, and the size is not really "large" on the scale of every else.

<hr />

@martintb was kind enough to share a Tiled database in use at NIST. It holds ~60k data sources but only 6 unique structures. This is exactly the kind of situation this change is designed to address. In this case, the structure descriptions happen to be fairly compact: they are all `array`, as opposed to the more verbose `table` or `awkward` structure JSON. But, as they a  little longer than hex-encoded MD5, we should still see some savings with this migration.

The initial database was 1.506 GB. The migration initially expands the footprint slightly, to 1.514 GB, because it leaves behind partially-filled database pages, holes where the in-lined `structure` column used to be. Running `VACCUM`, which I've now added as a final step in the SQLite migration, brings the size down to 1.496 GB.

```
-rw-rw-r-- 1 dallan dallan 1506365440 Jan 24 08:00 nist_catalog.orig.db
-rw-rw-r-- 1 dallan dallan 1514786816 Jan 24 08:08 nist_catalog.migrated.db
-rw-rw-r-- 1 dallan dallan 1496068096 Jan 24 08:10 nist_catalog.vacuumed.db
```

When we have #90 it will be straightforward to load this into a PostgreSQL based Tiled and run the same test. As #90 is next on my work list, I'll defer a PostgreSQL test until we have that.

danielballan:
Re-tested SQLite and PG migrations after rebaase. All looks good.

PR #638, opened 2024-01-23
Add Python 3.12 to CI matrix.

danielballan:
We intend to keep 3.8 (and maybe temporarily reinstate 3.7) up to v0.1.0 so that there is _a_ stable version installable in older environments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
         - "3.9"
         - "3.10"
         - "3.11"
+        - "3.12"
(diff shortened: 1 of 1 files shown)


danielballan:
We are waiting on `numba` to support 3.12. They have an RC out, so this should not be far off. Tracked at https://github.com/numba/numba/issues/9197

danielballan:
Now the issue is blosc, which has not published a Python 3.12 wheel.

danielballan:
Superseded by #719

PR #640, opened 2024-01-24
Scientific Metadata Search Engine (Fulltext) implementation for PostgreSQL 🔎

### LINKED ISSUE #457: Support FullText search in PostgreSQL

Kezzsim:
This feature adds a rudimentary scientific metadata search engine, implemented purely through Postgresql's native `ts_vector` and `ts_query` operations. Documented as part of the [Tiled client's FullText query](https://blueskyproject.io/tiled/reference/generated/tiled.queries.FullText.html#tiled.queries.FullText).

Tasks:
- [x]  Add fulltext logic to `adapter.py` to enable functionality
- [x]  Create a new computed index `metadata_search` for storing `jsonb_to_tsvector` data
- [x]  Generate Alembic migration to move existing tiled collections to the new standard
- [ ]   Add new tests that ensure the index is being invoked 

Resolves #457

### DIFF
--- a/tiled/_tests/test_queries.py
+++ b/tiled/_tests/test_queries.py
@@ -159,7 +159,7 @@ def test_contains(client):
 
 
 def test_full_text(client):
-    if client.metadata["backend"] in {"postgresql", "sqlite"}:
+    if client.metadata["backend"] in {"sqlite"}:
 
         def cm():
             return fail_with_status_code(400)
--- a/tiled/_tests/test_queries.py
+++ b/tiled/_tests/test_queries.py
@@ -45,6 +45,9 @@ mapping["does_contain_z"] = ArrayAdapter.from_array(
 mapping["does_not_contain_z"] = ArrayAdapter.from_array(
     numpy.ones(10), metadata={"letters": list(string.ascii_lowercase[:-1])}
 )
+mapping["full_text_test_case"] = ArrayAdapter.from_array(
+    numpy.ones(10), metadata={"color": "purple"}
+)
 
 mapping["specs_foo_bar"] = ArrayAdapter.from_array(numpy.ones(10), specs=["foo", "bar"])
 mapping["specs_foo_bar_baz"] = ArrayAdapter.from_array(
@@ -168,6 +171,9 @@ def test_full_text(client):
         cm = nullcontext
     with cm():
         assert list(client.search(FullText("z"))) == ["z", "does_contain_z"]
+        # plainto_tsquery fails to find certain words, weirdly, so it is a useful
+        # test that we are using tsquery
+        assert list(client.search(FullText("purple"))) == ["full_text_test_case"]
 
 
 def test_regex(client):
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -1073,7 +1073,7 @@ CatalogNodeAdapter.register_query(KeysFilter, keys_filter)
 CatalogNodeAdapter.register_query(StructureFamilyQuery, structure_family)
 CatalogNodeAdapter.register_query(SpecsQuery, specs)
 CatalogNodeAdapter.register_query(FullText, full_text)
-# TODO: FullText [in progress], Regex
+# TODO: Regex
 
 
 def in_memory(
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -1008,7 +1008,7 @@ def full_text(query, tree):
         tsvector = func.jsonb_to_tsvector(
             cast("simple", REGCONFIG), orm.Node.metadata_, cast(["string"], JSONB)
         )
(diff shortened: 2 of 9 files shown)


tiled/catalog/orm.py
```diff
@@ -95,6 +96,12 @@ class Node(Timestamped, Base):
             "metadata",
             postgresql_using="gin",
         ),
+        # This index supports full-text search.
+        Index(
+            "metadata_search",
```
danielballan, line 101: I think it's worth giving this a pretty descriptive name, since we anticipate adding others in the future. I could see potentially adding other types of text indexes, like trigram etc. and it will help if the names are unambigious.

I'm not quite sure how far to take it. Maybe `metadata_tsvector`? Or `metadata_tsvector_case_insensitive`?

tiled/catalog/adapter.py
```diff
@@ -997,6 +1001,21 @@ def contains(query, tree):
     return tree.new_variation(conditions=tree.conditions + [condition])
 
 
+def full_text(query, tree):
+    dialect_name = tree.engine.url.get_dialect().name
+    conditions = []
```
danielballan, line 1006: This pattern from other query implementations doesn't fit as well here, as we always add exactly one condition, as opposed to accumulating a list of conditions in loop.

Can probably just do

```py
condition = tsvector(...)
...
return tree.new_variation(conditions=tree.conditions + [condition])
```

tiled/catalog/migrations/versions/1cd99c02d0c7_create_index_for_fulltext_search.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def upgrade():
+    connection = op.get_bind()
+    if connection.engine.dialect.name == "postgresql":
+        with op.get_context().autocommit_block():
+            # There is no sane way to perform this using op.create_index()
+            op.execute(
+                """
+                CREATE INDEX metadata_search
```
danielballan, line 30: Note to selves: if we change index name in `orm.py` as suggested below, make sure to change it here too.

danielballan (commented):
Very excited to see this!

danielballan:
Also: our `FullText` query accepts a `case_sensitive` parameter. I added that somewhat speculatively, when I was  more naively about what is involved in supporting both options. As we're still in alpha, I wonder if it makes sense to remove that for now---being case insensitive only---and add it back later if there is demand for it.

There are tradeoffs in insert speed and database size that make me question whether case sensitive search is useful and important enough to justify that.

https://github.com/bluesky/tiled/blob/fca4330fc090ca577cdf764d435eb5160a22021e/tiled/queries.py#L42-L50

tiled/catalog/orm.py
danielballan: This is running against SQLite engines as well and failing a slew of tests. Needs to be made conditional on dialect somehow.

tiled/_tests/test_queries.py
```diff
@@ -168,6 +171,9 @@ def cm():
         cm = nullcontext
     with cm():
         assert list(client.search(FullText("z"))) == ["z", "does_contain_z"]
+        # plainto_tsquery fails to find certain words, weirdly, so it is a useful
+        # test that we are using tsquery
+        assert list(client.search(FullText("purple"))) == ["full_text_test_case"]
```
dylanmcreynolds, line 176: I don't know how much to expect of this index, but could we do partial word search? e.g. `urple`

Kezzsim:
Current C.I. issues are caused by the caching we set up previously, ORM and alembic need to be able to handle if an index already exists in that cache somehow.

Kezzsim:
> Also: our `FullText` query accepts a `case_sensitive` parameter. I added that somewhat speculatively, when I was more naively about what is involved in supporting both options. As we're still in alpha, I wonder if it makes sense to remove that for now---being case insensitive only---and add it back later if there is demand for it.
> 
> There are tradeoffs in insert speed and database size that make me question whether case sensitive search is useful and important enough to justify that.
> 
> https://github.com/bluesky/tiled/blob/fca4330fc090ca577cdf764d435eb5160a22021e/tiled/queries.py#L42-L50

I will remove the case sensitive flag from both the source code and the documentation, minimizing the API surface impact by ignoring any other kwargs sent to query other than `text`.

The [postgresql documentation](https://www.postgresql.org/docs/current/textsearch-intro.html#TEXTSEARCH-INTRO) writes:
> A lexeme is a string, just like a token, but it has been normalized so that different forms of the same  word are made alike. For example, normalization almost always includes folding upper-case letters  to lower-case

tiled/_tests/test_queries.py
danielballan: I think support is limited, but robust fuzzy text search is next up for @Kezzsim.

PR #641, opened 2024-01-25
Get asv benchmark suite running again.

danielballan:
Tiled has a nascent (nascent!) suite of benchmarks using [asv](https://asv.readthedocs.io/en/v0.6.1/). They have languished; the last update was in October 2021.

This PR gets them back into working order. Here is an example, running on this PR branch:

```
$ asv run --environment existing:same
· Discovering benchmarks
· Running 3 total benchmarks (1 commits * 1 environments * 3 benchmarks)
[ 0.00%] ·· Benchmarking existing-py_home_dallan_micromamba_envs_py311_bin_python3.11
[16.67%] ··· Running (benchmarks.TimeSuite.time_list_tree--)...
[66.67%] ··· benchmarks.TimeSuite.time_list_tree                                                                                                                                                 8.92±0.5ms
[83.33%] ··· benchmarks.TimeSuite.time_lookup                                                                                                                                                    4.63±0.2ms
[100.00%] ··· benchmarks.TimeSuite.time_lookup_and_read                                                                                                                                           12.9±0.4ms
```

Note: asv is designed to create dedicated conda environments to run its benchmarks in. This facilitates testing against different versions of Python, numpy, etc. if you configure it so. It is designed to run across ranged of commits, tracking a project's history, and publishing the results as nice plots in a static web site.

For quick tests though, I like the above incantation, which just runs the current commit in the current Python environment.

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -1,42 +1,29 @@
 # Write the benchmarking functions here.
 # See "Writing benchmarks" in the asv docs for more information.
-import subprocess
-import time
 
-from tiled.client import from_uri
+import numpy
 
-HOST = "0.0.0.0"
-PORT = 9040
+from tiled.adapters.array import ArrayAdapter
+from tiled.adapters.mapping import MapAdapter
+from tiled.client import Context, from_context
+from tiled.server.app import build_app
 
 
 class TimeSuite:
-    """
-    An example benchmark that times the performance of various kinds
-    of iterating over dictionaries in Python.
-    """
-
     def setup(self):
-        self.server_process = subprocess.Popen(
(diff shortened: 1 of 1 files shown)


benchmarks/benchmarks.py
```diff
@@ -1,42 +1,29 @@
...
 
 
 class TimeSuite:
-    """
-    An example benchmark that times the performance of various kinds
-    of iterating over dictionaries in Python.
-    """
-
     def setup(self):
-        self.server_process = subprocess.Popen(
```
danielballan, line 19: Notice that instead of starting a server in a subprocess, this PR runs the server and client in the same process. Messages are passed directly via Python function calls, the same as it works in unit tests. This takes networking out of the equation and will give me stable results, focused on actual Tiled performance.

padraic-shafer (approved):
👍

PR #646, opened 2024-01-31
Npm build prefix

tacaswell:
Still not quite right, but this + buliding from the sdist + seting the root_path on unvicorn + a matching nginx config half works.


danielballan (approved):

danielballan:
Thanks @tacaswell!

PR #649, opened 2024-02-07
Add support for directly downloading raw Assets

danielballan:
This PR adds two new routes to the server and some methods to the Python client objects for using (and unit testing) them.

## User-facing usage from the Python client

Let's start a server and upload a small array.

```
```

```
c = from_uri("http://localhost:8000", api_key="secret", include_data_sources=True)
c.write_array([1, 2, 3], key="x")
```

There are three ways to get this data back out:

### 1. Direct access to the file system.

This treats Tiled merely as a "directory service" that knows where the files are.

```py
>>> c.data_sources()
[{'id': 1,
  'structure': {'data_type': {'endianness': 'little',
    'kind': 'i',
    'itemsize': 8},
   'chunks': [[3]],
   'shape': [3],
   'dims': None,
   'resizable': False},
  'mimetype': 'application/x-zarr',
  'parameters': {},
  'assets': [{'data_uri': 'file://localhost/tmp/tmpahgtx_eu/data/x',
    'is_directory': True,
    'parameter': 'data_uri',
    'num': None,
    'id': 1}],
  'management': 'writable'}]
```

You can see `'file://localhost/tmp/tmpahgtx_eu/data/x'` in there. A convenience function can be used to pick out the relevant field and convert it to a `Path`. (Of course, this won't work for future non-`file:` based storage.)

```py
>>> from tiled.client.utils import get_filepaths
>>> get_asset_filepaths(c['x'])
[PosixPath('/tmp/tmpahgtx_eu/data/x')]
```

This is helpful for users that have direct access to the relevant filesystems and may have faster ways of loading the data than HTTP.

### 2. Download of raw assets via Tiled

The new method `raw_export()` downloads the asset file(s) through HTTP. Like clicking "Download" in a web app, the method requires no parameters. It names the files as suggested by the server (via `content-disposition` header). This is a simplest "Just get the files" use case.

Users can provide a destination directory (default is `cwd`) and can tune the number of parallel threaded workers that download data.

Progress bars are shown.

```py
>>> c["x"].raw_export()
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0.0% • 0/100 bytes • ? • -:--:--
[17:33:55] Requesting http://localhost:8000/api/v1/asset/bytes/x?id=1&relative_path=.zarray                                            download.py:37
           Requesting http://localhost:8000/api/v1/asset/bytes/x?id=1&relative_path=0                                                  download.py:37
           Downloaded /home/dallan/Repos/bnl/tiled/0                                                                                   download.py:59
           Downloaded /home/dallan/Repos/bnl/tiled/.zarray                                                                             download.py:59
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 312/312 bytes • ? • 0:00:00
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 40/40 bytes   • ? • 0:00:00
Out[10]: 
[PosixPath('/home/dallan/Repos/bnl/tiled/.zarray'),
 PosixPath('/home/dallan/Repos/bnl/tiled/0')]
```

Users can only download assets that are associated with nodes that they have access to.

### 3. Logically rather than via the assets

The original Tiled way, addressing the data _logically_ instead of via assets, letting Tiled slice and/or transcode as wanted:

```py
>>> c['x'].read()  # Download directly into numpy memory
array([1, 2, 3])
>>> c["x"][0]  # Download selective slice into numpy
1
>>> c['x'].export('data.csv')  # Transcode server-side and save as desired format
```

## HTTP API

An asset may be backed by:

- a single file, like a TIFF
- a directory of managed files, like Zarr

Downloading a single-file assets looks like this:

```
/api/v1/metadata/{path}?include_data_sources=true  # body includes description of assets, including asset_ids
/api/v1/asset/bytes/{path}?id={asset_id}  # binary file
```

Downloading directory-backed assets looks like this:

```
/api/v1/metadata/{path}?include_data_sources=true   # body includes description of assets, including asset_ids
/api/v1/asset/manifest/{path}?id={asset_id}  # body includes a list of relative paths, inside the directory
/api/v1/asset/bytes/{path}?id={asset_id}&relative_path={relative_path}  # binary file
```

With this API design, the endpoint sends exactly one file on disk per response. This means we can use starlette's `FileResponse` to serve it, effectively making this part of Tiled into a simple static server. This buys us a couple nice things:

- The file is spooled from disk into the socket, chunk by chunk, so this scales well even for huge files.
- The file is `stat`-ed prior to sending, and `FileResponse` uses this to provide info to web caches, as well as `content-length`.

The alternative, attemping to zip or tar up directories, could put a lot of load on the server. I am confident that one-file-per-response is the right place to start, and it may be the right place to stay.


tiled/server/router.py
```diff
@@ -1215,3 +1219,104 @@ async def delete_revision(
 
     await entry.delete_revision(number)
     return json_or_msgpack(request, None)
+
+
+@router.get("/asset/bytes/{path:path}")
+async def get_asset(
+    request: Request,
+    id: int,
+    relative_path: Optional[Path] = None,
```
padraic-shafer, line 1228: I believe fastAPI got rid of optional path parameters in recent versions

tiled/server/router.py
```diff
@@ -1215,3 +1219,104 @@ async def delete_revision(
...
+        status_code=200,
+        headers={"Content-Disposition": f'attachment; filename="{filename}"'},
+    )
+
+
+@router.get("/asset/manifest/{path:path}")
+async def get_asset_manifest(
+    request: Request,
+    id: int,
+    entry=SecureEntry(scopes=["read:data"]),  # TODO: Separate scope for assets?
```
padraic-shafer, line 1296: Is it likely that the assets contain juicy information that is not already returned by read()?

tiled/_tests/test_asset_access.py
```diff
@@ -0,0 +1,69 @@
...
+def client(context):
+    client = from_context(context)
+    yield client
+
+
+def test_include_data_sources_method_on_self(client):
+    "Calling include_data_sources() fetches data sources on self."
+    client.write_array([1, 2, 3], key="x")
+    with pytest.warns(UserWarning):
+        # This fetches the sources with a second request.
```
padraic-shafer, line 31:
```suggestion
        # This fetches the sources with an additional implicit request.
```

tiled/_tests/test_asset_access.py
```diff
@@ -0,0 +1,69 @@
...
+        client["x"].data_sources()
+    client["x"].include_data_sources().data_sources() is not None
```
padraic-shafer, line 33: This initially looked like a tedious idiom.  However, as you pointed out at the appropriate spots, the user can:
1. initialize the client to always include data sources; or
2. have the sources included implicitly with an extra latency, and a suitable warning that they are free to ignore :)

Nicely done

padraic-shafer (changes requested):
The linter is complaining, but the content looks good.

I made a few comments, but none are really firm requirements.  WIth linting this could be merged.

tiled/client/download.py
```diff
@@ -0,0 +1,108 @@
...
+            f"Received {len(urls) = } and {len(paths) = }."
+        )
+
+    def sigint_handler(signum, frame):
+        done_event.set()
+        original_sigint_handler(signal.SIGINT, frame)
+
+    done_event = Event()
+    original_sigint_handler = signal.getsignal(signal.SIGINT)
+    signal.signal(signal.SIGINT, sigint_handler)
```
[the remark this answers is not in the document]
danielballan: Right. This modeled on an (MIT-licensed) example from `rich`, and that is where some of these variable names come from.

tiled/server/router.py
danielballan: I can imagine there might be some use cases for taking the hard-line position that files are internal and not public API, so that file formats can be changed with 100% certainty that no user workflow could be broken.

I think we can wait to carve out finer scopes until this use case emerges. We might get there someday, or we might not.

tiled/server/router.py
danielballan: I could be wrong, but I think this comment is conflating:

- URL path parameters like (`{path:path}`) above, which is consumed by the `entry` dependency below
- this `relative_path` _query_ parameter, given as `?relative_path=...`, which happens to be typed as `Path`

If query parameters of type `Path` cannot be optional, can you link to the docs/changelog?

tiled/client/base.py
```diff
@@ -177,18 +201,150 @@ def structure_family(self):
...
+
+    def raw_export(self, directory=None, max_workers=4):
+        """
+        Download the raw assets backing this node.
+
+        This may produce a single file or a directory.
+
+        Parameters
+        ----------
+        directory : Path, optional
```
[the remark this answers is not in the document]
danielballan: No, this is a good note. I'll change it.

tiled/_tests/test_asset_access.py
danielballan: Thanks!

tiled/client/base.py
```diff
@@ -177,18 +201,150 @@ def structure_family(self):
...
             **kwargs,
         )
 
+    def asset_manifests(self, data_sources):
+        """
+        Return a manifest of the relative paths of the contents in each asset.
+
+        This return a dictionary keyed on asset ID.
+        Assets backed by a single file are mapped to None (no manifest).
+        Asset backed by a directory of files are mapped to a list of relative paths.
```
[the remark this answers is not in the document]
danielballan: Makes sense. Renamed to `asset_manifest`.

danielballan:
I used `rich` for the progress bar because I _thought_ we already had a dependency on it through `typer`. It turns out that we don't, so now I'm wondering if we should use a more single-purpose progress bar library like `tqdm`. Notably, `tqdm` is already a dependency of `bluesky` and `databroker-pack`.

tiled/client/base.py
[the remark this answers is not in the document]
danielballan: Changed to `destination_directory`.

tiled/_tests/test_asset_access.py
```diff
@@ -0,0 +1,69 @@
...
+    yield client
+
+
+def test_include_data_sources_method_on_self(client):
+    "Calling include_data_sources() fetches data sources on self."
+    client.write_array([1, 2, 3], key="x")
+    with pytest.warns(UserWarning):
+        # This fetches the sources with an additional implicit request.
+        client["x"].data_sources()
+    client["x"].include_data_sources().data_sources() is not None
```
DiamondJoseph, line 33: Am I missing some magics somewhere that's making them implicit or are you missing a bunch of `assert`s?

DiamondJoseph (changes requested):

tiled/server/router.py
```diff
@@ -1215,3 +1219,104 @@ async def delete_revision(
...
+@router.get("/asset/bytes/{path:path}")
+async def get_asset(
+    request: Request,
+    id: int,
+    relative_path: Optional[Path] = None,
+    entry=SecureEntry(scopes=["read:data"]),  # TODO: Separate scope for assets?
+):
+    if not hasattr(entry, "asset_by_id"):
+        raise HTTPException(
+            status_code=405,
```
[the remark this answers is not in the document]
danielballan: I buy that. I happened to be looking at another codebase the other day that used them, and I had the same thought. It’s better.

We’ll do that in a separate PR.

tiled/server/router.py
```diff
@@ -1215,3 +1219,104 @@ async def delete_revision(
...
+        )
+
+    asset = await entry.asset_by_id(id)
+    if asset is None:
+        raise HTTPException(
+            status_code=404,
+            detail=f"This node exists but it does not have an Asset with id {id}",
+        )
+    if asset.is_directory:
+        if relative_path is None:
```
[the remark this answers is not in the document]
danielballan: Good question.

A directory is not something you can put into an HTTP response. You would have to read and combine all the files somehow into a single byte stream.

A reasonable way to do this would be to ZIP the directory, similarly to what happens when you download a directory from Google Drive. But this opens up a bunch more complexity for the server.

Zipping up a big directory might take longer than the HTTP client (or proxy) timeout internal, so the request would need to launch an asynchronous job, where the user is given a link that they can use to check on the status and eventually download the result. And the server would need some storage area to stage the ZIP archive for later download, with rules about how long to keep a given result and how big they are allowed to be.

This could all be done, but it seems simpler for the moment to do one file per request, spooled from disk using `FileResponse`, and let the client piece it together to make things simple for the user. The client could even ZIP it if desired.

Are we missing a simpler way to handle this?

tiled/server/router.py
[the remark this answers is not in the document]
danielballan: Captured in https://github.com/bluesky/tiled/issues/650

tiled/server/router.py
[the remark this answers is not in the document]
danielballan: Having said all that….it might be possible to incrementally construct and send a ZIP in chunks, reading a chunk of each file, appending it to the ZIP, and then draining the ZIP-in-progress into the socket. In this way, similar to `FileResponse` the  server memory usage is bounded, regardless of file or directory size.

I’ll look at this. If it works, then maybe we should accept `/asset/bytes/{path}?id=…` with `relative_path` optional, even for directory-backed assets, and stream a zip of the whole directory if no `relative_path` is given.

I think it would still be worth keeping the option of specifying `relative_path` for selective/incremental access to directory-based assets.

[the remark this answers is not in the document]
danielballan: This doesn't quite work, which I suppose is why Google Drive _et. al._ don't support it either. This [failed attempt](https://gist.github.com/danielballan/20dfb2b41902cc9396a13f434ca15045) illustrates the idea in my last comment as code.

ZIP places a header before the bytes from each file, and the header contains a hash (CRC-32) of the uncompressed file. Thus, the server needs to read the whole file in order to send the hash before it begins to sends the file's contents. That leaves us with two bad options:

1. Read a file fully into memory once, hash it, and send it.
2. Stream through each file in chunks, twice. The first time, compute the hash. Send the file header. The second time, send the contents in chunks.

(1) puts too much memory pressure on the server, and files may be very large. (2) requires reading the file from disk twice.

So I think we're back to, "There is no simple way to stream a directory over HTTP."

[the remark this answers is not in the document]
DiamondJoseph: That's fine, I think I misunderstood tiled/client/base `raw_export` then- presumably that will never support >1 data_source?

tiled/server/router.py
```diff
@@ -1215,3 +1219,104 @@ async def delete_revision(
...
+
+
+@router.get("/asset/bytes/{path:path}")
+async def get_asset(
+    request: Request,
+    id: int,
+    relative_path: Optional[Path] = None,
+    entry=SecureEntry(scopes=["read:data"]),  # TODO: Separate scope for assets?
+):
+    if not hasattr(entry, "asset_by_id"):
```
callumforrester, line 1231: Is there any value in also having a top-level server config to enable/disable asset downloading?

dylanmcreynolds: Is your concern the the fear that we are making it too easy to download huge things and may see unintended consequences?

callumforrester: Yes, but I also think it would be inkeeping with Tiled's versatile deployment philosophy. For example I can imagine one facility wanting to hide the raw assets away entirely and use Tiled as a watertight gateway, while another might require raw assets for legacy use cases.

tiled/server/router.py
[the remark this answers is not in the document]
danielballan: I think it could. `raw_export` takes a directory as its argument (defaults to `cwd`) and can put 1 or N files in there, with names provided by the server. Today, this can support use cases like:

- one HDF5 file (1 data source, 1 asset, 1 file)
- a Zarr dataset (1 data source, 1 asset, 1 directory of N files)
- an HDF5 with virtual datasets (1 data source, N assets, N files)
- a sequence of TIFFs (1 data source, N assets, N files)

All of the above are 1 data source. How would we ever end up with N data sources? A Bluesky stream can be backed by an HDF5 for image data and some tabular format (Arrow, or even CSV, anything...) for time series tabular data.

The SQL schema supports a one-to-many relation between logical nodes and data sources. The goal is to present the user with one logical entity, which just happens to be backed by a couple different data sources, and the user doesn't have to care. But both the server and client code have `NotImplementedErrors` sprinkled into these code paths for now.

I _think_ `raw_export` can be extended to deal with this when the time comes, since it already supports downloading N files.

tiled/server/router.py
danielballan: That seems like a good idea. And simpler than adding a scope.

padraic-shafer:
> I used `rich` for the progress bar because I _thought_ we already had a dependency on it through `typer`. It turns out that we don't, so now I'm wondering if we should use a more single-purpose progress bar library like `tqdm`. Notably, `tqdm` is already a dependency of `bluesky` and `databroker-pack`.

I lean toward making the client environment easy to use rather than lightweight.  Rich brings nice UX features, and the client dependencies already necessarily include some rather large packages. I think this is not the straw that breaks the camel's back.

tiled/server/router.py
danielballan: Added config setting named `expose_raw_assets` (default true) with test in [135dcac](https://github.com/bluesky/tiled/pull/649/commits/135dcac224796191de22b9cae391ca4efd4d6096). If `False`, the two new `/asset/...` routes always return `403 Forbidden` with an explanation.

padraic-shafer (approved):
Feedback from earlier reviews incorporated.

Tests and configuration support an `expose_raw_assets` option. Looks good.

tiled/server/router.py
padraic-shafer: That makes sense. The new `expose_raw_assets` config parameter helps in this regard.

tiled/server/router.py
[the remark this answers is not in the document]
danielballan: I'll mark this as resolved, then. Glad we explored the question and understand the limitations more specifically.

PR #653, opened 2024-02-11
Xarray specs bug

### LINKED ISSUE #651: Tiled client ``from_context`` keeps adding specs until pydantic validation fails
I'm trying to use the ``from_context`` function to create a local tiled server for use in writing tests.

After creating the client a number of times, it suddenly starts failing with pydantic Validation errors saying there are more than 20 spec objects. Here is the tail end of the exception.

```
File /local/mambaforge/envs/haven/lib/python3.10/site-packages/tiled/server/core.py:492, in construct_resource(base_url, path_parts, entry, fields, select_metadata, omit_links, show_sources, media_type, max_depth, depth)
    485         else:
    486             attributes["sorting"] = [
    487                 {"key": key, "direction": direction}
    488                 for key, direction in entry.sorting
    489             ]
    490 d = {
    491     "id": path_parts[-1] if path_parts else "",
--> 492     "attributes": schemas.NodeAttributes(**attributes),
    493 }
    494 if not omit_links:
    495     d["links"] = {
    496         "self": f"{base_url}/metadata/{path_str}",
    497         "search": f"{base_url}/search/{path_str}",
    498         "full": f"{base_url}/container/full/{path_str}",
    499     }

File /local/mambaforge/envs/haven/lib/python3.10/site-packages/pydantic/main.py:342, in pydantic.main.BaseModel.__init__()

ValidationError: 1 validation error for NodeAttributes
specs
  ensure this value has at most 20 items (type=value_error.list.max_items; limit_value=20)
```

Here is a minimum working example: https://gist.github.com/canismarko/8d5a1a1850dffa7e532de05449383fac

With some breakpoints, it looks like ``entry.specs`` keeps getting longer, adding a new ``Spec("xarray_data")`` (or something like that) each time.

This also happens if I create the client once, and then just read the data multiple times, i.e. move everything but ``data.read()`` outside that for loop.

It does not happen if I use a normal ``from_uri(...)`` client and access a running tiled server.

danielballan:
Closes #651

First commit reproduces the issue in a unit test, which fails. Second commit fixes the bug, and the unit test passes.

The xarray Dataset adapter was employing a Python anti-pattern here: mutating user input, specifically the list of `specs`. In the reported issue, the server was instantiating a variant of the adapter in a loop, via `new_variation()`, and the `specs` list was accumulating more items on each round. Eventually, it reached a server-imposed limit on the number of Specs that a node can carry (20) and raised an error, which brought the issue to attention.

The fix is to alter the logic to say, "If the user provides `specs`, use them. If the user provides no `specs` (`specs=None`) default to `specs=[Spec("xarray_dataset")]`.

### DIFF
--- a/tiled/_tests/test_xarray.py
+++ b/tiled/_tests/test_xarray.py
@@ -103,16 +103,25 @@ def test_specs_mutation_bug(client):
 
 
 def test_specs_override(client):
+    "The 'xarray_dataset' is appended to the end if not present."
+    ds = pandas.DataFrame({"x": numpy.array([1, 2, 3])}).to_xarray()
     tree = MapAdapter(
         {
-            key: DatasetAdapter.from_dataset(ds, specs=[Spec("test")])
-            for key, ds in EXPECTED.items()
+            "a": DatasetAdapter.from_dataset(ds, specs=[Spec("test")]),
+            "b": DatasetAdapter.from_dataset(
+                ds, specs=[Spec("xarray_dataset"), Spec("test")]
+            ),
+            "c": DatasetAdapter.from_dataset(
+                ds, specs=[Spec("test"), Spec("xarray_dataset")]
+            ),
         }
     )
     app = build_app(tree)
     with Context.from_app(app) as context:
         client = from_context(context)
-    assert client["image"].specs == [Spec("test")]
+    assert client["a"].specs == [Spec("test"), Spec("xarray_dataset")]
+    assert client["b"].specs == [Spec("xarray_dataset"), Spec("test")]
+    assert client["c"].specs == [Spec("test"), Spec("xarray_dataset")]
 
 
 @pytest.mark.parametrize("key", ["image", "weather"])
--- a/tiled/_tests/test_xarray.py
+++ b/tiled/_tests/test_xarray.py
@@ -1,6 +1,7 @@
 import dask.array
 import numpy
 import orjson
+import pandas
 import pytest
 import xarray
 import xarray.testing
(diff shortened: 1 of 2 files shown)


padraicshafer:
> The fix is to alter the logic to say, "If the user provides specs, use them. If the user provides no specs (specs=None) default to specs=[Spec("xarray_dataset")].

It looks like original intent of this code may have been to ensure that `Spec("xarray_dataset")` is in the list of specs. If the user forgets to include it in the initialization parameter, that no longer happens.

It seems reasonable that `Spec("xarray_dataset")` should always be in the list of specs for a Dataset node regardless of what the user (additionally) indicates in the `specs` parameter. In that case, set logic would be helpful.

```python3
        specs = set(specs or [])
        specs.add(Spec("xarray_dataset"))
        specs = list(specs)
```

Is this too heavy handed?

canismarko (approved):
I checked this against the code that originally triggered the bug. It passes now without error. Thanks, @danielballan.

danielballan:
@padraicshafer I am genuinely torn between “Let the user take control” and “Help the user avoid doing something they probably don’t want to do….”

Maybe we can wait for the “take control” use case to emerge and stick to your suggested approach until it arrives, if it arrives.

Order matters in specs (first one is supposed to be most specific) so I will amend the logic to be “Append xarray_dataset to the end of the list if it is not present somewhere in the list.”

danielballan:
I implemented a compromise: the core `__init__` lets you do whatever, but the user-facing classmethod constructor `from_dataset` appends `Spec("xarray_dataset")` to `specs` if it is not present in the user-provided list.

PR #657, opened 2024-02-15
Use POST for otherwise long GET urls

### LINKED ISSUE #579: Add `POST` option for long `GET` URL queries

padraic-shafer:
Closes #579.

This will eventually create POST alternatives to the existing `/container/full`, `/table/full`, and `/table/partition` GET endpoints.

Tests and updates will be rolled out one-by-one where feasible for Containers (xarray.Dataset, file directory, Excel file, Zarr file, HDF5 file) and Tables (pandas.DataFrame).

Will also update `/table/partition/{path}` to accept a `column` query parameter rather than `field`, as was recently done for `/table/full/{path}`. `field` will be deprecated but kept for backward compatibility for some duration.

### DIFF
--- a/tiled/_tests/test_xarray.py
+++ b/tiled/_tests/test_xarray.py
@@ -1,5 +1,3 @@
-from enum import IntEnum
-
 import dask.array
 import numpy
 import orjson
@@ -11,10 +9,10 @@ import xarray.testing
 from ..adapters.mapping import MapAdapter
 from ..adapters.xarray import DatasetAdapter
 from ..client import Context, from_context, record_history
-from ..client.base import BaseClient
 from ..serialization.xarray import serialize_json
 from ..server.app import build_app
 from ..structures.core import Spec
+from .utils import URL_LIMITS
 
 image = numpy.random.random((3, 5))
 temp = 15 + 8 * numpy.random.randn(2, 2, 3)
@@ -151,27 +149,6 @@ def test_wide_table_optimization_off(client):
     with record_history() as history:
         wide.read(optimize_wide_table=False)
     assert len(history.requests) >= 10
-
-
-class URL_LIMITS(IntEnum):
-    HUGE = 80_000
-    ORIGINAL = BaseClient.URL_CHARACTER_LIMIT
-    TINY = 10
-
-
-@pytest.fixture
-def url_limit(request: pytest.FixtureRequest):
-    """Adjust the URL length limit for the client's GET requests.
-
-    Data will be fetched by GET requests when the URL_CHARACTER_LIMIT is long,
-    and by equivalent POST requests when the URL_CHARACTER_LIMIT is short.
-    """
-    URL_CHARACTER_LIMIT = int(request.param)
-    # Temporarily adjust the URL length limit to change client behavior.
-    BaseClient.URL_CHARACTER_LIMIT = URL_CHARACTER_LIMIT
-    yield
-    # Then restore the original value.
-    BaseClient.URL_CHARACTER_LIMIT = int(URL_LIMITS.ORIGINAL)
 
 
 @pytest.mark.parametrize(
--- a/tiled/_tests/test_xarray.py
+++ b/tiled/_tests/test_xarray.py
@@ -212,18 +212,23 @@ def test_url_limit_bypass(client):
         "ORIGINAL": original,
         "TINY": 10,
     }
-    for URL_MAX_LENGTH in URL_LIMITS.values():
-        try:
-            # It should never be necessary to tune this for real-world use,
-            # but we use this knob as a way to test its operation.
-            xarray_client.URL_CHARACTER_LIMIT = URL_MAX_LENGTH
-            requests.clear()  # Empty the Request cache.
-            actual = dsc.read()
-            xarray.testing.assert_equal(actual, expected)
-            assert len(requests) == expected_requests
-        finally:
-            # Restore default.
-            xarray_client.URL_CHARACTER_LIMIT = original
+    with record_history() as history:
+        for URL_MAX_LENGTH in URL_LIMITS.values():
+            try:
+                # It should never be necessary to tune this for real-world use,
+                # but we use this knob as a way to test its operation.
+                xarray_client.URL_CHARACTER_LIMIT = URL_MAX_LENGTH
+                requests.clear()  # Empty the Request cache.
+                actual = dsc.read()
+                xarray.testing.assert_equal(actual, expected)
+                assert len(requests) == expected_requests
+            finally:
+                # Restore default.
+                xarray_client.URL_CHARACTER_LIMIT = original
+
+    request_methods = list(request.method for request in history.requests)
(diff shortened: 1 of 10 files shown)


padraic-shafer:
@danielballan Could you please give this draft PR a sanity check before this goes further?

padraic-shafer:
`URL_CHARACTER_LIMIT` or some variant is now showing up in `xarray` and in `awkward`.  It probably makes sense to pull this into a project-wide constant, possibly a server config variable.

tiled/server/router.py
```diff
@@ -604,14 +604,65 @@ async def table_full(
...
+@router.post(
+    "/container/full/{path:path}",
+    response_model=schemas.Response,
+    name="full 'container' metadata and data",
+)
+async def post_container_full(
+    request: Request,
+    entry=SecureEntry(scopes=["read:data"]),
+    principal: str = Depends(get_current_principal),
+    field: Optional[List[str]] = Body(None, min_length=1),
```
danielballan, line 639: Presumably `field` will be removed from the `POST` variant.

tiled/_tests/test_xarray.py
```diff
@@ -151,46 +151,42 @@ def test_wide_table_optimization_off(client):
...
+                requests.clear()  # Empty the Request cache.
+                actual = dsc.read()
+                xarray.testing.assert_equal(actual, expected)
+                assert len(requests) == expected_requests
+            finally:
+                # Restore default.
+                xarray_client.URL_CHARACTER_LIMIT = original
+
+    request_methods = list(request.method for request in history.requests)
+    assert "GET" in request_methods  # URL_LIMITS.HUGE
```
danielballan, line 188: Might be worth moving the `record_history()` and the `assert` _inside_ the loop over `URL_LIMITS` so that it can particularly be checked that each case worked out as expected. Debugging failures will be easier.

danielballan (commented):
Looks on track. I left small comments.

tiled/server/router.py
padraic-shafer: In the POST route, `field` is now bound to the request Body rather than a Query parameter.  I could rename it to `body` if you think that would add clarity.

Let me know if you meant something else.

tiled/_tests/test_xarray.py
padraic-shafer: Done.

tiled/_tests/test_xarray.py
```diff
@@ -151,46 +153,52 @@ def test_wide_table_optimization_off(client):
...
+    _xarray_client.URL_CHARACTER_LIMIT = URL_LIMITS.ORIGINAL
+
+
+@pytest.fixture(scope="function")
+def remaining_methods(set_and_deplete):
+    "Methods must be removed by calling test when corresponding request is made"
+    yield set_and_deplete
+
+
+@pytest.mark.parametrize("set_and_deplete", (("GET", "POST"),), indirect=True)
```
padraic-shafer, line 178: Is this perhaps overly clever (aka complicated)?

padraic-shafer: I'm removing this and saving it until a more general approach is needed.

padraic-shafer:
> Looks on track. I left small comments.

Thanks for the quick feedback!

padraic-shafer:
_UPDATE_

* Tests of POST routes now exist for `xarray` and `dataframe` clients, corresponding to `/container/full` and `/table/partition`, respectively.
* For GET `/table/partition` the `field` query parameter has been deprecated in favor of `column`
* I'm having difficulty finding a portion of the python client that calls GET `/table/full`; I only see PUT `/table/full` to write a dataframe. I'm probably overlooking something obvious.
    * I could add the POST equivalent, but the tests might be a bit forced.

There is still some cleanup to do, particularly with regard to [the repeated `URL_CHARACTER_LIMIT` definition](https://github.com/bluesky/tiled/pull/657#issuecomment-1945411022).

padraic-shafer:
At least for the `DataFrameClient.export()` that accepts `column` query parameters, it could be worth making `export_util()` POST-compatible. However, `export_util()` is used for several other clients that would not use POST, so I'm inclined to pause for now. What do other people think about this?

padraic-shafer:
_UPDATE_: Added `POST /table/full` route and tests for DataFrame

danielballan:
I think the only place `/table/full` is used is in export

https://github.com/bluesky/tiled/blob/fcf01467ecc81c019b084f4d8306629639c5af01/tiled/client/dataframe.py#L230

and it sounds like you found that. I think deferring the export refactor to another PR is fine. This at least makes it _possible_ to request the desired result, which is a meaningful step forward.

padraic-shafer:
This PR is now ready for review.

padraic-shafer:
Refactoring `<client>.export()` to use POST as well as GET will need to wait for a subsequent PR.  First, some underlying changes are needed to make `export()` more robust for containers.  A preview of those changes is in https://github.com/padraic-shafer/tiled/pull/2 .

These changes are tested with several Tiled containers:
* Directory of files
* Excel file with multiple spreadsheets
* Zarr group with multiple datasets
* HDF5 file with multiple top-level objects

jmaruland (approved):

Issue #528, opened 2023-07-21
Support streaming Bluesky data into Tiled


danielballan:
Currently, we store Bluesky as we always have, in MongoDB + external files. The Databroker package provides a Tiled Adapter specialized to read data from this storage and arrange it into Tiled’s data model for the Tiled server.

Next, we want to try writing Bluesky data directly into Tiled, using Tiled’s built-in catalog to store metadata, using tabular formats to store Event data, and using external files for data written out by detectors. This works now for uploading completed Bluesky runs. But there is no good story yet for _streaming_ data in. You would have to overwrite everything to publish an update.

What we need:

- [ ] PATCH route for appending to a data source
- [ ] Support for backing a node with more than one data source (parquet for event data, HDF5 for externally-written image data)
- [ ] Support for registering files from the client visible to the server, using a new scope that is not granted to ordinary Users by default, with an ACP to customize which paths are allowed to be registered on which nodes (based on metadata…)
- [ ] A small utility, outside of Tiled core, for consuming Bluesky documents and submitting requests to Tiled, bunching updates based on rate, similar to how IOCs work under low and high load

PR #659, opened 2024-02-20
Add structure_family to data_sources table.

danielballan:
This adds `structure_family` to the `data_sources` table, in support of #528.

Migrations tested on the SQLite and PostgreSQL [example data](https://github.com/bluesky/tiled-example-database).


padraic-shafer (approved):
Looks good. I'll leave timing of the merge to you, in case it needs to be synch'ed with anything else.

PR #660, opened 2024-02-20
Export directories (and other container files) by enabling support for `/containers/full` route

padraic-shafer:
**TL;DR**:  Make `export()` more robust for containers.

---
Calling `client.Container.export()` fails when exporting to the default serialization format, "application/x-hdf5".

This PR fixes that functionality with the following updates:

* `CatalogContainerAdapter.items_range()` now supports slicing for item generators (in addition to previously supported subscriptable items).
* `serialize_hdf5()` now support both async and sync reads, including `CatalogArrayAdapter.read()`.

Additionally exporting selected "fields" now works correctly, using `client.Container.export(..., fields=...)` or `GET /container/full?field=...`:

* `/container/full` routes now pass requested fields to `CatalogArrayAdapter.read()` as originally intended.
* `ensure_awaitable()` now correctly handles `**kwargs` for synchronous callables.

---
These changes are tested with several Tiled containers:
* Directory of files
* Excel file with multiple spreadsheets
* Zarr group with multiple datasets
* HDF5 file with multiple top-level objects

### DIFF
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -1,5 +1,6 @@
 import collections
 import importlib
+import itertools as it
 import operator
 import os
 import re
@@ -859,9 +860,11 @@ class CatalogContainerAdapter(CatalogNodeAdapter):
 
     async def items_range(self, offset, limit):
         if self.data_sources:
-            return (await self.get_adapter()).items()[
-                offset : (offset + limit) if limit is not None else None  # noqa: E203
-            ]
+            return it.islice(
+                (await self.get_adapter()).items(),
+                offset,
+                (offset + limit) if limit is not None else None  # noqa: E203
+            )
         statement = select(orm.Node).filter(orm.Node.ancestors == self.segments)
         for condition in self.conditions:
             statement = statement.filter(condition)
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -863,7 +863,7 @@ class CatalogContainerAdapter(CatalogNodeAdapter):
             return it.islice(
                 (await self.get_adapter()).items(),
                 offset,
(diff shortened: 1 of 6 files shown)


padraic-shafer:
By adding some `xfail` tests to the suite, it became clear that the results of `test_metrics` depend very much on which tests were run before it. That test inspects **ALL** metrics generated during the entire testing session, rather than only the metrics generated during the test.

I'll push an update that isolates the metrics inspected by `test_metrics`.

padraic-shafer:
Near the beginning of `test_metrics` a baseline of the total_request_time is now computed from a `/metrics` endpoint call before the true testing begins. I did not see a simple way to clear the metrics, so this seemed like a reasonable alternative.

danielballan (approved):

jmaruland:
I was getting an error while trying to save 2 bluesky runs in on hdf5 file. On the client side, I was getting a 500 error, but I saw that the code stopped while trying to access one of the nodes here.
https://github.com/jmaruland/client_tools/blob/multiple-scan-exporter/client_exporter.py#L100
But this should help to unblock the issues affecting my code.
Thanks @padraic-shafer

tiled/catalog/adapter.py
```diff
@@ -838,9 +839,11 @@ async def update_metadata(self, metadata=None, specs=None):
...
     async def keys_range(self, offset, limit):
         if self.data_sources:
-            return (await self.get_adapter()).keys()[
-                offset : (offset + limit) if limit is not None else None  # noqa: E203
-            ]
+            return it.islice(
+                (await self.get_adapter()).keys(),
+                offset,
+                (offset + limit) if limit is not None else None,  # noqa: E203
+            )
```
jmaruland, line 846: I loved this solution.

jmaruland (approved):

PR #654, opened 2024-02-11
add howto for registering files

dylanmcreynolds:

PR #661, opened 2024-02-21
Register externally-managed files via the client

danielballan:
Tiled enables registration of externally-managed files with a catalog. In `main`, this is a server-side activity where the user interacts directly with the SQL database via Python, and thus requires the database credentials. In this PR, this becomes a client-side activity where the user interacts with a new HTTP endpoint, `POST /register/{path}` and must hold a new scope, `register`.

The module `tiled.catalog.register` becomes `tiled.client.register`.

The CLI `tiled catalog register <DATABASE_URI>` becomes `tiled register <TILED_URL> --api-key <API_KEY>`.

## Why didn't we do this in the first place?

We originally considered doing this client-side from the start. I had two concerns:

1. Security. This PR enables clients to ask the server to expose files on the server. This is naturally a security concern. However, giving the file-registering program direct access to the database is _also_ a security concern, arguably a worse one. And we have the following protections:
    - In multiple-user deployments, the required scope, `register`, is given only to Tiled admins and service accounts that they elect to grant it to.
    - The server configuration must explicitly list areas of its local storage where files may be registered, via its `readable_storage` configuration. Anything outside of that will be rejected. This is enforced at read time as well, to be sure that even with a corrupted database there is no way that Tiled would serve an arbitrary file off of the system.

2. Latency. When scanning large directories, this may be slow. But there is a path to making it faster. In a future PR, we can add `POST /nodes/metadata/{path}` and `POST /nodes/register/{path}` for _bulk_ creation of nodes, with internally- and externally-managed data respectively.

I have stacked @dylanmcreynolds commits from #654 on top of my commits here. My intent is to update and ideally simplify it before merging, but I'll need some input from @dylanmcreynolds and @Wiebke on that.


padraic-shafer:
_TODO_: After https://github.com/bluesky/tiled/pull/660 gets merged, will need to apply these analogous updates to `test_container_fields.py`.

https://github.com/bluesky/tiled/blob/053d3d2eedf3f523898e63f620773ed3def6249d/tiled/_tests/test_container_files.py#L21

dylanmcreynolds:
Excellent!  

Should we consider adding cleint-side register text to the how-to document rather than replacing it? Even with this, are there cases where we would want to support both?

danielballan:
Ideally I'd like to add whatever convenience APIs we need to reduce that chunk of Python to something pithier. I am aware of these use cases:

- Walk a directory. We have that covered.
- Register a _specific file_ manually providing details (e.g. custom specs...) instead of taking whatever the automatic discovery generates. We need to document and perhaps streamline this.

Can you go into more detail about what needs to be addressed?

dylanmcreynolds:
I think I was trying to ask a different question. My doc descrbed a labor-intensive but effective method for registering files in code that has direct access to the database, while this PR adds functionality to do the same through the client. Do we want to keep the documentation of the more involved server-side register, adding the client side as an alternative? Or do we want to abandon my documentation in favor of new client-side register text?

padraic-shafer (commented):
I like this plan.  At a high level the code changes look good so far.

dylanmcreynolds:
After talking to @danielballan and @Wiebke , I'm pretty excited about this PR. To clarify my previous understanding was incorrect, and this moves the register function from the server to the client.

danielballan:
Capturing sketch of docs from chat. To be refined/tested. Probably missing a parameter or two.

## Quickstart

```
tiled serve directory ...
```

## Production

Start a catalog

```
tiled serve catalog <DATABASE_URI> --api-key secret
```

Simple cases: Walk the directory and let Tiled use its Adapters to describe the files.

```
tiled register [--prefix /a/b/c] http://localhost:8000 --api-key secret
```

Complex cases: Manually describe the files.

```py
from tiled.client.data_source import Asset, DataSource

# POST /register/{path}
client.new(
    structure_family=StructureFamily.container,
    data_sources=DataSource(
    	assets=[Asset(data_uri="file:///...", num=1), Asset(data_uri="file:///...", num=2)],
	mimetype="multipart/related;type=image/tiff",
	structure_family=StructureFamily.container,
),
    metadata={},
    specs=["DLSflavoredNXisthemostdeliciousflavor", "NXtomo", "NXgeneric"],
)
```

tiled/server/router.py
```diff
@@ -1077,11 +1077,53 @@ async def post_metadata(
...
+        if data_source.assets:
+            raise HTTPException(
+                "Externally-managed assets cannot be registered "
+                "using POST /metadata/{path} Use POST /register/{path} instead."
+            )
     if not getattr(entry, "writable", False):
         raise HTTPException(
             status_code=405, detail=f"Data cannot be written at the path {path}"
         )
+    return await _create_node(
```
padraic-shafer, line 1090: Is `writable` actually needed here?  My understanding is that this route simply modifies the catalog, not the storage (file system).

```python3
    if not getattr(entry, "writable", False):
        raise HTTPException(
            status_code=405, detail=f"Data cannot be written at the path {path}"
        )
```

padraic-shafer: I ask because it is causing the "test_container_fields.py" tests to fail for the top-level container node (with no data sources). Whereas removing this check enables the entire Tiled test suite to pass.

For comparison, the new /register route does not include this check.

tiled/client/register.py
```diff
@@ -158,40 +158,43 @@ async def register(
...
                 key=key,
             )
-            child_catalog = await catalog.lookup_adapter([segment])
-        catalog = child_catalog
+            # TODO When we have a tiled AsyncClient, use that.
+            child_node = await anyio.to_thread.run_sync(node.get, segment)
+        node = child_node
     if path.is_dir():
         # Recursively enter the directory and any subdirectories.
         if overwrite:
```
dylanmcreynolds, line 177: I have a question about this...when I was doing this on the server side before this PR, I couldn't figure out why you would ever want to to set this to true. It doesn't effec the node you're writing, but it deletes collection parents of the node. Am i missing something? Is there a use case for that?

tiled/server/router.py
padraic-shafer: Perhaps this check is only needed if data sources are present?

```python3
    if data_sources and not getattr(entry, "writable", False):
        raise HTTPException(
            status_code=405, detail=f"Data cannot be written at the path {path}"
        )
```

padraic-shafer: I now see that this check was to ensure that internally managed assets get the backing storage initialized. Have submitted a short [PR to this branch](https://github.com/danielballan/tiled/pull/1) using the approach in my comment above.

tiled/client/register.py
danielballan: It's meant to delete the node and all its children. The use case is when you walk a directory and then walk it again later, wanting to replace whatever Tiled had from directory with new nodes representing the current (maybe changed) contents.

docs/source/how-to/register.md
```diff
@@ -0,0 +1,95 @@
...
+formats, and registers the metadata, structure, and filepaths.
+
+```
+tiled register http://localhost:8000 [--api-key <SECRET>] <DIRECTORY>
+```
+
+### Complex cases
+
+Sometimes it is necessary to take more manual control of this registration
+process. Using the Python client,
```
dylanmcreynolds, line 72: Can we add sentence here?

```
Sometimes it is necessary to take more manual control of this registration
process. For example, if you want to add a `spec` or custom `metadata` to the data set, you will want to do that programmatically usign the Tiled client.

**Review by dylanmcreynolds — APPROVED** [review]
Great

**Review by dylanmcreynolds — APPROVED** [review]
Great

File: docs/source/how-to/register.md [hunk 4]
```diff
@@ -0,0 +1,98 @@
+# Register Content in Tiled
+
+The usage `tiled serve directory ...` is mostly for demos and small-scale use.
+For more sophisticated control over this process, see
+
+## Quickstart
+
+The tutorial {doc}`../how-to/register` demonstrates the usage:
```
padraic-shafer, line 8: There's some strange self-referencing here...probably from merging earlier edits,

docs/source/how-to/register.md
```diff
@@ -0,0 +1,98 @@
...
+
+```
+tiled serve directory [--watch] [--public] [--api-key <SECRET>] <DIRECTORY>
+```
+
+which is a shorthand for:
+
+1. Walk a directory tree, identifying formats it recognizes, ingesting their
```
padraic-shafer, line 16: I find this easier to read...

```suggestion
1. Walk a directory tree to identify formats it recognizes and then ingest their
```

docs/source/how-to/register.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Limitations of `tiled serve directory ...`
+
+The shorthand is great for quickly getting started, but it has numerous
+limitations.
+
+- Tiled walks the entire directory at server startup. This can be slow.
+- Tiled creates an ephemeral database (SQLite in a temporary directory)
+  just for this process. That work is discarded when the server shuts down.
+- One database per server not horizontally scalable.
```
padraic-shafer, line 31:
```suggestion
- One database per server is not horizontally scalable.
```

docs/source/how-to/register.md
danielballan: I missed this specific suggest but I'm glad I capture the gist of it.

docs/source/how-to/register.md
danielballan: fixed

padraic-shafer, line 8:
```suggestion
The following guide demonstrate more sophisticated control over this process.

## Quickstart

```

padraic-shafer (approved):
This looks great.  One small change noted in the comment.

danielballan:
....Plus one last commit for `s/demonstrate/demonstrates/`.

padraic-shafer (approved):
Thanks for updating the docs!

Wiebke (commented):
I am late to the party, but this is great!

padraic-shafer:
@Wiebke Thanks for adding your support!  The stacked PRs were starting to pile up, so with one approval from ALS + another from NSLS-II, I made a call that we were ok to merge this without further review.  I'll aim to not make a habit of that though. :)

PR #662, opened 2024-02-21
Thread-safety for the client Cache()

canismarko:
Based on a mattermost conversation. I wanted to use the Cache() from multiple threads so I could make an awaitable wrapper around the Tiled client. 

sqlite3 does support thread-safety if it's built into the underlying sqlite library. So this PR has the cache check if the sqlite3 library supports thread-safety, and if it does then allows writes from other threads.

https://docs.python.org/3/library/sqlite3.html#sqlite3.threadsafety

Even still, I'm not sure how parallel writes get handled, so I also included a thread lock forcing parallel writes to wait until previous writes are done. Might be overkill but I don't have a way to check this easily so figured I'd err on the side of caution.

### DIFF
--- a/tiled/_tests/test_client_cache.py
+++ b/tiled/_tests/test_client_cache.py
@@ -11,7 +11,7 @@ import pytest
 from ..adapters.array import ArrayAdapter
 from ..adapters.mapping import MapAdapter
 from ..client import Context, from_context, record_history
-from ..client.cache import Cache, CachedResponse, with_thread_lock
+from ..client.cache import Cache, CachedResponse, ThreadingMode, with_thread_lock
 from ..server.app import build_app
 
 tree = MapAdapter(
@@ -177,7 +177,7 @@ def test_clear_cache(client):
 
 def test_not_thread_safe(client, monkeypatch):
     # Check that writes fail if thread safety is disabled
-    monkeypatch.setattr(sqlite3, "threadsafety", 0)
+    monkeypatch.setattr(sqlite3, "threadsafety", ThreadingMode.SINGLE_THREAD)
     cache = client.context.cache
     # Clear the cache in another thread
     with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
@@ -187,7 +187,8 @@ def test_not_thread_safe(client, monkeypatch):
 
 
 @pytest.mark.skipif(
-    sqlite3.threadsafety < 2, reason="sqlite not built with thread safe support"
+    sqlite3.threadsafety != ThreadingMode.SERIALIZED,
+    reason="sqlite not built with thread safe support",
 )
 def test_thread_safety(client):
     cache = client.context.cache
--- a/tiled/_tests/test_client_cache.py
+++ b/tiled/_tests/test_client_cache.py
@@ -1,9 +1,9 @@
-import time
-import sqlite3
-from contextlib import closing
-import concurrent.futures
 import asyncio
+import concurrent.futures
+import sqlite3
 import threading
+import time
+from contextlib import closing
 
 import numpy
 import pytest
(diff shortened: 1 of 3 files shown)


tiled/client/cache.py
```diff
@@ -171,17 +187,23 @@ def __init__(
...
+        underlying sqlite library was built with thread safety
+        enabled. Even still, it may be a good idea to use a thread
+        lock (``@with_thread_lock``) to prevent parallel writes.
 
-        SQLite is not threadsafe for concurrent _writes_.
         """
-        return threading.current_thread().ident == self._owner_thread
+        is_main_thread = threading.current_thread().ident == self._owner_thread
+        sqlite_is_safe = sqlite3.threadsafety > 2
+        return is_main_thread or sqlite_is_safe
```
padraic-shafer, line 206: I'm not sure there is any guarantee about ordinality of the `threadsafety` value in `sqlite3`. It's probably safer and arguably clearer to use a named constant here.

```suggestion
        SERIALIZED = 3  # Could be defined in an enum elsewhere
        is_main_thread = threading.current_thread().ident == self._owner_thread
        sqlite_is_safe = sqlite3.threadsafety == SERIALIZED
        return is_main_thread or sqlite_is_safe
```

* https://docs.python.org/3/library/sqlite3.html#sqlite3.threadsafety
* https://github.com/python/cpython/blob/347acded845d07b70232cade8d576b6f0aaeb473/Modules/_sqlite/module.c#L521-L541

tiled/_tests/test_client_cache.py
```diff
@@ -169,3 +173,52 @@ def test_clear_cache(client):
     assert cache.count() > 0
     cache.clear()
     assert cache.size() == cache.count() == 0
+
```
padraic-shafer, line 176: See comment regarding `sqlite3.threadsafety` and comparing to a named constant/enum.

padraic-shafer (changes requested):
Thanks for adding this functionality.  I've added a simple suggestion to use enumerated constant values to refer to the magic numbers of `sqlite3.threadsafety`.

canismarko:
Thanks @padraic-shafer, makes sense. I went with a full IntEnum since we're using those values in at least two places.

padraic-shafer (approved):
👍
Thanks for adding those updates!  This looks great.

danielballan:
This demonstrates that, with this PR, we will cache (small) data responses. In `main`, we cannot because the request is made from a dask worker running on a different thread than the `Cache` and its SQLite connection were instantiated on.

```py
from tiled.client.cache import Cache

cache = Cache("main.db")
c = from_profile('demo', cache=cache)
c['generated']['tiny_image'][:]
```

```
$ sqlite3 main.db 'select cache_key, size from responses;'
https://tiled-demo.blueskyproject.io/api/v1/metadata/?include_data_sources=false|361
https://tiled-demo.blueskyproject.io/api/v1/metadata/generated?include_data_sources=false|397
https://tiled-demo.blueskyproject.io/api/v1/metadata/generated/tiny_image?include_data_sources=false|517
```

Now with this PR branch and the cache filename `pr_662.db`, notice the `/array/block/...` response cached.

```
$ sqlite3 pr_662.db 'select cache_key, size from responses;'
https://tiled-demo.blueskyproject.io/api/v1/metadata/?include_data_sources=false|361
https://tiled-demo.blueskyproject.io/api/v1/metadata/generated?include_data_sources=false|397
https://tiled-demo.blueskyproject.io/api/v1/metadata/generated/tiny_image?include_data_sources=false|517
https://tiled-demo.blueskyproject.io/api/v1/array/block/generated/tiny_image?block=0%2C0&expected_shape=50%2C50|20000
```

danielballan (approved):

danielballan (changes requested):
Just before merging, I noticed a subtle issue. I have opened a PR into this PR branch for your review @canismarko https://github.com/spc-group/tiled/pull/1

danielballan:
Awesome! Thanks again @canismarko.

PR #664, opened 2024-02-22
Support service principal in `SimpleAccessPolicy`

danielballan:
This makes it possible to add a service principal's uuid (as it has no identity/username...) to a `SimpleAccessPolicy`. Test included.

In support of C2QA / @jmaruland.


danielballan:
Commit [c58f7fb](https://github.com/bluesky/tiled/pull/664/commits/c58f7fbeda4c4aeaf945a841d20030a7fb6fc87c) is not really related, but it made it easier to test, and it's something we've been wanting.

PR #667, opened 2024-02-23
Catch (and test for) a warning about deprecated 'field'

danielballan:


tiled/_tests/test_dataframe.py
```diff
@@ -193,10 +193,13 @@ def test_redundant_query_parameters(context):
...
-    for param in ("field", "column"):
-        params = original_params.copy()
-        params.pop(param)
+    params = original_params.copy()
+    params.pop("column")
+    with pytest.warns(DeprecationWarning):
         context.http_client.get(url_path, params=params).raise_for_status()
+    params = original_params.copy()
+    params.pop("field")
+    context.http_client.get(url_path, params=params).raise_for_status()
```
padraic-shafer, line 202: Because `test_deprecated_query_parameter()` is now effectively folded into `test_redundant_query_parameters()`, should we remove `test_deprecated_query_parameter()`?

Or if the intent here is simply to silence this warning during the test session, should we instead use `with warnings.catch_warnings(category=DeprecationWarning):` here?

I think either could work, but I'd like to understand the intended purpose.

padraic-shafer: ...or better yet `@pytest.mark.filterwarnings("ignore::DeprecationWarning")` for the latter case.

danielballan: Sorry I was focused on one "tree" that produced a warning in pytest output, and I failed to pause to take a view of the whole forest. Filtering the warning feels best.

padraic-shafer (approved):
DeprecationWarning is now suppressed when not directly being tested.

PR #677, opened 2024-03-04
Replace `numpy.product()` with `.prod()`

padraic-shafer:
This PR replaces the deprecated `numpy.product()` calls with equivalent calls to `numpy.prod()`, which is compatible with numpy 2.0.

### DIFF
--- a/tiled/client/array.py
+++ b/tiled/client/array.py
@@ -33,9 +33,8 @@ class _DaskArrayClient(BaseClient):
     @property
     def nbytes(self):
         structure = self.structure()
-        return (
-            numpy.prod(structure.shape) * structure.data_type.to_numpy_dtype().itemsize
-        )
+        itemsize = structure.data_type.to_numpy_dtype().itemsize
+        return numpy.prod(structure.shape) * itemsize
 
     @property
     def chunks(self):
--- a/tiled/client/array.py
+++ b/tiled/client/array.py
@@ -24,7 +24,7 @@ class _DaskArrayClient(BaseClient):
 
     @property
     def size(self):
-        return numpy.product(self.structure().shape)
+        return numpy.prod(self.structure().shape)
 
     @property
     def dtype(self):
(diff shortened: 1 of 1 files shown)


jmaruland (commented):
I think it would be important to define this last version of `numpy` as the minimum requirement in `pyproject.toml`, just to make sure that the right version gets installed because it looks like this is not backwards compatible.
I mean, changing every line where 'numpy' is defined in `pyproject.toml` to something like `'numpy >=2.0.0b1'`

danielballan:
It is possible that the new approach here works in both 1.x and 2.x, whereas the old one works only in 1.x. Can you verify?

Specify a minimum numpy version (and testing against it) is a good idea in any case, but we should continue to support 1.x for awhile yet, per [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

padraic-shafer:
"New in version 1.7.0"
https://numpy.org/doc/stable/reference/generated/numpy.prod.html#numpy.prod

jmaruland:
I can confirm that it works with the current requirements in tiled

```
(tiled-dev) C:\Users\jmaruland>conda list | grep numpy
msgpack-numpy             0.4.8                    pypi_0    pypi
numpy                     1.26.2                   pypi_0    pypi
numpydoc                  1.6.0                    pypi_0    pypi

(tiled-dev) C:\Users\jmaruland>ipython
Python 3.11.5 | packaged by Anaconda, Inc. | (main, Sep 11 2023, 13:26:23) [MSC v.1916 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.15.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: arr = np.ones([4,3])

In [3]: arr.shape
Out[3]: (4, 3)

In [4]: np.product(arr.shape)
Out[4]: 12

In [5]: np.prod(arr.shape)
Out[5]: 12
```

jmaruland (approved):

PR #679, opened 2024-03-04
Implement appendable tabular data

skarakuzu:
work in progress implementing append data to file

### DIFF
--- a/tiled/_tests/test_writing.py
+++ b/tiled/_tests/test_writing.py
@@ -15,7 +15,9 @@ import pytest
 import sparse
 
 from ..catalog import in_memory
+from ..catalog.adapter import CatalogContainerAdapter
 from ..client import Context, from_context, record_history
+from ..client.dataframe import DataFrameClient
 from ..mimetypes import PARQUET_MIMETYPE
 from ..queries import Key
 from ..server.app import build_app
@@ -25,7 +27,6 @@ from ..structures.sparse import COOStructure
 from ..structures.table import TableStructure
 from ..validation_registration import ValidationRegistry
 from .utils import fail_with_status_code
-from pandas.testing import assert_frame_equal
 
 validation_registry = ValidationRegistry()
 validation_registry.register("SomeSpec", lambda *args, **kwargs: None)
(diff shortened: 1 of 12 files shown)


danielballan:
How can we make the `CSVAdapter` work with the `register_single_item` code?

https://github.com/bluesky/tiled/blob/acfa2a29ceed2d78dcb3881fa13406b1f9d56191/tiled/client/register.py#L296-L326

This code typically operates with adapters that expect a single parameter `data_uri` that takes one path. But we want `CSVAdapter` to work, in the general case, for `data_uris`, a _list_ of partitioned files. How can we manage that?

Notice that in the code above, we construct an instance of the `adapter`

```py
adapter = await anyio.to_thread.run_sync(adapter_factory, ensure_uri(item))
```

and then we ask `adapter` to describe aspects of this data.... `adapter.structure_family`, `adapter.metadata()`, `adapter.structure()`, and so on. I think we should extend this same pattern: we should ask the adapter to write the whole `data_sources` list for us, since it knows the specific parameters it expects. Thus, we should change:

```py
    return await create_node_or_drop_collision(
        ...,
        data_sources=[
            DataSource(
                structure_family=adapter.structure_family,
                mimetype=mimetype,
                structure=dict_or_none(adapter.structure()),
                parameters={},
                management=Management.external,
                assets=[
                    Asset(
                        data_uri=ensure_uri(item),
                        is_directory=is_directory,
                        parameter="data_uri",
                    )
                ],
            )
        ],
    )
```

to

```py
    if hasattr(adapter, "generate_data_sources"):
        # Let the Adapter describe the DataSouce(s).
        data_sources = adapter.generate_data_sources()
    else:
        # Back-compat: Assume one Asset passed as a
        # parameter named 'data_uri'.
        data_source = [
            DataSource(
                structure_family=adapter.structure_family,
                mimetype=mimetype,
                structure=dict_or_none(adapter.structure()),
                parameters={},
                management=Management.external,
                assets=[
                    Asset(
                        data_uri=ensure_uri(item),
                        is_directory=is_directory,
                        parameter="data_uri",
                    )
                ],
            )
        ],
    return await create_node_or_drop_collision(
        ...,
        data_sources=data_sources,
    )
```

Then, a new method `CSVAdapter.generate_data_sources()` can be defined, which returns:
```py
[
            DataSource(
                structure_family=adapter.structure_family,
                mimetype=mimetype,
                structure=dict_or_none(adapter.structure()),
                parameters={},
                management=Management.external,
                assets=[
                    Asset(
                        data_uri=ensure_uri(item),
                        is_directory=is_directory,
                        parameter="data_uris",  # <-- PLURAL!
                        num=0  # <-- denoting that the Adapter expects a list, and this is the first element
                    )
                ],
            )
        ]
```

padraic-shafer:
Naive question:  Why does `CsvAdapter` use multiple data sources?

Does the strategy include writing appended rows to a new file? Or is this "chunking" of columns?  or...?

danielballan:
This is chunking of partitions:

```
# partition-0.csv
1, 10.0, a
2, 20.0 b
3, 30.0, c
```

```
# partition-1.csv
4, 4.0, d
5, 5.0, e
6, 6.0, f
```

When Tiled _wrties_ CSVs, it adopts this strategy for scability. But when CSV _registers_ CSVs may encounter partitioned or non-partitioned files and should be able to manage both, potentially.

Appending extends a specific partition, i.e. adds rows for an existing file. New partitions may also be created as needed.

danielballan:
Rebased on #686. The second force-push was to retain @skarakuzu's authorship is retained in the commit history.

tiled/_tests/test_writing.py
```diff
@@ -451,3 +453,68 @@ async def test_container_export(tree):
...
+            key="x",
+        )
+        x.write(df)
+
+        df2 = pandas.DataFrame({"A": [11, 12, 13], "B": [14, 15, 16]})
+
+        x.append_partition(df2, 0)
+        df3 = pandas.DataFrame({"A": [1, 2, 3, 11, 12, 13], "B": [4, 5, 6, 14, 15, 16]})
+
+        assert [row for col in x.columns for row in x[col]] == [
```
danielballan, line 518: Happily, pandas has a utility for this: [`pandas.testing.assert_frame_equal`](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html).

skarakuzu: I tried it but the `assert_frame_equal` functions expects the two data types be the same. The `df3` is of panda type but the `x` is `tiled.client.dataframe.DataFrameClient` so the function raises error.

danielballan: I see. You can call `.read()` on the `DataFrameClient` to get a plain old dataframe out.

skarakuzu: `.read()` worked, thanks!

Issue #675, opened 2024-03-04
Updates needed to remove `DeprecationWarning`s


padraic-shafer:
Some housekeeping is needed to respond to multiple `DeprecationWarning`s that are raised during testing.
* DeprecationWarning: on_event is deprecated, use lifespan event handlers instead.
    * Read more about it in the [FastAPI docs for Lifespan Events](https://fastapi.tiangolo.com/advanced/events/).
    * I will submit a PR for this.
* DeprecationWarning: `product` is deprecated as of NumPy 1.25.0, and will be removed in NumPy 2.0. Please use `prod` instead.
    * I will submit a PR for this.
* DataFrame._data is deprecated and will be removed in a future version. Use public APIs instead.
    * This comes from the `cachey` dependency. It will likely require an upstream PR.

padraic-shafer:
Submitted PR https://github.com/bluesky/tiled/pull/676 for the FastAPI lifespan events

padraic-shafer:
Submitted PR https://github.com/bluesky/tiled/pull/677 to use `numpy.prod()` rather than `numpy.product()`.

padraic-shafer:
> * DataFrame._data is deprecated and will be removed in a future version. Use public APIs instead.
>   * This comes from the `cachey` dependency. It will likely require an upstream PR.

It looks like Pandas wants to remove its `BlockManager`. The roadmap is not yet clear on what the replacement will look like, and there is not suggested alternate for accessing the blocks.  This access has remained stable for several years, and discussions suggest there is no hurry to change that.

So our best option here is probably to just ignore this warning.  We could specifically filter it with action: `ignore:DataFrame._data is deprecated and will be removed in a future version. Use public APIs instead.:DeprecationWarning`.

## References
* ..
* https://github.com/pandas-dev/pandas/issues/17302

padraic-shafer:
Submitted PR https://github.com/bluesky/tiled/pull/682 to ignore the warnings from `cachey.nbytes()`.

danielballan:
Maybe we can solve the cachey problem by simply [dropping the dependency](https://github.com/bluesky/tiled/issues/683).

PR #682, opened 2024-03-07
Ignore DeprecationWarning from Pandas via cachey

padraic-shafer:
Calls to `cachey.nbytes()` issue a `DeprecationWarning` because of direct access to the deprecated internals of Pandas.
* https://github.com/bluesky/tiled/issues/675#issuecomment-1983581882

This PR catches those warnings at the corresponding calls in Tiled's `object_cache` module. It also shows where this could be limited to just the tests instead.

I favor doing this in the code rather than the tests, but it does add some uncertainty for thread safety.

### DIFF
--- a/tiled/_tests/test_writing.py
+++ b/tiled/_tests/test_writing.py
@@ -25,6 +25,10 @@ from ..structures.sparse import COOStructure
 from ..validation_registration import ValidationRegistry
 from .utils import fail_with_status_code
 
+WARNING_PANDAS_BLOCKS = (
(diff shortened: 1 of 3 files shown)


genematx (approved):
Should we also fix other deprecation warnings, e.g. "DeprecationWarning: `product` is deprecated as of NumPy 1.25.0, and will be removed in NumPy 2.0. Please use `prod` instead."?

Haha, never mind, just saw PR #677 😄

PR #687, opened 2024-03-11
Support registering internal groups/dataset in HDF5.

danielballan:
A wrapper function takes the same parameters as `HDF5Adapter` but additionally exposes an optional `path`, a list of strings navigating to a group or dataset nested inside an HDF5 file.

### DIFF
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -685,20 +685,47 @@ class CatalogNodeAdapter:
                 self.context, refreshed_node, access_policy=self.access_policy
             )
 
-    async def patch_data_source(self, data_source):
-        values = dict(
-            structure_family=data_source.structure_family,
-            mimetype=data_source.mimetype,
-            management=data_source.management,
-            parameters=data_source.parameters,
-            structure_id=structure_id,
+    async def put_data_source(self, data_source):
+        # Obtain and hash the canonical (RFC 8785) representation of
+        # the JSON structure.
+        structure = _prepare_structure(
+            data_source.structure_family, data_source.structure
         )
+        structure_id = compute_structure_id(structure)
+        # The only way to do "insert if does not exist" i.e. ON CONFLICT
+        # is to invoke dialect-specific insert.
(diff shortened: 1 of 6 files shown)


danielballan:
Demo of registering a DataSource with shape `(1, 5, 5)` and then updating that shape to be `(2, 5, 5)` via `PUT /data_source/{path}?data_source={data_source_id}`

```py
from tiled.structures.data_source import Asset, DataSource, Management
from tiled.structures.array import ArrayStructure, BuiltinDtype
import numpy

arr = numpy.ones((1, 5, 5))
structure = ArrayStructure.from_array(arr)
data_source = DataSource(management='external', mimetype='application/x-hdf5', structure_family='array', structure=structure, assets=[Asset(data_uri='file://localhost/tmp/test.h5', is_directory=False, parameter='data_uri')])
ac = c.new(structure_family='array', data_sources=[data_source])
url = ac.uri.replace('/metadata/', '/data_source/')
ac.refresh()
ds = ac.data_sources()[0]
ds['structure']['shape'] = [2, 5, 5]
ds['structure']['chunks'] = [[dim] for dim in [2, 5, 5]]
ac.context.http_client.put(url, json={'data_source': ds}, params={'data_source': 1})
```

tiled/catalog/adapter.py
```diff
@@ -658,22 +658,19 @@ async def create_node(
...
-                    try:
-                        # Find an asset if it exists
-                        statement = select(orm.Asset.id).where(
-                            orm.Asset.data_uri == asset.data_uri
+                    # Find an asset_id if it exists, otherwise create a new one
+                    statement = select(orm.Asset.id).where(
+                        orm.Asset.data_uri == asset.data_uri
+                    )
+                    result = await db.execute(statement)
+                    if result := result.fetchone():
```
danielballan, line 666: Great use the "walrus operator". Minor nit: can we use a new variable name (maybe `row`?) instead of shadowing `result`?

genematx: done!

genematx (approved):

PR #688, opened 2024-03-11
Add support for patching, merging, replacing metadata

hyperrealist:
* patch_metadata uses a http patch request with application/json-patch+json content type
* merge_metadata can be triggered with a http patch request and application/merge-patch+json
* update_metadata uses a similar approach to merge_metadata, but constructs a json-patch on the client-side
* add jsonpatch and json-merge-patch to requirements


tiled/client/metadata_update.py
```diff
@@ -0,0 +1,60 @@
+from collections import OrderedDict
+
+# adapted from https://github.com/OpenDataServices/json-merge-patch
+
+
+class DELETE_KEY:
```
danielballan, line 6: A pattern that we use elsewhere in the codebase could be applied here: https://github.com/bluesky/tiled/blob/3b07e5f5add4d5ad2c5096a3694bee63d1a7308a/tiled/utils.py#L398-L415

tiled/server/router.py
```diff
@@ -1352,6 +1354,76 @@ async def put_awkward_full(
...
+
+    structure_family, structure, specs = (
+        entry.structure_family,
+        entry.structure(),
+        body.specs if body.specs is not None else entry.specs,
+    )
+
+    metadata_modified = False
+
+    # Specs should be ordered from most specific/constrained to least.
```
danielballan, line 1392: Can this block be factored out into a utility function instead of duplicated?

tiled/client/base.py
```diff
@@ -192,6 +196,12 @@ def metadata(self):
         # persistent.
         return DictView(self._item["attributes"]["metadata"])
 
+    def metadata_copy(self):
```
danielballan, line 199: While we consider whether or not to keep `DictView` I'd suggest we hold off on adding this just yet. In the mean time we can recommend `copy.deepcopy(self.metadata)` in docstrings.

If we decide to keep `DictView`, we can add this in a follow-up PR. But once added it's harder to remove it.

danielballan (commented):
Looks good. I think the main thing that's missing is implementing the new adapters methods on the SQL-backed adapter in Tiled itself. https://github.com/bluesky/tiled/blob/3b07e5f5add4d5ad2c5096a3694bee63d1a7308a/tiled/catalog/adapter.py#L814

In a later PR, we could even look into pushing the patch representation all the way into SQL, but to start I think applying the path in Python and doing a full replace in SQL is the way to go.

tiled/client/base.py
danielballan: Thinking about this more, and speaking with @Kezzsim, I'm appreciating that there isn't really a way around this. I've come around on keeping this method.

tiled/server/router.py
hyperrealist: Will do. I think that is a sensible approach.

tiled/client/metadata_update.py
hyperrealist: Ah.. I have seen `Sentinel` before but totally forgot of its existence! Will use that instead. But perhaps it still makes sense to import `DELETE_KEY` from `tiled.client.metadata_update`? So I'll create `DELETE_KEY` there using `Sentinel`

Kezzsim (commented):
Looked at this earlier on a call with Dan, not seeing anything too out of the ordinary that he hasn't already touched on.

hyperrealist:
@danielballan, @Kezzsim this PR is ready for another review. Changes since your last review are minimal, though it took me an obscene amount of time to figure this out.

tiled/server/schemas.py
```diff
@@ -444,4 +445,42 @@ def specs_uniqueness_validator(cls, v):
...
+)
+
+
+class PatchMetadataRequest(pydantic.BaseModel):
+    # These fields are optional because None means "no changes; do not update".
+    patch: Optional[Union[List[JSONPatchSpec], Dict]]  # Dict for merge-patch
+    specs: Optional[Specs]
+
+    # Wait for fix https://github.com/pydantic/pydantic/issues/3957
+    # to do this with `unique_items` parameters to `pydantic.constr`.
```
genematx, line 467: Looks like they've fixed it already in ver 1.10.12 (we use 1.10.13). Would it be worthwhile just uncommenting line 86?

genematx: or maybe in a separate PR?

tiled/client/base.py
```diff
@@ -370,6 +380,99 @@ def update_metadata(self, metadata=None, specs=None):
...
+        ).json()
+
+        if patch is not None:
+            if "metadata" in content:
+                # Metadata was accepted and modified by the specs validator on the server side.
+                # It is updated locally using the new version.
+                self._item["attributes"]["metadata"] = content["metadata"]
+            else:
+                # Metadata was accepted as it is by the server.
+                # It is updated locally with the version submitted buy the client.
```
danielballan, line 462:
```suggestion
                # It is updated locally with the version submitted by the client.
```

tiled/server/router.py
```diff
@@ -1374,6 +1356,57 @@ async def put_awkward_full(
...
+):
+    if not hasattr(entry, "replace_metadata"):
+        raise HTTPException(
+            status_code=HTTP_405_METHOD_NOT_ALLOWED,
+            detail="This node does not support update of metadata.",
+        )
+
+    if request.headers["content-type"] == "application/json-patch+json":
+        metadata = apply_json_patch(
+            entry.metadata(), body.patch if body.patch is not None else []
```
danielballan, line 1375: This is a nice pattern when an empty collection (`[]` in this case) or `None` should both lead to the same result.

```suggestion
            entry.metadata(), (body.patch or [])
```

tiled/server/router.py
```diff
@@ -1374,6 +1356,57 @@ async def put_awkward_full(
...
+        )
+    elif request.headers["content-type"] == "application/merge-patch+json":
+        metadata = apply_merge_patch(
+            entry.metadata(), body.patch if body.patch is not None else {}
```
danielballan, line 1379: Same as above

```suggestion
            entry.metadata(), (body.patch or {})
```

tiled/client/base.py
```diff
@@ -370,6 +380,99 @@ def update_metadata(self, metadata=None, specs=None):
 
         This is subject to change or removal without notice
 
+        Parameters
+        ----------
+        metadata : dict, optional
+            User metadata. May be nested. Must contain only basic types
+            (e.g. numbers, strings, lists, dicts) that are JSON-serializable.
+            Hint:
```
danielballan, line 388: Consider replacing this explanation with an example:

```
Examples
--------

Make a mutable copy of the metadata, for editing.

>>> md = x.metadata_copy()

Add, remove, or change contents, e.g.

>>> md['a'] = 1
>>> del md['b']

Update the copy on the server.

>>> x.update_metadata(metadata)
```

Easier to quickly parse than an explanation.

tiled/client/base.py
```diff
@@ -370,6 +380,99 @@ def update_metadata(self, metadata=None, specs=None):
...
+                You only need to specify the relative paths in the nested
+                structure. `metadata_copy()` is only meant as an assistance.
+                `update_metadata()` implements a similar approach to json merge
+                patch (RFC 7386 https://datatracker.ietf.org/doc/html/rfc7386).
+                However, None (or null) does not take a special meaning,
+                although `tiled.client.metadata_update.DELETE_KEY` does!
+        specs : List[str], optional
+            List of names that are used to label that the data and/or metadata
+            conform to some named standard specification.
+
```
danielballan, line 401: Consider adding:

```
See Also
--------
patch_metadata
replace_metadata
```

And likewise on the other methods, to mutually cross-reference

tiled/client/base.py
```diff
@@ -370,6 +380,99 @@ def update_metadata(self, metadata=None, specs=None):
...
 
+        Parameters
+        ----------
+        metadata : dict, optional
+            User metadata. May be nested. Must contain only basic types
+            (e.g. numbers, strings, lists, dicts) that are JSON-serializable.
+            Hint:
+                You can use `metadata_copy()` to retrieve a mutable copy of
+                the current metadata that can be passed here with modifications.
+            Note:
```
danielballan, line 391: The numpydoc standard supports a `Note` section. I'd suggest moving this out to keep the `Parameters` section streamlined.

I don't quite understand what this first sentence means. Maybe in the Note, explain, "This compares the input to the current metadata and constructs a JSON patch," and add additional example(s) to the Example section to illustrate `DELETE_KEY` and  this statement about relative paths.

tiled/client/base.py
```diff
@@ -370,6 +380,99 @@ def update_metadata(self, metadata=None, specs=None):
...
+
+        if specs is not None:
+            self._item["attributes"]["specs"] = normalized_specs
+
+    def replace_metadata(self, metadata=None, specs=None):
+        """
+        EXPERIMENTAL: Replace metadata.
+
+        This is subject to change or removal without notice
+
```
genematx, line 475: Do we need to make any changes to `metadata_revisions`, here or elsewhere? Or is it unrelated?

tiled/client/base.py
```diff
@@ -370,6 +380,99 @@ def update_metadata(self, metadata=None, specs=None):
 
         This is subject to change or removal without notice
 
+        Parameters
```
genematx, line 383: Would it be useful to expand the main description in the docstring a little bit. I guess, it's fairly obvious what the function is doing just from it's name, so maybe just emphasising that the logic here is the same as in the usual `old_dict.update(new_dict)`, i.e. some values would be overwritten or added, but none of them would be removed, whereas `replace_metadata` can lead to some key-value pairs being dropped. (At least this is my understanding of how it works; am I right?)

The distinction between `update_` and `patch_` metadata is less obvious, I guess.

tiled/client/base.py
hyperrealist: Good point. When you call `update_metadata` it creates a revision before updating the metadata/specs. https://github.com/bluesky/tiled/blob/3b07e5f5add4d5ad2c5096a3694bee63d1a7308a/tiled/catalog/adapter.py#L836-L844

`metadata_revisions` always queries latest revisions from the server. So I think we should be fine there.

tiled/client/base.py
hyperrealist: "relative paths" was a bad word choice on my part.  I will add examples to demonstrate what I actually mean.

tiled/client/base.py
hyperrealist: I agree. I'll slightly expand the main description and add more details where appropriate.

hyperrealist:
I pushed a couple of commits addressing suggestions from @danielballan and @genematx

tiled/client/base.py
```diff
@@ -364,11 +374,139 @@ def formats(self):
...
+        if metadata is None:
+            metadata = {}
+
+        patch = jsonpatch.JsonPatch.from_diff(
+            dict(self.metadata),
+            apply_update_patch(self.metadata_copy(), metadata),
+            dumps=orjson.dumps,
+        ).patch
+
+        if not exec:
```
danielballan, line 440: I like the addition of a way to inspect the patch before applying it. I'd like to explore other ways to implement that, because this violates the principle that we generally try to follow: an optional parameter should not alter the return type.

I can think of two ways to achieve this:

1. Pull this out into a convenience function, `build_patch`.
2. Make it so that `update_metadata` returns the patch unconditionally, and have `exec` (or perhaps the inverse, `dry_run`) control whether or not it is applied, not what is returned. This might cut against the "user friendly" purpose though, as the user looking at the return value might wonder, "What am I supposed to _do_ with this? Did it work?" Perhaps this points toward (1).

danielballan (commented):
One note in line on the latest changes.

Also:
- @skarakuzu recently added a [CHANGELOG](https://github.com/bluesky/tiled/blob/main/CHANGELOG.md). Can you add a line item for this?
- Can these methods be added to the [API reference](https://github.com/bluesky/tiled/blob/c2f4e606309cb937336181c1fc9212b597c8d6a4/docs/source/reference/python-client.md?plain=1#L139)? It looks like `update_metadata` was never added, and the new methods should be added too.

tiled/client/base.py
hyperrealist: Thanks @danielballan, as is always this is another very helpful comment. I "knew" I was violating some design principle, but couldn't figure which one. I am leaning toward the convenience function approach for the same reason you mention in 2.

danielballan (approved):

danielballan:
I’ve asked @Kezzsim to test drive this locally a bit before we merge. Merge away when you are satisfied, Kari!

genematx (approved):

tiled/server/schemas.py
hyperrealist: I marked this one as resolved by mistake. This came from a part of code I copied from @danielballan. Not exactly sure what needs to be done here, but I agree maybe it should be a separate PR.

https://github.com/bluesky/tiled/blame/e056aa9694375d2b034d69a0143d40845b6ee2bb/tiled/server/schemas.py#L436-L437

danielballan:
Rebased and force-pushed to resolve merge conflicts with #695

danielballan:
Tests failed due to an incompatibility with pydantic 2.x, introduced by #695. I have pushed a commit with a fix.

Kezzsim:
This looks good, however I'm noticing a break from convention when it comes to how HTTP is implemented. Pulling up swagger docs on http://127.0.0.1:8000/docs reveals that `application/json` is the only mimetype in the fastAPI dropdown.
jsonpatchproblem1
 Technically this is correct, as the body in @hyperrealist 's spec is a json object which _contains_ valid json patches, but the overall request body itself is not a json patch. 
 
 Example of current request body [dict]:
 ```json
 {
  "patch": [
    {
      "op": "replace",
      "path": "/test",
      "value": "testing"
    }
  ],
  "specs": null
}
 ```
 Proposed json patch implementation of body [list]:
 ([conforming to RFC6902](https://datatracker.ietf.org/doc/html/rfc6902))
 ```json-patch+json
 [
    {
      "op": "replace",
      "path": "metadata/test",
      "value": "testing"
    }
  ]
 ```
 Fixing this to have it be within spec would require one of two potential changes:
 1) Split the endpoint into `PATCH /api/v1/metadata/{path}` and `PATCH /api/v1/specs/{path}` which both take valid json-patch objects that alter their respective properties at the root path.
 2) (As seen in the example above) change the body to be a valid json-patch but make the RFC6902 `path` roots be `metadata` and `specs` respectively.
 
 @danielballan mentioned pulling in @dylanmcreynolds since he has a good eye for making similar API design decisions that are both conforming to spec and aesthetically pleasing. I'd like to see what his suggestion is.

hyperrealist:
Thanks @Kezzsim, this is something that had been bothering me in the background at times. I appreciate your attention to detail!

>  1. Split the endpoint into `PATCH /api/v1/metadata/{path}` and `PATCH /api/v1/specs/{path}` which both take valid json-patch objects that alter their respective properties at the root path.
> 
>  2. (As seen in the example above) change the body to be a valid json-patch but make the RFC6902 `path` roots be `metadata` and `specs` respectively.

1. If it is typical to have to update both `metadata` and `specs` together this would necessitate extra http requests. If that is not the case I think this is a very much viable solution.
2. This is a creative way to conform to the http spec but I don't like how arguments are mixed together, especially since _how_ they are mixed together would change depending on the mimetype (we support both `application/json-patch+json` and `application/merge-patch+json`. See [here](https://github.com/bluesky/tiled/blob/2faa460a44da2b08fbc0f37592ec3815fbfa3812/tiled/server/router.py#L1397-L1405)).
3. A third solution might be to let the payload be of `application/json` mimetype and split the endpoint into something like `PATCH /api/v1/patch_metadata/{path}` and `PATCH /api/v1/merge_metadata/{path}`, each requiring a valid json with an embedded json patch or a merge patch respectively. But I think this is achieving little at the expense of polluting the API and adding to the complexity.
4. A fourth solution is to enforce a sub-mimetype at the content level:
```py
{
 "type": "application/json-patch+json",  # could default to this. may also support aliases  "json-patch"
 "metadata": [
   {
     "op": "replace",
     "path": "/test",
     "value": "testing"
   }
 ],
 "specs": [],
}
```
and
```py
{
 "type": "application/merge-patch+json",  # or "merge-patch" / "merge"
 "metadata": {"test": "testing"},
 "specs": [],
}
```

All these solutions would conform to http spec, but I am leaning toward solutions 1 or 4 for their conceptual simplicity and ease of implementation. Anyway, I am completely open to other ideas and design considerations.

danielballan:
Yes, good catch @Kezzsim! And thanks for presenting some additional options @hyperrealist.

Some additional design considerations:

Option (1) would break symmetry between GET and PATCH, as `GET /metadata/{path}` currently returns all "metadata" about a node, broadly defined, including structure family, structure, specs, data sources, and metadata. Making separate routes for PATCH seems asymmetrical in this sense, in addition to introducing overhead from separate requests for some use cases.

I can foresee adding support for patching data sources, alongside metadata and specs, which would make the downsides with (2) and (3) even more salient.

I can see no problems with (4). There was something appealing about specifying the patch format in the `content-type` header, but there is also something clean about making the request plain JSON (everybody knows JSON!) with the patch format specified in the body. It seems to keep our options open.

If we go that way, it is worth considering options for what to name the key. Is it `content-type`, `format`, `minetype`, `type`?

hyperrealist:
I very much intentionally wanted to avoid calling it `content-type`. I wouldn't envy whoever is documenting `content-type` the http header and `content-type` the subtype for this endpoint. `mimetype` sounds too restrictive to me as I would hesitate to support aliases under that name. Between `format` and `type` I like `type` because it hints at this being a mimetype without being overly restrictive. If adding a bit to the boilerplate is not an issue I would go with `patch-type`, `metadata-patch`, and `specs-patch`.

danielballan:
Seeing :+1:s from @Kezzsim and myself, I think going ahead with (4) with name `type` makes sense.

dylanmcreynolds:
> I very much intentionally wanted to avoid calling it content-type. I wouldn't envy whoever is documenting content-type the http header and content-type the subtype for this endpoint. mimetype sounds too restrictive to me as I would hesitate to support aliases under that name.

I'm not sure I understand the hesitance about calling it `content-type`. There's precedence here. For example, it normally appears in both the header and the body of multi-part form posts: https://swagger.io/docs/specification/describing-request-body/multipart-requests/

hyperrealist:
CI errors for Python 3.11 are coming from dask's incompatibility with 3.11.9 ([fixed in 2024.4.1](https://github.com/dask/dask/issues/11038)). Github CI uses 3.11.9, so it is broken at the moment.

I pinned python<3.11.9 for now, but I am not sure if I should attempt to resolve this issue in this PR. We should probably update to dask 2024.4.1.

tiled/server/router.py
```diff
@@ -1390,6 +1372,62 @@ async def put_awkward_full(
...
+    settings: BaseSettings = Depends(get_settings),
+    entry=SecureEntry(scopes=["write:metadata"]),
+):
+    if not hasattr(entry, "replace_metadata"):
+        raise HTTPException(
+            status_code=HTTP_405_METHOD_NOT_ALLOWED,
+            detail="This node does not support update of metadata.",
+        )
+
+    if body.content_type in [
```
danielballan, line 1389: In the spirit of, "Adhere to standards wherever possible," I would prefer to start with strictly accepting `application/json-patch+json` and `application/merge-patch+json` here. The use case for aliases seems narrow to me. The MIME type are unambiguous, Google-able, and easy enough to copy/paste from examples, or achieve through tooling like the Tiled Python client.

At least, I would wait to add aliases until/unless there is demonstrated demand for it that can be easily addressed another way. It's easy to add this later if we need to, but hard to remove once supported.

tiled/server/schemas.py
```diff
@@ -499,3 +500,45 @@ def specs_uniqueness_validator(cls, v):
...
+        "from": str,
+        "value": Any,
+    },
+    total=False,
+)
+
+
+class HyphenizedBaseModel(pydantic.BaseModel):
+    # This model configuration allows aliases like "content-type"
+    model_config = ConfigDict(alias_generator=lambda f: f.replace("_", "-"))
```
danielballan, line 520: Clever!

danielballan:
FYI, @hyperrealist, you may have seen this go by already but in case not, the CI failures on Python 3.11 are not the fault of this PR and can be ignored. See #715.

danielballan:
Rebased to pull in latest `CHANGELOG` and place this change entry on "Next" release.

danielballan:
Revisiting @hyperrealist's [comment](https://github.com/bluesky/tiled/pull/688#issuecomment-2022135250) proposing this API:

```py
{
 "type": "application/json-patch+json",  # could default to this. may also support aliases  "json-patch"
 "metadata": [
   {
     "op": "replace",
     "path": "/test",
     "value": "testing"
   }
 ],
 "specs": [],
}
```
and
```py
{
 "type": "application/merge-patch+json",  # or "merge-patch" / "merge"
 "metadata": {"test": "testing"},
 "specs": [],
}
```

I now realize that this was a little ambiguous about what's happening with `specs`. As currently implemented in this PR, `specs` is handled the same as in `PUT`. The content of `specs` in the request _replaces_ the current `specs`.

```py
class PatchMetadataRequest(HyphenizedBaseModel):
    content_type: str

    # These fields are optional because None means "no changes; do not update".
    patch: Optional[Union[List[JSONPatchSpec], Dict]]  # Dict for merge-patch
    specs: Optional[Specs]
```

I think it would be more useful (and more consistent) to make `specs` _also_ a patch.

```py
class PatchMetadataRequest(HyphenizedBaseModel):
    content_type: str

    # These fields are optional because None means "no changes; do not update".
    metadata: Optional[Union[List[JSONPatchSpec], Dict]]  # Dict for merge-patch
    specs: Optional[Union[List[JSONPatchSpec], Dict]]  # Dict for merge-patch
```

Notes:
- The user has to pick one patch `content-type`  for the whole request; they cannot be mixed. This seems like a good simplifying constraint.
- For specs, the performance advantage of "patch" vs "replace" is low. The list of `specs` is not expected to become long.
- But **in a future PR** I foresee adding `data_sources` here, and at that point I think we'll be glad for a consistent way to amend the `metadata`, `specs` , and `data_sources`, each with their own patch.

Existing unit tests cover `update_metadata`, but I think it's worthwhile to add a couple tests covering the new methods.

PR #691, opened 2024-03-12
More fixes to the DataSource registration how-to giude

danielballan:
Somehow in the flurry of recent activity here I still had not tested that this example _runs_. Now it runs.


padraic-shafer (approved):
I'm not sure when I'll have a chance to test this. But even if this turns out to be an incomplete fix, I favor adopting this improvement.

genematx (approved):

PR #694, opened 2024-03-19
Refactor to use fast api status constants

### LINKED ISSUE #650: Refactor to use FastAPI status constants
Suggested by @DiamondJoseph (and I agree) we should use human-readable status code constants. https://fastapi.tiangolo.com/tutorial/response-status-code/#shortcut-to-remember-the-names

In fact these [are defined in starlette](https://github.com/encode/starlette/blob/master/starlette/status.py) so we can import them from there.

skarakuzu:
Changed the HTTP messages to the format that starlette library uses.

Closes #650

### DIFF
--- a/tiled/_tests/test_access_control.py
+++ b/tiled/_tests/test_access_control.py
@@ -2,6 +2,7 @@ import json
 
 import numpy
 import pytest
+from starlette.status import HTTP_403_FORBIDDEN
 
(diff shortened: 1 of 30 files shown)


danielballan (approved):
Looks great

danielballan:
Thanks @DiamondJoseph for suggesting this. Much improved for readability.

PR #695, opened 2024-03-21
Update to pydantic v2

### LINKED ISSUE #669: Update to pydantic v2
Pydantic provides an automated migration tool that should do the bulk of the work in an automated way, such that we can:

1. Run the tool.
2. Run our test suite.
3. Fix whatever broke.

https://docs.pydantic.dev/latest/migration/

This is worth doing soon, as `event-model` [recently updated to pydantic 2.0](https://github.com/bluesky/event-model/pull/283).

skarakuzu:
update pydantic to v2. note: there is still bug
closes #669


tiled/server/schemas.py
```diff
@@ -62,9 +61,17 @@ class EntryFields(str, enum.Enum):
 
 
 class NodeStructure(pydantic.BaseModel):
-    contents: Optional[Dict[str, Resource[NodeAttributes, ResourceLinksT, EmptyDict]]]
+    # contents: Optional[Dict[str, Resource[NodeAttributes, ResourceLinksT, EmptyDict]]]
```
danielballan, line 64: These commented lines can be removed.

danielballan (approved):
Looks good! Just one last bit of cleanup to do, and then I think this can go in.

danielballan (approved):

PR #698, opened 2024-03-27
Include correlation ID in exception message.

danielballan:
For 5XX errors, client will include correlation ID in error message:

```
httpx.HTTPStatusError: Server error '500 Internal Server Error' for url 'http://testserver/error'
For more information, server admin can search server logs for correlation ID a13bf544cf8c2539.
```

### DIFF
--- a/tiled/client/utils.py
+++ b/tiled/client/utils.py
@@ -14,11 +14,54 @@ from ..utils import path_from_uri
 MSGPACK_MIME_TYPE = "application/x-msgpack"
 
 
+def raise_for_status(response) -> None:
+    """
+    Raise the `httpx.HTTPStatusError` if one occurred. Include correlation ID.
+    """
+    # This adapted from the method httpx.Response.raise_for_status, modified to
+    # remove the generic link to HTTP status documentation and include the
+    # correlation ID.
+    request = response._request
+    if request is None:
+        raise RuntimeError(
+            "Cannot call `raise_for_status` as the request "
+            "instance has not been set on this response."
+        )
+
+    if response.is_success:
+        return
+
+    if response.has_redirect_location:
+        message = (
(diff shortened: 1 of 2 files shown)


tiled/client/utils.py
```diff
@@ -14,11 +14,54 @@
 MSGPACK_MIME_TYPE = "application/x-msgpack"
 
 
+def raise_for_status(response) -> None:
+    """
+    Raise the `httpx.HTTPStatusError` if one occurred. Include correlation ID.
+    """
+    # This adapted from the method httpx.Response.raise_for_status, modified to
```
hyperrealist, line 21:
```suggestion
    # This is adapted from the method httpx.Response.raise_for_status, modified to
```

tiled/client/utils.py
```diff
@@ -14,11 +14,54 @@
...
+    # correlation ID.
+    request = response._request
+    if request is None:
+        raise RuntimeError(
+            "Cannot call `raise_for_status` as the request "
+            "instance has not been set on this response."
+        )
+
+    if response.is_success:
+        return
```
hyperrealist, line 32: Should `response` be returned to be consistent with `httpx.Response.raise_for_status`?

https://github.com/encode/httpx/blob/392dbe45f086d0877bd288c5d68abf860653b680/httpx/_models.py#L739-L740

hyperrealist (commented):
Looks like a straightforward addition. But I have a couple of comments.

tiled/client/utils.py
danielballan: Ah, yeah, this recently [changed](https://github.com/encode/httpx/blob/master/CHANGELOG.md#0250-11th-september-2023), and I was going off of a slightly old version of the source. Good catch.

Fix pushed.

hyperrealist (approved):

Issue #616, opened 2023-12-02
Error initializing database


dylanmcreynolds:
Running the va109 docker image, I exec into the container and try to initialize the database using

```
$ tiled admin initialize-database postgresql+asyncpg://user:pass@postgres/tiled
```

It looks like it's talking to the database just fine, but then blows up with some sort of sqlalchemy error.

```

Database postgresql+asyncpg://tiled:***@postgres/tiled is new. Creating tables and marking revision c7bd2573716d.
Traceback (most recent call last):

  File "/opt/venv/lib/python3.11/site-packages/tiled/commandline/_admin.py", line 35, in do_setup
    await check_database(engine, REQUIRED_REVISION, ALL_REVISIONS)

  File "/opt/venv/lib/python3.11/site-packages/tiled/alembic_utils.py", line 139, in check_database
    raise UninitializedDatabase(

tiled.alembic_utils.UninitializedDatabase: The database postgresql+asyncpg://tiled:***@postgres/tiled has no revision stamp. It may be empty. It can be initialized with `initialize_database(engine)`.


During handling of the above exception, another exception occurred:


Traceback (most recent call last):

  File "/opt/venv/lib/python3.11/site-packages/sqlalchemy/dialects/postgresql/asyncpg.py", line 528, in _prepare_and_execute
    prepared_stmt, attributes = await adapt_connection._prepare(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/venv/lib/python3.11/site-packages/sqlalchemy/dialects/postgresql/asyncpg.py", line 779, in _prepare
    prepared_stmt = await self._connection.prepare(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/venv/lib/python3.11/site-packages/asyncpg/connection.py", line 565, in prepare
    return await self._prepare(
           ^^^^^^^^^^^^^^^^^^^^

  File "/opt/venv/lib/python3.11/site-packages/asyncpg/connection.py", line 583, in _prepare
    stmt = await self._get_statement(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/opt/venv/lib/python3.11/site-packages/asyncpg/connection.py", line 397, in _get_statement
    statement = await self._protocol.prepare(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

asyncpg.exceptions.UndefinedTableError: relation "roles" does not exist


[...]

PR #703, opened 2024-03-29
Commit create_all transaction.

danielballan:
via @kleinhenz in https://github.com/bluesky/tiled/issues/616


danielballan:
We call `create_all()` in one other place, for the catalog database, but that one already has the necessary commit call:

https://github.com/bluesky/tiled/blob/1369d48036454a7aff9f8c88bf8386eaeb2c89f6/tiled/catalog/core.py#L28-L35

dylanmcreynolds (approved):
Looks good.

kleinhenz:
Thanks!

PR #705, opened 2024-04-01
on startup only add admin role if not already present

kleinhenz:
Previously `make_admin_by_identity` unconditionally appended the admin role to the given principal.
This caused restarting the server to fail:
```
#config.yaml
database:
  uri: "sqlite+aiosqlite:///./authn.db"

authentication:
  providers:
  - provider: toy
    authenticator: tiled.authenticators:DictionaryAuthenticator
    args:
      users_to_passwords:
        alice: abc
  tiled_admins:
    - provider: toy
      id: alice
trees:
  - path: /
    tree: tiled.examples.toy_authentication:tree
```

```
> tiled admin initialize-database sqlite+aiosqlite:///./authn.db
> tiled serve config config.yaml
> tiled serve config config.yaml # fails when run the second time

...
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: principal_role_association.principal_id, 
principal_role_association.role_id
[SQL: INSERT INTO principal_role_association (principal_id, role_id) VALUES (?, ?)]
[parameters: (1, 2)]
```

To fix, check if the principal already has the admin role before appending it again.

Also includes two flyby fixes:
* provide fallback-version so install can succeed without full repo which can happen when building a docker container in a worktree
* handle missing correlation id in `raise_for_status`. This can happen if the request didn't make it to the tiled server, for example because the k8s ingress gave 413.

### DIFF
--- a/continuous_integration/scripts/start_LDAP.sh
+++ b/continuous_integration/scripts/start_LDAP.sh
@@ -3,5 +3,6 @@ set -e
 
 # Start LDAP server in docker container
 docker pull bitnami/openldap:latest
+pip install docker-compose
 docker-compose -f continuous_integration/docker-configs/ldap-docker-compose.yml up -d
 docker ps
--- a/continuous_integration/scripts/start_LDAP.sh
+++ b/continuous_integration/scripts/start_LDAP.sh
@@ -3,6 +3,5 @@ set -e
 
 # Start LDAP server in docker container
 docker pull bitnami/openldap:latest
-pip install docker-compose
-docker-compose -f continuous_integration/docker-configs/ldap-docker-compose.yml up -d
+docker compose -f continuous_integration/docker-configs/ldap-docker-compose.yml up -d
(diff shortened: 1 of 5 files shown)


danielballan:
Nice, thanks!

I didn't know about `fallback-version`.

I added a commit with a test that starts the server twice, reproducing the bug fixed by this PR:

```
================================================= short test summary info =================================================
FAILED tiled/_tests/test_authentication.py::test_admin - sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: principal_role_association.principal...
```

danielballan:
Saw a new CI failure here, where the `docker-compose` executable was missing on some but not all Python jobs. Added a line to install it explicitly; can't hurt.

danielballan:
https://stackoverflow.com/a/77651191/1221924

Issue #663, opened 2024-02-22
Investigate and optimize SQLite connection pool management


danielballan:
We have observed what appears to be frequently (re-)opening of the SQLite catalog database. I have read that applications should hold a connection pool with one read-write connection and N read connections.

At minimum, we should avoid thrashing the file like this (unless I am confused). At best, we should optimize our pool management.

Demo:

```
# Start server and strace it. Note -f.
$ strace -fe open,openat tiled serve catalog /tmp/strace_test.db --init --api-key secret -w /tmp/strace_test_data/
```

Note that this has to be warmed up, as Tiled lazily imports some Python modules. 

```py
from tiled.client import from_uri
c = from_uri('http://localhost:8000', api_key='secret')
c.write_array([1,2,3])  # Warm up
c.write_array([1,2, 3])  # Test
```

```
strace: Process 197660 attached
[pid 197660] openat(AT_FDCWD, "/tmp/strace_test.db", O_RDWR|O_CREAT|O_NOFOLLOW|O_CLOEXEC, 0644) = 15
[pid 197660] openat(AT_FDCWD, "/tmp/strace_test.db-wal", O_RDWR|O_CREAT|O_NOFOLLOW|O_CLOEXEC, 0644) = 16
[pid 197660] openat(AT_FDCWD, "/tmp/strace_test.db-shm", O_RDWR|O_CREAT|O_NOFOLLOW|O_CLOEXEC, 0644) = 17
[pid 197635] openat(AT_FDCWD, "/tmp/strace_test_data/65d1e164-2a35-4422-959e-222ccbc57d73/.zarray", O_RDONLY|O_CLOEXEC) = 15
[pid 197635] openat(AT_FDCWD, "/tmp/strace_test_data/65d1e164-2a35-4422-959e-222ccbc57d73/0.f490a5d74b2a44f487f4f5c796f62228.partial", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 15
[pid 197657] +++ exited with 0 +++
INFO:     127.0.0.1:36470 - "PUT /api/v1/array/full/65d1e164-2a35-4422-959e-222ccbc57d73 HTTP/1.1" 200 OK
```

danielballan:
I think the first mystery is: where do all these subprocesses come from?

nmaytan:
[...]

PR #708, opened 2024-04-03
Remove "object cache" misfeature; add simpler "resource cache"

### LINKED ISSUE #683: Remove "object cache" and replace it with something much simpler
The ["object cache"](https://github.com/bluesky/tiled/blob/main/tiled/server/object_cache.py) was a speculative development, added in early days when we were still evaluating the viability of the overall endeavor. Its contribution to performance was never thoroughly tested.

The most important problem it is solving is holding onto a file handle so that burst of requests for data from the same file do not repeatedly re-open the file. I think we could solve this with a simple cache of file handles. It should be bounded by both time (never hold a file open for longer than T seconds) and number of files. Perhaps a mix of TTL and LRU.

I think it might make sense to simple strip out the object cache entirely, replace it with the simpler thing, and only add back more complexity if it is proven to be needed and we find the resources to do it well.

danielballan:
The "object cache" was a misfeature because it mixed two distinct resource pools:

(1) Caching chunks of _data_
(2) Caching file handles

(I admit this should have been immediately clear at the time it was added. This was very early on when we were prototyping a lot of aspects to validate the general idea.)

This PR removes the "object cache" entirely. The only remaining mention is in the service configuration, where `object_cache` is still accepted, for backward-compatibility, but a warning is emitted that any configuration here has no effect:

```
$ git grep object_cache
tiled/config.py:        if "object_cache" in config:
tiled/config_schemas/service_configuration.yml:  object_cache:
```

Then the PR adds in its place a "resource cache" which:

- addresses only file handles (2), not cached data (1)
- involves much less custom code, using [`cachetools.TLRUCache`](https://cachetools.readthedocs.io/en/latest/#cachetools.TLRUCache) to do the heavy lifting

Here we can see it in action. The file is only opened while processing the first request. The second request uses a cached file handle.

```
$ strace -fe open,openat tiled serve directory files/ --public
```

This is the relevant output. (Some noise related to #663 has been removed for clarity.)

```
[pid 2870299] openat(AT_FDCWD, "/home/dallan/Repos/bnl/tiled/files/a.tif", O_RDONLY|O_CLOEXEC) = 17
INFO:     127.0.0.1:57168 - "GET /api/v1/array/full/a HTTP/1.1" 200 OK
INFO:     127.0.0.1:57176 - "GET /api/v1/array/full/a HTTP/1.1" 200 OK
```

```
[pid 2871270] openat(AT_FDCWD, "/home/dallan/Repos/bnl/tiled/files/stuff.h5", O_RDONLY) = 18
INFO:     127.0.0.1:52242 - "GET /api/v1/array/full/stuff/data HTTP/1.1" 200 OK
INFO:     127.0.0.1:52246 - "GET /api/v1/array/full/stuff/data HTTP/1.1" 200 OK
```

After waiting for the cache expiry time (default 60 seconds) accessing it again opens the file again, as expected.

The cache parameters are tunable via environment variables, but this is intentionally not exposed through the service configuration or CLI. I think it's possible that we can choose sensible defaults that will not need to be overridden except in debugging scenarios.

Closes #683


danielballan:
A server side data cache, use case (1), is still of interest, but should be explored in future issues and PRs. The primary goal of this PR is to remove the object cache in preparation for a beta release, and to put in place a way to do basic file handle caching, which we know can be important, to avoid thrashing file handles when serving a burst of parallel requests.

tiled/adapters/resource_cache.py
```diff
@@ -0,0 +1,61 @@
...
+):
+    """
+    Use value from cache or, if not present, call factory(*args, **kwargs) and cache result.
+    """
+    if _resource_cache is None:
+        cache = get_resource_cache()
+    else:
+        cache = _resource_cache
+    # Return cached value if found.
+    value = cache.get(cache_key)
```
genematx, line 52: I couldn't easily find what happens to the `ttu` when an item is returned from the cache: is `ttu` updated for another 60 seconds or just let to expire? It doesn't matter that much, of course, especially for this PR, but could be interesting to see if some resources are requested so often that _recaching_ them creates a considerable overhead.

tiled/adapters/hdf5.py
```diff
@@ -114,7 +115,10 @@ def from_uri(
         access_policy=None,
     ):
         filepath = path_from_uri(data_uri)
-        file = h5py.File(filepath, "r", swmr=swmr, libver=libver)
+        cache_key = (cls.__module__, cls.__name__, filepath)
```
genematx, line 118: Just a thought... I wonder about implications of keying the resources by `cls.__name__` and `cls.__module__`... Could there be situations when the same file is cached from different adapters at (roughly) the same time? (Probably not a big issue for reading). But this key definitely has great observability and traceability.
Another possible option could be to delegate the creation of cache keys to the `with_resource_cache` function itself (possibly with an option to override), so that whoever writes the adapter doesn't need to worry about keys at all. Then a key could be a hash of `filename`, all `*args`, and all `*kwargs` (but again, it might create problems if multiple threads would be trying to access a file concurrently).

genematx (approved):
I really like the mechanics of `with_resource_cache`, and how it simplifies the code overall. Just a few notes from me; nothing major.

tiled/adapters/resource_cache.py
danielballan: Good thought.

The expiry time is computed once, at cache insertion, and not renewed.

https://github.com/tkem/cachetools/blob/d8b5c608273051f2ebddaec8a4d033c66369dd7d/src/cachetools/__init__.py#L546

For our use case it does seem better to let the expiry renew (the clock reset) on access. I cannot see anyway to make it renew using public APIs. Worth following up in a later PR, and maybe even a PR into cachetools.

tiled/adapters/hdf5.py
danielballan: My concern was two adapters stepping on each other, maybe even accessing the same file in some weird scenario but caching different types of objects, say from different I/O libraries like tifffile versus imageio. If Adapter A caches a type of object and then Adapter B gets it from the cache this could lead to very confusing bugs.

But maybe I am overengineering this for a very weird hypothetical case. I like your idea of making the cache key optional, and by default hashing the inputs.

tiled/adapters/resource_cache.py
```diff
@@ -0,0 +1,61 @@
+import os
+from typing import Callable, Optional
+
+import cachetools
+
+# Cached items will be evicted if not used for a specified time interval
+# ("time to use"). If the cache size reaches its max size, the least recently
+# used cache item will be evicted.
```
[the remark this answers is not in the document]
danielballan: Done

docs/source/explanations/caching.md
```diff
@@ -11,17 +11,11 @@ Tiled has two kinds of caching:
 
 1. **Client-side response cache.** The Tiled Python client implements a standard
    web cache, similar in both concept and implementation to a web browser's cache.
-3. **Service-side object cache.** The _response_ caches operate near the outer
+3. **Service-side resource cache.** The _response_ caches operate near the outer
```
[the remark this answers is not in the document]
danielballan: Done

tiled/adapters/hdf5.py
genematx: Potential collisions with other `kwargs` is a really good point; I haven't thought about it tbh. Though, I agree, it's probably rather unlikely, especially if we call it `_cache_key` with underscore (or even `_resource_cache_key` to be extra safe, but probably unnecessary).
I personally would be in favour of option 2, as it adds the least overhead on the caller of `with_resource_cache`; option 3 is nice, but I feel it could be too complicated. Actually, if there are any clashes with `kwargs` in option 2, one can always resort to `functools.partial`.

dylanmcreynolds: On one hand, I like the idea of implementing the `cache_key` in the framework, taking control of criticl piece of infrastructure that is so very easy to get wrong and has terrible consequences when it's wrong.

On the other hand,  I also like the idea of the adapter developer having to produce a `cache_key` that is appropriate for their adapter's underlying file type. I have already seen stupid use cases for adapters a am little skeptical of our ability to produce the right `cache_key` for all cases.

danielballan: Another option: mash up these ideas by keeping the explicit `cache_key` but generally using the pattern:

```py
cache_key = (f, args, sorted(kwargs.items())
x = with_resource_cache(cache_key, f, *args, **kwargs)
```

rather than my `(type(self).__module__, type(self).__name__, ...)` pattern.

tacaswell: Despite the risk of abuse / poor choices, I think leaving the cache key generation up to the adapter is the right choice.  Trying to do something generic by typesniffing + ensure that kwargs are hashable (see some of the wild performance issues we found it intake) is also something that is very easy to get wrong and "sometimes framework does it, sometimes the handlers do it" makes thing harder to debug / track down.

If we expect 100s-1000s of adapters to exist / to be provided by the end-users scientists then maybe I see the case for the framework doing it, but we see <100 adapters mostly written by controls / computing group staff I think making sure the cache key is "good" is easier and clearer in the long run.

danielballan: Decision then, with @genematx on Zoom: Adapter must specify the key. We'll update the cache key to be more specific so that potentially different adapters can share cache items if they are both opening the same file with an identical function call.

PR #712, opened 2024-04-08
Add starlette to client requirements

### LINKED ISSUE #711: `client` requirements should include `starlette`
#694 and #650 added `starlette` status constants to the client, but `starlette` never made it into the client requirements causing an import error if you only install `tiled[client]`.

I believe it should be in both [client] and [minimal-client].

A broader question as this has happened before: could we add a CI test that installs the package with just `[minimal-client]` or `[client]` and verifies (to first order) that the package imports, and ideally that a basic client function test passes, maybe against the public test server?  This would likely have to be its own GHA/non-`pytest` script since afaik there is no environment control at the `pytest` level.

pbeaucage:
Closes #711 by patching requirements.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,6 +125,7 @@ client = [
     "pyyaml",
     "rich",
     "sparse",
+    "starlette",
     "typer",
     "xarray",
     "zstandard"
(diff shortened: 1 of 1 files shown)


danielballan:
I think we should address this by using `httpx.codes.*` to represent status codes on the client side.

danielballan:
Superseded byhttps://github.com/bluesky/tiled/pull/716

PR #714, opened 2024-04-09
Log catalog init

dylanmcreynolds:
Tiled can start up in a mode where it initializes the catalog database. This is done in a subprocess call. I have had several cases where that failed, but I didn't know. So, I'd like to add logging from the subprocess to tiled's main log. Happy to take advice on a more graceful way to do it, but this seems to work.


danielballan (approved):

danielballan:
CI failure is unrelated, see #715

Issue #715, opened 2024-04-09
Another problem with a change in `dask.dataframe`


danielballan:
New CI failures on Python 3.11 only:

```
________________ ERROR collecting tiled/_tests/test_writing.py _________________
tiled/_tests/test_writing.py:10: in <module>
    import dask.dataframe
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/__init__.py:100: in <module>
    from dask.dataframe import backends, dispatch, rolling
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/backends.py:15: in <module>
    from dask.dataframe.core import DataFrame, Index, Scalar, Series, _Frame
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/core.py:36: in <module>
    from dask.dataframe import methods
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/methods.py:34: in <module>
    from dask.dataframe.utils import is_dataframe_like, is_index_like, is_series_like
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/utils.py:20: in <module>
    from dask.dataframe import (  # noqa: F401 register pandas extension types
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/_dtypes.py:9: in <module>
    from dask.dataframe.extensions import make_array_nonempty, make_scalar
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/extensions.py:8: in <module>
    from dask.dataframe.accessor import (
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/accessor.py:126: in <module>
    class DatetimeAccessor(Accessor):
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/accessor.py:81: in __init_subclass__
    _bind_property(cls, pd_cls, attr, min_version)
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/dask/dataframe/accessor.py:35: in _bind_property
    setattr(cls, attr, property(derived_from(pd_cls, version=min_version)(func)))
[...]

PR #716, opened 2024-04-10
On client side, use client library for HTTP status code constants.

### LINKED ISSUE #711: `client` requirements should include `starlette`

danielballan:
Closes #711 
Supersedes #712 

attn @pbeaucage

### DIFF
--- a/tiled/client/auth.py
+++ b/tiled/client/auth.py
@@ -2,7 +2,6 @@ import os
 from pathlib import Path
 
 import httpx
-from starlette.status import HTTP_401_UNAUTHORIZED
 
 from .utils import SerializableLock, handle_error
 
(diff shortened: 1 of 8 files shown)


danielballan:
The Python 3.11 CI failure is unrelated (see #715); I think we can go ahead and merge this when it's reviewed and ready, handling #715 separately.

skarakuzu (approved):

PR #719, opened 2024-04-17
Switch to blosc2; add Python 3.12 support

danielballan:
The library [blosc2](https://pypi.org/project/blosc2/) is evidently superseding [blosc](https://pypi.org/project/blosc), which has not published a release since 2022 and has yet to publish Python 3.12-compatible wheels. This is blocking Tiled support for Python 3.12.

Blosc2 claims to be fully backward-compatible with Blosc1-serialized bytes. I'd like to verify that old tiled clients can talk to new tiled servers before we merge.

As an aside, while looking into this I noticed that [`distributed` dropped support for blosc](https://github.com/dask/distributed/pull/6027). Someday it would be good to evaluate how much value we are deriving from this dependency, but I'm inclined to keep it for now.

### DIFF
--- a/tiled/media_type_registration.py
+++ b/tiled/media_type_registration.py
@@ -328,14 +328,8 @@ if modules_available("lz4"):
     ]:
         compression_registry.register(media_type, "lz4", LZ4Buffer)
 
-if modules_available("blosc"):
-    import blosc
-
-    # The choice of settings here is cribbed from distributed.protocol.compression.
-    n = blosc.set_nthreads(2)
-    if hasattr("blosc", "releasegil"):
-        blosc.set_releasegil(True)
-    blosc_settings = {"cname": "lz4", "clevel": 5}
+if modules_available("blosc2"):
+    import blosc2
 
     class BloscBuffer:
         """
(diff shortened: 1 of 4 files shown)


danielballan:
> I'd like to verify that old tiled servers can talk to new tiled clients before we merge.

Looks good, as advertised :+1:

tiled/media_type_registration.py
```diff
@@ -328,14 +328,8 @@ def close(self):
     ]:
         compression_registry.register(media_type, "lz4", LZ4Buffer)
 
-if modules_available("blosc"):
-    import blosc
-
-    # The choice of settings here is cribbed from distributed.protocol.compression.
```
danielballan, line 334: I cargo-culted these settings from `dask.distributed` in the first place without any benchmarking. They aren't directly transferrable to blsoc2, so I have reverted to just going with the defaults for now until we have time to benchmark.

hyperrealist: Do you want to retain this TODO as a comment or as a low-priority github issue?

hyperrealist (approved):

tiled/media_type_registration.py
danielballan: Good call. Done https://github.com/bluesky/tiled/issues/724

PR #720, opened 2024-04-18
Fix Comparison Query Documentation

jmaruland:
This PR fixes a typo that it was noticed in the documentation for the Comparison class


danielballan:
Thanks for following up on this @jmaruland! Lovely.

Yes, the CI failure is #715 and can be ignored for the purposes of this PR.

PR #723, opened 2024-04-19
Support FullText search on SQLite

danielballan:
This PR has commits from @Kezzsim and me. It adds support for full text search backed by SQLite.

SQLite full text search support via [FTS5](https://www.sqlite.org/fts5.html) is solid and feature-ful, but the implementation is somewhat different. It employs a `VIRTUAL TABLE` instead of an `INDEX`. Database triggers are used to keep the content of this table in sync with the `nodes` table. We must perform a `JOIN` on that virtual table in order to filter our results with a `MATCH` query against the indexed node `metadata`.

We get queries that look like:

```
SELECT nodes."key" 
FROM nodes JOIN metadata_fts5 ON metadata_fts5.rowid = nodes.id 
WHERE nodes.ancestors = ? AND metadata_fts5.metadata MATCH ? ORDER BY nodes.time_created, nodes.id
 LIMIT ? OFFSET ?
```

(as shown by setting `TILED_ECHO_SQL=1` when starting the server).

I think the implementation is about as clean as it can be, until we go and refactor all the SQLAlchemy code into a separate module. But I'd like to kick the tires a bit more before we merge.

We also need to test the migration interactively, and particularly ensure that _existing rows_ in the `nodes` table are added to the `VIRTUAL TABLE`.

Closes #546

### DIFF
--- a/tiled/catalog/migrations/versions/ed3a4223a600_create_sqlite_table_for_fulltext_search.py
+++ b/tiled/catalog/migrations/versions/ed3a4223a600_create_sqlite_table_for_fulltext_search.py
@@ -39,9 +39,9 @@ def upgrade():
             """,
             """
             CREATE TRIGGER nodes_metadata_fts5_sync_au AFTER UPDATE ON nodes BEGIN
-              INSERT INTO metadata_ft5_index(metadata_ft5_index, rowid, metadata)
+              INSERT INTO metadata_fts5_index(metadata_fts5_index, rowid, metadata)
               VALUES('delete', old.id, old.metadata);
-              INSERT INTO metadata_ft5_index(rowid, metadata)
+              INSERT INTO metadata_fts5_index(rowid, metadata)
               VALUES (new.id, new.metadata);
             END;
             """,
--- a/tiled/catalog/migrations/versions/ed3a4223a600_create_sqlite_table_for_fulltext_search.py
+++ b/tiled/catalog/migrations/versions/ed3a4223a600_create_sqlite_table_for_fulltext_search.py
@@ -1,0 +1,55 @@
+"""Create sqlite table for fulltext search
+
+Revision ID: ed3a4223a600
+Revises: e756b9381c14
+Create Date: 2024-04-11 16:41:01.369520
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'ed3a4223a600'
+down_revision = 'e756b9381c14'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    connection = op.get_bind()
+    if connection.engine.dialect.name == "sqlite":
+        statements = [
+            # Create an external content fts5 table.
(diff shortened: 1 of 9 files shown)


danielballan:
Locally, I am seeing these test failures. I think there is something race-y about table creation. The failures are not all reproducible if run in isolation.

```
FAILED tiled/_tests/test_access_control.py::test_create_and_update_allowed - sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: main.metadata_ft5_index
FAILED tiled/_tests/test_decoders.py::test_blosc - AssertionError: assert 'blosc' in 'gzip, deflate, zstd'
FAILED tiled/_tests/test_object_cache.py::test_detect_content_changed_or_removed - assert 1 == 2
FAILED tiled/_tests/test_queries.py::test_full_text[sqlite] - tiled.client.utils.ClientError: 400: The query type 'full_text' is not supported on this node. http://local-tiled-app/api/v...
FAILED tiled/_tests/test_writing.py::test_write_dataframe_partitioned - dask.base.TokenizationError: Object <function Container.write_dataframe.<locals>.write_partition at 0x7a554502b420> cannot ...
FAILED tiled/_tests/test_writing.py::test_limits - sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: main.metadata_ft5_index
FAILED tiled/_tests/test_writing.py::test_metadata_revisions - sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: main.metadata_ft5_index
```

danielballan:
Some of the above failures are unrelated, and simply due to this branch being old and carrying some upstream incompatibility issues that have been fixed on `main`. I have rebased on `main`.

danielballan:
Unit tests pass locally for me now.

- [ ] The migration `tiled/catalog/migrations/versions/ed3a4223a600_create_sqlite_table_for_fulltext_search.py` must be extended to index all the existing rows in `nodes` at the time at the migration is running, effectively like running the trigger retroactively for all the rows that were added before we starting indexing them.
- [ ] Test for each of the triggers
    - [ ] A node deleted no longer appears in full text search results.
    - [ ] A node that has had its metadata updated matches its new value but not its old value.

tiled/catalog/migrations/versions/ed3a4223a600_create_sqlite_table_for_fulltext_search.py
```diff
@@ -24,6 +24,11 @@ def upgrade():
             """
             CREATE VIRTUAL TABLE metadata_fts5 USING fts5(metadata, content='nodes', content_rowid='id');
             """,
+            # Insert all existing node content into the fts5 table.
+            """
+            INSERT INTO metadata_fts5(rowid, metadata)
+            SELECT id, metadata FROM nodes;
+            """
```
danielballan, line 31: Missing trialing comma after close-quotes.

danielballan:
For reference I wanted to drop this link here --- this is where I found the approach we settled on for creating a `VIRTUAL TABLE` via SQLAlchemy: https://github.com/sqlalchemy/sqlalchemy/discussions/9466

PR #728, opened 2024-04-24
Do softer version check.

danielballan:
The Tiled Python client reports the version of Tiled that it is using its `user-agent` header, e.g. `python-tiled/0.1.0a118`. If this client version is known to be too old to interoperate with the Tiled server version, the server returns a 400 status code with an explanatory error message.

This can go wrong in a development install.. If the user has not recently fetched the upstream tags, it may report a version like `0.1.dev2335+g58bc300.d20240409`. This has created confusion in two unrelated incidents within the last two weeks.

This PR makes the version check softer:

1. If the version cannot be parsed, log a warning in the server logs and ignore.
2. If the version is a dev release, ignore.

This is not perfect. But I think the change does more good than harm.

### DIFF
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -716,17 +716,14 @@ Back up the database, and then run:
             try:
                 parsed_version = packaging.version.parse(raw_version)
             except Exception:
-                return JSONResponse(
-                    status_code=HTTP_400_BAD_REQUEST,
-                    content={
-                        "detail": (
-                            f"Python Tiled client is version is reported as {raw_version}. "
-                            "This cannot be parsed as a valid version."
-                        ),
-                    },
+                logger.warning(
+                    f"Python Tiled client is version is reported as {raw_version}. "
+                    "This cannot be parsed as a valid version."
                 )
             else:
-                if parsed_version < MINIMUM_SUPPORTED_PYTHON_CLIENT_VERSION:
+                if (not parsed_version.is_deverlease) and (
+                    parsed_version < MINIMUM_SUPPORTED_PYTHON_CLIENT_VERSION
+                ):
                     return JSONResponse(
                         status_code=HTTP_400_BAD_REQUEST,
                         content={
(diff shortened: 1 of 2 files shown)


tiled/server/app.py
```diff
@@ -716,17 +716,14 @@ async def client_compatibility_check(request: Request, call_next):
...
-                            "This cannot be parsed as a valid version."
-                        ),
-                    },
+                logger.warning(
+                    f"Python Tiled client is version is reported as {raw_version}. "
+                    "This cannot be parsed as a valid version."
                 )
             else:
-                if parsed_version < MINIMUM_SUPPORTED_PYTHON_CLIENT_VERSION:
+                if (not parsed_version.is_deverlease) and (
```
padraic-shafer, line 724:
```suggestion
                if (not parsed_version.is_devrelease) and (
```

tiled/server/app.py
```diff
@@ -716,17 +716,14 @@ async def client_compatibility_check(request: Request, call_next):
...
-                return JSONResponse(
-                    status_code=HTTP_400_BAD_REQUEST,
-                    content={
-                        "detail": (
-                            f"Python Tiled client is version is reported as {raw_version}. "
-                            "This cannot be parsed as a valid version."
-                        ),
-                    },
+                logger.warning(
+                    f"Python Tiled client is version is reported as {raw_version}. "
```
padraic-shafer, line 720:
```suggestion
                    f"Python Tiled client version is reported as {raw_version}. "
```

padraic-shafer:
I took the liberty of fixing a couple typos that should hopefully make the tests pass.

padraic-shafer (commented):
The basic concept seems like a good trade-off.  Some tests that expected a `HTTP_400_BAD_REQUEST` response will need to be updated.

padraic-shafer:
@danielballan 
I've pushed a possible update for the tests, and included a warning as well.  Please revert or update as desired.

danielballan:
Thanks for the assist @padraic-shafer! And thanks for reporting the issue @hyperrealist and @dylanmcreynolds.

PR #729, opened 2024-04-25
Give blosc2 distinct content-encoding

danielballan:
Follow-up on #719, making an important correction.

Blosc2 can read blosc1-encoded data, but blosc1 cannot read blosc2-encoded data. Therefore, new Tiled servers will send old tiled clients bytes encoded as `content-encoding: blosc` but the clients were not be able to decode the bytes. @pbeaucage hit this in the wild after we deployed Tiled v0.1.0a119 to tiled.nsls2.bnl.gov yesterday.

This PR renamed the `content-encoding` header from `blosc` to `blosc2`. I considered this in #719 but could not quite tell which was correct. In retrospect it seems unambiguous that this is a distinct encoding and needs a distinct header value.


CHANGELOG.md
```diff
@@ -3,6 +3,14 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- The `content-encoding` `blosc` was recently upgraded from Blosc to Blsoc2.
```
padraic-shafer, line 10:
```suggestion
- The `content-encoding` `blosc` was recently upgraded from Blosc to Blosc2.
```

docs/source/explanations/compression.md
```diff
@@ -50,7 +50,7 @@ entries in this table above later ones.
 
 | Method                                                           | Accept-Encoding | Required Python Package |
 | ---------------------------------------------------------------- | --------------- | ----------------------- |
-| [blosc](https://www.blosc.org/)                                  | `blosc`         | `blosc`                 |
+| [blosc2](https://www.blosc.org/)                                 | `blosc2`        | `blosc2`                |
 | [lz4](https://en.wikipedia.org/wiki/LZ4_(compression_algorithm)) | `lz4`           | `lz4`                   |
 | [Zstandard](https://facebook.github.io/zstd/)                    | `zstd`          | `zstandard`             |
 | [gzip](https://en.wikipedia.org/wiki/Gzip)                       | `gzip`          | none (built in)         |
```
padraic-shafer, line 56:
```suggestion
| [gzip](https://en.wikipedia.org/wiki/Gzip)                       | `gzip`          | none (built in)         |
```

The line below this references `blosc` rathern than `blosc2`.  Should this get updated also?

```text
The Tiled Python *client* currently supports gzip and blosc (if the Python
package `blosc` is installed).
```

padraic-shafer (approved):
Overall this looks right.  I noted a couple things as comments.

Also, is it worth updating this how-to in the current PR?
https://github.com/bluesky/tiled/blob/main/docs/source/how-to/client-logger.md

docs/source/explanations/compression.md
danielballan: Done!

Issue #731, opened 2024-04-30
`select_metadata` parameter causes a Pydantic validation error


pbeaucage:
`select_metadata` is a minimally-documented, server-side-only feature implemented in response to #42 that down-filters metadata using `JMESPath` before responding to a call to `/search` or `/node/metadata`.  However, valid JMESPath queries result in a 500 Internal Server Error, with an error like:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for NodeAttributes
metadata
  Input should be a valid dictionary [type=dict_type, input_value='great', input_type=str]
    For further information visit https://errors.pydantic.dev/2.7/v/dict_type
```
i.e, we enforce that `metadata` should be a dictionary, but the returned, down-sampled metadata from `JMESPath` is either a string or a list depending on the return.
While there is a longer-term path to substitute JSONPath for JMESPath #217, it seems worth patching this.  One simple approach would be to coerce the output into a dictionary `{'selected': return_value}`, but I'm not sure if this is the best way, or if relaxing the type validation on the metadata makes more sense.

PR #732, opened 2024-04-30
Coerce metadata to dict if `select_metadata` is used

pbeaucage:
See description in #731.  This PR implements a minimal fix for this problem by forcing `metadata` into a one-member dictionary if `select_metadata` is used.

I'm not sure if a test should be added to cover this behavior, given that we intend to soon deprecate this feature in favor of a newer solution.  I would appreciate input here.  One complication is that the default example datasets do not have associated metadata...

 ### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 - SQLite-backed catalogs now employ connection pooling. This results in a
   significant speed-up and avoids frequently re-opening the SQLite file.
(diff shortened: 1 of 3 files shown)


pbeaucage:
Closes #731

danielballan:
Thanks @pbeaucage! 

I added a simple unit test in 5aeaf97d40729c9c30e32c0f4a99babff6c1547c just to protect this against major accidental regressions while we work out the permanently-supported solution. Will merge when CI passes.

PR #735, opened 2024-05-06
Add utility for syncing nodes between two Tiled instances

danielballan:
### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/_tests/test_writing.py
+++ b/tiled/_tests/test_writing.py
@@ -483,7 +483,7 @@ def test_write_with_specified_mimetype(tree):
             x.write_partition(df, 0)
             x.read()
             x.refresh()
-            x.data_sources()[0]["mimetype"] == mimetype
+            x.data_sources()[0].mimetype == mimetype
 
         # Specifying unsupported mimetype raises expected error.
         with fail_with_status_code(415):
--- a/tiled/_tests/test_writing.py
+++ b/tiled/_tests/test_writing.py
@@ -483,7 +483,7 @@ def test_write_with_specified_mimetype(tree):
             x.write_partition(df, 0)
             x.read()
             x.refresh()
-            x.data_sources()[0].mimetype == mimetype
+            assert x.data_sources()[0].mimetype == mimetype
 
         # Specifying unsupported mimetype raises expected error.
         with fail_with_status_code(415):
--- a/tiled/client/sync.py
+++ b/tiled/client/sync.py
@@ -1,4 +1,3 @@
-import collections.abc
 import itertools
 
 from ..structures.core import StructureFamily
@@ -39,11 +38,8 @@ def copy(
     if hasattr(source, "structure_family"):
         # looks like a client object
         _DISPATCH[source.structure_family](source.include_data_sources(), dest)
-    elif isinstance(source, list):
-        # such as result of source.items().head()
+    else:
         _DISPATCH[StructureFamily.container](dict(source), dest)
-    elif isinstance(source, collections.abc.Mapping):
-        _DISPATCH[StructureFamily.container](source, dest)
 
 
 def _copy_array(source, dest):
--- a/tiled/client/sync.py
+++ b/tiled/client/sync.py
@@ -1,3 +1,4 @@
+import collections.abc
 import itertools
 
 from ..structures.core import StructureFamily
@@ -18,8 +19,18 @@ def sync(
     source : tiled node
     dest : tiled node
     """
-    source = source.include_data_sources().refresh()
-    _DISPATCH[source.structure_family](source, dest, copy_internal, copy_external)
+    if hasattr(source, "structure_family"):
+        # looks like a client object
+        _DISPATCH[source.structure_family](
+            source.include_data_sources(), dest, copy_internal, copy_external
+        )
+    elif isinstance(source, list):
+        # such as result of source.items().head()
+        _DISPATCH[StructureFamily.container](
+            dict(source), dest, copy_internal, copy_external
+        )
+    elif isinstance(source, collections.abc.Mapping):
+        _DISPATCH[StructureFamily.container](source, dest, copy_internal, copy_external)
 
 
 def _sync_array(source, dest, copy_internal, copy_external):
@@ -38,7 +49,7 @@ def _sync_table(source, dest, copy_internal, copy_external):
 
 def _sync_container(source, dest, copy_internal, copy_external):
     for key, child_node in source.items():
-        original_data_sources = child_node.data_sources()
+        original_data_sources = child_node.include_data_sources().data_sources()
         if not original_data_sources:
             if child_node.structure_family == StructureFamily.container:
                 data_sources = []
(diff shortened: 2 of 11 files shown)


danielballan:
Testing like:

```py
from tiled.client.sync import sync


a = from_uri("http://localhost:8000", api_key="secret")
b = from_uri("http://localhost:9000", api_key="secret")


sync(a, b)
```

tiled/client/sync.py
```diff
@@ -0,0 +1,112 @@
...
+        dest.write_block(array, block)
+
+
+def _sync_table(source, dest):
+    for partition in range(source.structure().npartitions):
+        df = source.read_partition(partition)
+        dest.write_partition(df, partition)
+
+
+def _sync_container(source, dest):
```
danielballan, line 62: This function is more complex than I would like but after about 20 minutes of fiddling I couldn't land of a significantly simpler implementation. Might  be worth revisiting later.

tiled/client/sync.py
```diff
@@ -0,0 +1,112 @@
...
+            original_data_sources
+            and (original_data_sources[0].management != Management.external)
+        ) or (
+            child_node.structure_family == StructureFamily.container
+            and (not original_data_sources)
+        ):
+            _DISPATCH[child_node.structure_family](child_node, node)
+
+
+_DISPATCH = {
```
danielballan, line 108: This covers the data structures used by Bluesky but does not cover `awkward` or `sparse`. It would not be hard to add those...might be worth doing before merging.

tiled/client/sync.py
```diff
@@ -0,0 +1,112 @@
+import collections.abc
+import itertools
+
+from ..structures.core import StructureFamily
+from ..structures.data_source import DataSource, Management
+from .base import BaseClient
+
+
+def sync(
```
genematx, line 9: It might be just me, but when I think about synchronizing two objects, I usually imagine them as equivalent (in terms of source and destination), and whichever object has older data gets updated. I wonder if it would be better to call this function something like "copy" or "transfer"?
Another option could be to define an `.update()` method (abstract method?) on the `BaseClient` itself and subclass it in specific clients; this way, we wouldn't need `_DISPATCH`. It could work as `update` in dictionaries but also recursively descend into catalogs, and this would make it unambiguous, which argument is the source and which which one is the destination.

tiled/client/sync.py
```diff
@@ -0,0 +1,112 @@
...
+    Copy select data.
+
+    >>> sync(a.items().head(), b)
+    >>> sync(a.search(...), b)
+
+    """
+    if hasattr(source, "structure_family"):
+        # looks like a client object
+        _DISPATCH[source.structure_family](source.include_data_sources(), dest)
+    elif isinstance(source, list):
```
genematx, line 41: I feel like this condition could be simplified a little (albeit it might not work with awkward and sparse...):
```
if hasattr(source, "structure_family"):
   _DISPATCH[source.structure_family](source.include_data_sources(), dest)
else:
   _DISPATCH[StructureFamily.container](dict(source), dest)
```

tiled/client/sync.py
```diff
@@ -0,0 +1,112 @@
...
+    for key, child_node in source.items():
+        original_data_sources = child_node.include_data_sources().data_sources()
+        if not original_data_sources:
+            # A container with no data sources is just an organizational
+            # entity in the database.
+            if child_node.structure_family == StructureFamily.container:
+                data_sources = []
+            else:
+                raise ValueError(
+                    f"Unable to copy {child_node} which has is a "
```
genematx, line 72:
```suggestion
                    f"Unable to copy {child_node} which is a "
```

tiled/client/sync.py
```diff
@@ -0,0 +1,112 @@
...
+                )
+        else:
+            (original_data_source,) = original_data_sources
+            if original_data_source.management == Management.external:
+                data_sources = [original_data_source]
+            else:
+                if child_node.structure_family == StructureFamily.container:
+                    data_sources = []
+                else:
+                    data_sources = [
```
genematx, line 83: could `copy.deepcopy` work here? Although, spelling out all the arguments doesn't hurt, i think.

tiled/client/sync.py
```diff
@@ -0,0 +1,112 @@
...
+            # entity in the database.
+            if child_node.structure_family == StructureFamily.container:
+                data_sources = []
+            else:
+                raise ValueError(
+                    f"Unable to copy {child_node} which has is a "
+                    f"{child_node.structure_family} but has no data sources."
+                )
+        else:
+            (original_data_source,) = original_data_sources
```
genematx, line 76: Is it always going to be just a single data source? I think we've made this assumption before, just wondering if it's still okay to do this without any explicit checks.

tiled/client/sync.py
danielballan: Good call!

tiled/client/sync.py
danielballan: Good call. I may this check explicit.

tiled/_tests/test_writing.py
```diff
@@ -483,7 +483,7 @@ def test_write_with_specified_mimetype(tree):
             x.write_partition(df, 0)
             x.read()
             x.refresh()
-            x.data_sources()[0]["mimetype"] == mimetype
+            x.data_sources()[0].mimetype == mimetype
```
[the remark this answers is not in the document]
danielballan: Yikes! Done.

tiled/client/sync.py
danielballan: Renamed to `copy`

tiled/client/sync.py
danielballan: Fixed manually on top of other changes

tiled/client/sync.py
```diff
@@ -0,0 +1,134 @@
...
+            raise NotImplementedError(
+                "Multiple Data Sources in one Node is not supported."
+            )
+        node = dest.new(
+            key=key,
+            structure_family=child_node.structure_family,
+            data_sources=data_sources,
+            metadata=dict(child_node.metadata),
+            specs=child_node.specs,
+        )
```
pbeaucage, line 117:
```suggestion
        try:
            node = dest.new(
                key=key,
                structure_family=child_node.structure_family,
                data_sources=data_sources,
                metadata=dict(child_node.metadata),
                specs=child_node.specs,
            )
        except tiled.client.util.ClientError as err:
            if err.response.status_code == httpx.codes.CONFLICT:
                # print('Skipped existing entry (or UUID conflict)')
                continue
            else:
                raise e
        
```
this is a crude and probably relatively non-performant, but functional way to deal with a user request to copy already-copied items.  also needs two imports:
```
from .utils import ClientError
import httpx 
```

danielballan: I think this comment just missed @genematx’s merge commit.

I can see the argument for a simple "skip conflicts" option. Let's do a little exploration of what options we may want in the future and then start with something simple.

danielballan: Would you make a separate follow-up PR out of this commit, @pbeaucage ?

pbeaucage: No worries!  See #737

PR #737, opened 2024-05-07
Skip duplicate data during copy from one catalog to another

pbeaucage:
Following up from comment on #735, this PR provides extremely rudimentary duplicate skipping during a `tiled.client.sync.copy`.  ~~I think it makes sense to put this behavior behind a flag `skip_duplicates` and leave it False by default.~~  Duplicate handling is a little tricky for performance; I think the scope of `.copy()` is pretty clear that it doesn't update nodes in the destination catalog if the source has changed (though this, of course, might eventually be desirable behavior).  

(edit per @danielballan's suggestion below: this flag is now a str called `on_conflict` with implemented options 'error' (default), 'warn', and 'skip')

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/_tests/test_sync.py
+++ b/tiled/_tests/test_sync.py
@@ -6,6 +6,7 @@ import awkward
 import h5py
 import numpy
 import pandas
+import pytest
 import sparse
 import tifffile
 
@@ -14,6 +15,7 @@ from tiled.client import Context, from_context
 from tiled.client.register import register
 from tiled.client.smoke import read
 from tiled.client.sync import copy
+from tiled.client.utils import ClientError
 from tiled.queries import Key
 from tiled.server.app import build_app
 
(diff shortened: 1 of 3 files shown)


danielballan:
Just one suggestion to lay track for future extensions. Could we make the parameter `on_conflict` and make the value a str or StrEnum with values `skip` and `error` (default)? In the future we might add `warn`, `overwrite`, etc. More discussion is needed, and this can be put off for a future PR, but if we choose an enum instead of a Boolean flag we leave space for extensibility.

tiled/_tests/test_sync.py
```diff
@@ -88,6 +90,35 @@ def test_copy_internal():
...
 
+def test_copy_skip_conflict():
+    with client_factory() as dest:
+        with client_factory() as source:
+            populate_internal(source)
+            copy(source, dest)
+            copy(source, dest, on_conflict="skip")
+            assert list(source) == list(dest)
+            assert list(source["c"]) == list(dest["c"])
+            read(dest, strict=True)
```
pbeaucage, line 101: These assertions are really not germane to this test, a better test case would be to try to copy *different* data in the second copy and assert that the data in `dest` matches the first and not the second (really a test for a future `overwrite` option for `on_conflict`).  This could wait until we have `overwrite`.

danielballan: My two cents: fine to wait.

genematx (approved):

danielballan (approved):

PR #740, opened 2024-05-09
Remove unused dependencies.

danielballan:
Updates to dependencies:
- Some libraries were used in the past but dropped, and never removed from our dependencies list. Example: `cachey`
- Some libraries were formerly direct dependencies but now are only secondary dependencies and should not be listed as direct ones. Example: `click`.
- We used to depend on `watchgod` but now we depend on `watchfiles`. Same author, new project. Things were working anyway because `watchfiles` is a secondary dependency of `fastapi` so it was typically installed anyway. But it is a direct dependency of tiled and should be listed. Also, registration moved from server to client so the dependency is moved.

### Checklist
- [x] Add a Changelog entry


maffettone (approved):

maffettone:
Checked for redundancy or omissions, looks good. Passes CI, ready to merge.

PR #745, opened 2024-05-16
Show auth state in Context repr.

### LINKED ISSUE #739: Context repr with auth info
The `tiled.client.context.Context` object currently has a generic repr:

```py
>>> client.context
<tiled.client.context.Context at 0x7ff6400a7490>
```

I think a good use of this would be to display authentication-related state. The Context handles the API key (if set) _or_ the `tiled.client.auth.TiledAuth` object that manages refresh flow. It's in a good position to display, as applicable, the first eight characters of the API key:

```py
>>> client.context
<Context api_key="abcdef12...">
```

or the username:

```py
>>> client.context
<Context id="dallan">
```

or else:

```py
>>> client.context
<Context unauthenticated>
```

danielballan:
Against a mutli-user server:

```py

In [2]: c = from_profile('nsls2')

In [3]: c.context
Out[3]: <Context authenticated as 'dallan'>

In [4]: c.context.api_key = '<redacted>'

In [5]: c.context
Out[5]: <Context authenticated as 'dallan' with API key '86d32845...'>

In [6]: c.logout()

In [7]: c.context
Out[7]: <Context (unauthenticated)>
```

Against a single-user server:
```py

In [2]: c = from_profile('local', api_key='secret')

In [3]: c.context
Out[3]: <Context authenticated with API key 'secret...'>

In [4]: c.context.api_key = None

In [5]: c.context
Out[5]: <Context (unauthenticated)>
```

Needs unit tests

Closes #739 

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -172,7 +172,9 @@ class Context:
                     ",".join(f"'{identity['id']}'" for identity in whoami["identities"])
                 )
             if self.api_key is not None:
-                auth_info.append(f"with API key '{self.api_key[:min(len(self.api_key)//2, 8)]}...'")
+                auth_info.append(
+                    f"with API key '{self.api_key[:min(len(self.api_key)//2, 8)]}...'"
+                )
         auth_repr = " ".join(auth_info)
         return f"<{type(self).__name__} {auth_repr}>"
 
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -172,7 +172,7 @@ class Context:
                     ",".join(f"'{identity['id']}'" for identity in identities)
                 )
             if self.api_key is not None:
-                auth_info.append(f"with API key '{self.api_key[:8]}...'")
+                auth_info.append(f"with API key '{self.api_key[:min(len(self.api_key)//2, 8)]}...'")
         auth_repr = " ".join(auth_info)
         return f"<{type(self).__name__} {auth_repr}>"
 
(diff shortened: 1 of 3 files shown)


tiled/client/context.py
```diff
@@ -159,6 +159,23 @@ def __init__(
...
+        else:
+            auth_info.append("authenticated")
+            identities = self.whoami()["identities"]
+            if identities:
+                auth_info.append("as")
+                auth_info.append(
+                    ",".join(f"'{identity['id']}'" for identity in identities)
+                )
+            if self.api_key is not None:
+                auth_info.append(f"with API key '{self.api_key[:8]}...'")
```
padraic-shafer, line 175: Presumably API keys are a respectable length, but perhaps just to guard against short keys being leaked in error reports that people post...

```suggestion
                auth_info.append(f"with API key '{self.api_key[:min(len(self.api_key)//2, 8)]}...'")
```

danielballan: Fair point.

Just for info for anyone following along, in a multi-user server, the API keys are only ever generated by the user and are always 32 bits of entropy. It's only in a single-user server context, where the user specified an API key, that this would come up.

danielballan:
I am not sure why the container-canary check is acting up. I re-ran and it failed a second time. I think it's unlikely to be related to this PR, but it might be detecting some upstream issue that is important.

danielballan:
Canary failure was a fluke. Re-ran today; it's fine. Possibly an upstream issue.

Issue #751, opened 2024-05-31
Potentially Insecure Evaluation of User Input in Tiled Array Slicing


hyperrealist:
Tiled uses `eval()` on user input for parsing array slices, which is not ideal.

https://github.com/bluesky/tiled/blob/9059fd0dd803e784e226d91ef76b5c2fb432a552/tiled/server/dependencies.py#L174

danielballan:
Thanks for working on this @hyperrealist.

I want to be clear for anyone reading that there are no known exploitable vulnerabilities in the current codebase. This `regex` input validation:

https://github.com/bluesky/tiled/blob/9059fd0dd803e784e226d91ef76b5c2fb432a552/tiled/server/dependencies.py#L163

rejects malicious input. For example:

```
$ http "https://tiled-demo.blueskyproject.io/api/v1/array/full/generated/small_image?slice=import os;os.system('rm -rf /')"
HTTP/1.1 422 Unprocessable Entity
Connection: keep-alive
Content-Length: 261
Content-Type: application/json
Date: Fri, 31 May 2024 20:02:40 GMT
Server: nginx/1.18.0 (Ubuntu)
Set-Cookie: tiled_csrf=GKH9tESfusL742-9G0XY5Rn8TZvqxLQyKg1gka7-42o; HttpOnly; Path=/; SameSite=lax
server-timing: app;dur=8.2
set-cookie: tiled_csrf=GKH9tESfusL742-9G0XY5Rn8TZvqxLQyKg1gka7-42o; HttpOnly; Path=/; SameSite=lax
x-tiled-request-id: b558539f3ba5d4fa

{
    "detail": [
        {
            "ctx": {
                "pattern": "^[-0-9,:]*$"
            },
            "input": "import os;os.system('rm -rf /')",
            "loc": [
                "query",
                "slice"
            ],
            "msg": "String should match pattern '^[-0-9,:]*$'",
            "type": "string_pattern_mismatch",
            "url": "https://errors.pydantic.dev/2.7/v/string_pattern_mismatch"
        }
    ]
}
```

However, I very much support refactoring to remove `eval` for defense in depth.

danielballan:
Closed by #752

PR #752, opened 2024-05-31
Parse array slices without `eval(<user_input>)`

hyperrealist:
This PR addresses issue #751

### DIFF
--- a/tiled/server/dependencies.py
+++ b/tiled/server/dependencies.py
@@ -165,17 +165,17 @@ def expected_shape(
     return tuple(map(int, expected_shape.split(",")))
 
 
+def np_style_slicer(indices: tuple):
+    return indices[0] if len(indices) == 1 else slice_func(*indices)
+
+
+def parse_slice_str(dim: str):
+    return np_style_slicer(tuple(int(idx) if idx else None for idx in dim.split(":")))
+
+
 def slice_(
-    slice: str = Query(None, pattern=SLICE_REGEX),
+    slice: Optional[str] = Query(None, pattern=SLICE_REGEX),
 ):
     "Specify and parse a block index parameter."
-
-    def np_style_slicer(indices: tuple):
-        return indices[0] if len(indices) == 1 else slice_func(*indices)
-
-    def parse_slice_str(dim: str):
-        return np_style_slicer(
-            tuple(int(idx) if idx else None for idx in dim.split(":"))
-        )
 
     return tuple(parse_slice_str(dim) for dim in (slice or "").split(",") if dim)
--- a/tiled/server/dependencies.py
+++ b/tiled/server/dependencies.py
@@ -163,7 +163,7 @@ def expected_shape(
 
 
 def slice_(
-    slice: str | None = None,
+    slice: Optional[str] = None,
 ):
     "Specify and parse a block index parameter."
 
--- a/tiled/server/dependencies.py
+++ b/tiled/server/dependencies.py
@@ -17,6 +17,9 @@ from .authentication import get_current_principal, get_session_state
 from .core import NoEntry
 from .utils import filter_for_access, record_timing
 
+# saving slice() to rescue after using "slice" for FastAPI dependency injection of slice_(slice: str)
+slice_func = slice
+
 
 @lru_cache(1)
 def get_query_registry():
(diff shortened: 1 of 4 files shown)


danielballan:
Echoing comments from #751 for visiblity:

> Thanks for working on this @hyperrealist.

> I want to be clear for anyone reading that there are no known exploitable vulnerabilities in the current codebase. This `regex` input validation:

https://github.com/bluesky/tiled/blob/9059fd0dd803e784e226d91ef76b5c2fb432a552/tiled/server/dependencies.py#L163

> rejects malicious input. For example:

```
$ http "https://tiled-demo.blueskyproject.io/api/v1/array/full/generated/small_image?slice=import os;os.system('rm -rf /')"
HTTP/1.1 422 Unprocessable Entity
Connection: keep-alive
Content-Length: 261
Content-Type: application/json
Date: Fri, 31 May 2024 20:02:40 GMT
Server: nginx/1.18.0 (Ubuntu)
Set-Cookie: tiled_csrf=GKH9tESfusL742-9G0XY5Rn8TZvqxLQyKg1gka7-42o; HttpOnly; Path=/; SameSite=lax
server-timing: app;dur=8.2
set-cookie: tiled_csrf=GKH9tESfusL742-9G0XY5Rn8TZvqxLQyKg1gka7-42o; HttpOnly; Path=/; SameSite=lax
x-tiled-request-id: b558539f3ba5d4fa

{
    "detail": [
        {
            "ctx": {
                "pattern": "^[-0-9,:]*$"
            },
            "input": "import os;os.system('rm -rf /')",
            "loc": [
                "query",
                "slice"
            ],
            "msg": "String should match pattern '^[-0-9,:]*$'",
            "type": "string_pattern_mismatch",
            "url": "https://errors.pydantic.dev/2.7/v/string_pattern_mismatch"
        }
    ]
}
```

> However, I very much support refactoring to remove `eval` for defense in depth.

tiled/server/dependencies.py
```diff
@@ -160,19 +163,16 @@ def expected_shape(
 
 
 def slice_(
-    slice: str = Query(None, pattern="^[-0-9,:]*$"),
+    slice: Optional[str] = None,
```
danielballan, line 166: I would like to retain a regex pattern. It no longer has a security function, with `eval` removed, but it does have a role in both documenting the acceptable inputs in the OpenAPI spec and providing a detailed error message. (See my example in PR Conversation tab.)

Additionally, it would be useful to test that all the failing examples (typo and malicious) will not match the regex, and therefore should get bounced by FastAPI before they parsed by our code. That is, the `ValueError` or `TypeError` should never actually get raised when used in integration with FastAPI. 

Finally, the pattern I've used here is a little under-specified. It is sufficient to block malicious examples but not sufficient to block malformed (e.g. typo-ed) examples. A more precise regex was drafted [in these lines of a dormant PR](https://github.com/bluesky/tiled/pull/391/files#diff-d41273077415e8ba3ad05d118aaa7167ef40343ad385a84666ea1cdfda8454e7R122-R127). Perhaps that would be reused here (with the `mean` aspect removed for now).

hyperrealist: Thanks for the feedback! I'll write a commit with updates.

padraic-shafer: I agree.  The rest looks good to me.

tiled/server/dependencies.py
```diff
@@ -160,19 +166,16 @@ def expected_shape(
...
-    # functions or expensive arithmetic.
-    return tuple(
-        [
-            eval(f"numpy.s_[{dim!s}]", {"numpy": numpy})
-            for dim in (slice or "").split(",")
-            if dim
-        ]
-    )
+
+    def np_style_slicer(indices: tuple):
```
danielballan, line 173: Might as well define these at global scope, rather than re-defining them on every call. Unless I'm missing something clever happening with closures here.

tiled/server/dependencies.py
```diff
@@ -163,7 +166,7 @@ def expected_shape(
 
 
 def slice_(
-    slice: Optional[str] = None,
+    slice: str = Query(None, pattern=SLICE_REGEX),
```
padraic-shafer, line 169: Optional here indicates to OAS3 that `None` is a valid value.

```suggestion
slice: Optional[str] = Query(None, pattern=SLICE_REGEX),
```

padraic-shafer (commented):
Looks good.  A couple of minor things are noted in the comments.

danielballan:
I rebased on `main` to move this along.

While resolving conflicts in `CHANGELOG.md`, I refined the language to clarity that there was no known exploitable vulnerability, but that this change improves our defense in depth.

PR #755, opened 2024-06-17
Add arrow adapter

### LINKED ISSUE #744: Add an `ArrowAdapter`
We have various Adapters for tabular data formats. The existing Adapters for parquet-formatted and CSV-formatted data have different advantages. Parquet is a binary format with a rich data type system and built-in compression. CSV is a human-readable format that supports appending rows (which parquet does not).

We propose to add support for Arrow-formatted data (sometimes also called "[Feather](https://arrow.apache.org/docs/python/feather.html)") which, like CSV is append-able, but like Parquet is binary and has a rich data type system. Like the CSV Adapter, this should implement `write_partition` and `append_partition`, with the same signatures:

https://github.com/bluesky/tiled/blob/307524d0026ef93b25e246a70869ee195ef0ead9/tiled/adapters/csv.py#L153-L185

We have functions in the codebase for reading and writing Arrow-formatted data because the server and the client _transmit_ data as Arrow by default. 

```py
c['some_table'].read()  # data is sent as Arrow
```

These are the functions:

https://github.com/bluesky/tiled/blob/307524d0026ef93b25e246a70869ee195ef0ead9/tiled/serialization/table.py#L9-L24

What these functions _don't_ need is deal with partitioning (a.k.a. chunking) of rows. To implement `read_partition` and `write_partition` and `append_partition` we'll need to dig deeper into the Arrow IPC (inter-process communication) API to write "row batches" to an existing table, rather than write/read an _entire_ table in a single call.

https://arrow.apache.org/docs/python/ipc.html

skarakuzu:
### Checklist
- [x] Add a Changelog entry

closes #744

### DIFF
--- a/tiled/_tests/adapters/test_arrow.py
+++ b/tiled/_tests/adapters/test_arrow.py
@@ -1,0 +1,71 @@
+import tempfile
+import unittest
+
+import pyarrow as pa
+
+from tiled.adapters.arrow import ArrowAdapter
+from tiled.structures.table import TableStructure
+
+names = ["f0", "f1", "f2"]
+data0 = [
+    pa.array([1, 2, 3, 4, 5]),
+    pa.array(["foo0", "bar0", "baz0", None, "goo0"]),
+    pa.array([True, None, False, True, None]),
+]
+data1 = [
+    pa.array([6, 7, 8, 9, 10, 11, 12]),
+    pa.array(["foo1", "bar1", None, "baz1", "biz", None, "goo"]),
+    pa.array([None, True, True, False, False, None, True]),
+]
+data2 = [pa.array([13, 14]), pa.array(["foo2", "baz2"]), pa.array([False, None])]
+
+
+class TestMyFunction(unittest.TestCase):
+    def setUp(self):
+        data_uri = "file://localhost/" + tempfile.gettempdir()
+        table = pa.Table.from_arrays(data0, names)
+        structure = TableStructure.from_arrow_table(table, npartitions=3)
+        assets = ArrowAdapter.init_storage(data_uri, structure=structure)
+        self.adapter = ArrowAdapter(
+            [asset.data_uri for asset in assets], structure=structure
+        )
+
+    def test_attributes(self):
+        assert self.adapter.structure().columns == names
+        assert self.adapter.structure().npartitions == 3
+
+        assert isinstance(self.adapter.reader_handle[0], pa.ipc.RecordBatchFileReader)
+
+    def test_write_read(self):
+        batch0 = pa.record_batch(data0, names=names)
+        batch1 = pa.record_batch(data1, names=names)
+        batch2 = pa.record_batch(data2, names=names)
+
+        # test writing to a partition and reading it
+        self.adapter.write_partition(batch0, 0)
+        assert pa.Table.from_arrays(data0, names) == pa.Table.from_pandas(
+            self.adapter.read_partition(0)
+        )
+
+        self.adapter.write_partition([batch0, batch1], 1)
+        assert pa.Table.from_batches([batch0, batch1]) == pa.Table.from_pandas(
+            self.adapter.read_partition(1)
+        )
+
+        self.adapter.write_partition([batch0, batch1, batch2], 2)
+        assert pa.Table.from_batches([batch0, batch1, batch2]) == pa.Table.from_pandas(
+            self.adapter.read_partition(2)
+        )
+
+        # test write to all partitions and read all
+        self.adapter.write_partition([batch0, batch1, batch2], 0)
+        self.adapter.write_partition([batch2, batch0, batch1], 1)
+        self.adapter.write_partition([batch1, batch2, batch0], 2)
+
+        assert pa.Table.from_pandas(self.adapter.read()) == pa.Table.from_batches(
+            [batch0, batch1, batch2, batch2, batch0, batch1, batch1, batch2, batch0]
+        )
+
+        # test adapter.read() raises NotImplementedError when there are more than 1 partitions
+        with self.assertRaises(NotImplementedError):
+            self.adapter.write(batch0)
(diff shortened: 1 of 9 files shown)


danielballan:
Summarizing chat from this morning:

- Remove the `dataframe_adapter` property from `arrow.py`. Instead of adding code to `table.py` but it directly in the `read`, `read_partition` (etc.) methods in `arrow.py`.
- Make _two_ classes in `arrow.py`, one addressing the "file" format (supports random access of row batches) and one addressing "stream" format (supports appending row batches to an existing file).

tiled/_tests/adapters/test_arrow.py
```diff
@@ -0,0 +1,69 @@
...
+]
+data1 = [
+    pa.array([6, 7, 8, 9, 10, 11, 12]),
+    pa.array(["foo1", "bar1", None, "baz1", "biz", None, "goo"]),
+    pa.array([None, True, True, False, False, None, True]),
+]
+data2 = [pa.array([13, 14]), pa.array(["foo2", "baz2"]), pa.array([False, None])]
+
+
+class TestArrowAdapter(unittest.TestCase):
```
danielballan, line 23: Since about 2016, scientific Python libraries have generally moved away from `unittest.TestCase` and preferred functional-style test construction. In the modern style, the setup functions becomes a fixture and the `test_` methods become functions, like this:

```py
@pytest.fixture
def adapter():
    # Do setup.
    return ArrowAdapter(...)

def test_attributes(adapter):
    assert adapter.structure().columns == names
    assert self.adapter.structure().npartitions == 3
```

tiled/adapters/arrow.py
```diff
@@ -0,0 +1,322 @@
...
+        -------
+
+        """
+        yield from (
+            (key, ArrayAdapter.from_array(self.read([key])[key].values))
+            for key in self._structure.columns
+        )
+
+    @property
+    def reader_handle(self) -> List[pyarrow.RecordBatchFileReader]:
```
danielballan, line 207: I think this is not actually helpful and that the file-opening code should be moved into `read` and `read_partition` directly. I'll comment more on why below.

I believe this idea came from imitating the approach used in CSV. In the case of CSV we were leveraging the "laziness" (deferred execution) of `dask.dataframe.read_csv` and that does not apply here. We can do something simpler. :-)

tiled/adapters/arrow.py
```diff
@@ -0,0 +1,322 @@
...
+    def read(self, *args: Any, **kwargs: Any) -> pandas.DataFrame:
+        """
+        The concatenated data from given set of partitions as pyarrow table.
+        Parameters
+        ----------
+        Returns
+        -------
+        Returns the concatenated pyarrow table as pandas dataframe.
+        """
+        if any(p is None for p in self.reader_handle):
```
danielballan, line 297: This is inefficient because it tries to open all the files and _then_ checks whether any are missing. In fact, as soon as we find one missing we can stop the loop and raise an error.

tiled/adapters/arrow.py
```diff
@@ -0,0 +1,322 @@
...
+        Parameters
+        ----------
+        partition : the index of the partition to read.
+        fields : optional fields parameter.
+
+        Returns
+        -------
+        The pyarrow table corresponding to a given partition and batch as pandas dataframe.
+        """
+        df = self.reader_handle[partition]
```
danielballan, line 321: This opens all the files and then only uses one. Better to just open the file we need.

tiled/adapters/arrow.py
```diff
@@ -0,0 +1,322 @@
...
+                batches = data
+
+        schema = batches[0].schema
+
+        uri = self._partition_paths[partition]
+
+        with pyarrow.ipc.new_file(uri, schema) as file_writer:
+            for batch in batches:
+                file_writer.write_batch(batch)
+            file_writer.close()
```
danielballan, line 254: This line can be removed because it serves no purpose. The context manager guarantees that the file will be closed:

```py
with pyarrow.ipc.new_file(uri, schema) as file_writer:
    ...
# When the interpreter exits the `with ...` code block, file_write.close() is called.
```

danielballan (commented):
This is on track. I left some suggest regarding the implementation.

tiled/adapters/arrow.py
```diff
@@ -0,0 +1,329 @@
...
+
+        if self.structure().npartitions != 1:
+            raise NotImplementedError
+        uri = self._partition_paths[0]
+
+        with pyarrow.ipc.new_file(uri, schema) as file_writer:
+            for batch in data:
+                file_writer.write_batch(batch)
+
+    def read(self, *args: Any, **kwargs: Any) -> pandas.DataFrame:
```
danielballan, line 298: This currently accepts any positional or keyword arguments and then ignores them. Notice that in CSV, we accept any arguments but we _pass them into another method_ which accepts specific arguments and applies them:

https://github.com/bluesky/tiled/blob/15abbc98df9bfd25f8c713656f664a558136e26f/tiled/adapters/csv.py#L203-L217

It should accept an optional `fields` parameter and apply it if it is not `None`. https://github.com/bluesky/tiled/blob/15abbc98df9bfd25f8c713656f664a558136e26f/tiled/adapters/table.py#L226-L227

danielballan (commented):
I missed some things in my last review.

Also, these two files should be removed:
- `main.py` (empty)
- `test.arrow` (large binary file)

tiled/_tests/adapters/test_arrow.py
```diff
@@ -0,0 +1,70 @@
+import tempfile
```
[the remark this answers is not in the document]
skarakuzu: Added also MyPy check for the `adapter/test_arrow.py` . I was thinking of migrating unit tests in relevant directories in the future and thought that we can type check them too.

danielballan (approved):

PR #758, opened 2024-06-18
Make Tiled compatible with numpy 2.0

### LINKED ISSUE #757: Numpy 2.0.0 breaks Tiled
In the recent [scheduled CI runs](https://github.com/bluesky/tiled/actions/runs/9541880111) Numpy 2.0.0 breaks Tiled. As a temporary measure, PR #756 pins to Numpy <2.0.0, but Tiled should be fully migrated to support Numpy 2.0.0. We also have a dependency on blosc2 that seems to depend on Numpy ABI.

Here is the migration guide: https://numpy.org/devdocs/numpy_2_0_migration_guide.html

danielballan:
### Checklist
- [x] Add a Changelog entry

Closes #757

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 # Changelog
 
+## Unreleased
+
+### Changed
+
+- Minor implementation changes were necessary to make Tiled compatible with
+  Numpy 2.0.
+
 ## v0.1.0b3 (2024-06-04)
 
 ### Added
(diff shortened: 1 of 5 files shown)


danielballan:
Recent releases of [blosc2](https://pypi.org/project/blosc2/) dropped support for Python 3.9 and older, but [numpy](https://pypi.org/project/numpy/) still supports Python 3.9. Thus we see binary incompatibility errors on the 3.9 build.

Our `blosc2` dependency is optional anyway. (Tiled supports a [variety of compression methods](https://blueskyproject.io/tiled/explanations/compression.html#supported-compression-methods) conditional on what is installed in the server and client environments.) I think the pragmatic move here is to make `blosc2` condition on Python version.

Someone stuck in an older Python environment can install `numpy<2` and the latest `blosc2` release that still supports their verison of Python.

hyperrealist (approved):
Looks good to me!

PR #759, opened 2024-06-18
Bump container base image to Python 3.12

danielballan:
### Checklist
- [x] Add a Changelog entry


danielballan:
There is no particular reason to do this _now_; this is just general housekeeping. We may see some performance improvements, as Python 3.12 is generally faster than 3.11.

hyperrealist (approved):
Looks good!

danielballan:
The container canary failed twice in a row, so there may be a real issue there.

danielballan:
Superseded by #781. This problem was canary here was simply that we were not waiting long enough for the container to start.

PR #761, opened 2024-06-19
Add "json_merge_patch" dependency

padraic-shafer:


pyproject.toml
```diff
@@ -65,6 +65,7 @@ all = [
     "canonicaljson",
     "jinja2",
     "jmespath",
+    "json_merge_patch",
```
hyperrealist, line 68: I believe this is a duplicate as `json_merge_patch` and `json-merge-patch` refer to the same pip package. We should remove one of them and adhere to the chosen naming convention.

https://github.com/bluesky/tiled/blob/8c717a3555dc4f129fd1bd20946b9e3ced7602bf/pyproject.toml#L71

pyproject.toml
```diff
@@ -240,6 +244,7 @@ server = [
...
```
hyperrealist, line 247: This one is also a duplicate of `json-merge-patch` in https://github.com/bluesky/tiled/blob/8c717a3555dc4f129fd1bd20946b9e3ced7602bf/pyproject.toml#L250

pyproject.toml
```diff
@@ -178,6 +180,7 @@ minimal-client = [
     "appdirs",
     "entrypoints",
     "httpx >=0.20.0,!=0.23.1",
+    "json_merge_patch",
```
hyperrealist, line 183: Kudos for catching this one! I think we also need `jsonpatch` here.

```suggestion
    "json_merge_patch",
    "jsonpatch",
```

pyproject.toml
```diff
@@ -201,6 +204,7 @@ minimal-server = [
     "canonicaljson",
     "jinja2",
     "jmespath",
+    "json_merge_patch",
```
hyperrealist, line 207: Here too, we need `jsonpatch`.

```suggestion
    "json_merge_patch",
    "jsonpatch",
```

hyperrealist (dismissed):
Thanks for catching this @padraic-shafer! I added some review comments.

danielballan:
I confirmed that [`json-merge-patch` with `-` not `_`](https://pypi.org/project/json-merge-patch/) is the canonical spelling on PyPI. (PyPI does redirects for `_` versus `-`.) I sorted all the deps to reduce the chance of confusion.

PR #762, opened 2024-06-19
Range requests

### LINKED ISSUE #734: Support `Range` requests on `/assets/bytes/{path}`
## Use case

```py
import pandas

c = from_profile('local', api_key='secret')
df = pandas.DataFrame({"A": [1,2,3], "B": [4,5,6], "C": [7,8,9]})
c.write_dataframe(df, key="y")

import duckdb
db = duckdb.connect(':memory:')
db.execute('install httpfs; load httpfs;')
db.execute('SELECT * FROM read_parquet("http://localhost:8000/api/v1/asset/bytes/y?id=1&api_key=secret")')  # uses Range header
```

## Background

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Range
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206
https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

Starlette has postponed consideration of support for this. It might make sense to vendor a copy of this approach for now: https://github.com/encode/starlette/pull/1999

danielballan:
As noted in #734 the motivating use cases here was DuckDB's httpfs, which tries to do partial access of files over HTTP. Given a server with one DataFrame in it:

```sh
tiled serve catalog --temp
```

```py
from tiled.cilent import from_uri
df = pandas.DataFrame({"A": [1,2,3], "B": [4,5,6], "C": [7,8,9]})
c.write_dataframe(df, key="x")
```

DuckDB's httpfs plugin issues requests that use the `Range` header:

```py
import duckdb
db = duckdb.connect(':memory:')
db.execute('install httpfs; load httpfs;')
db.execute('SELECT * FROM read_parquet("http://localhost:8000/api/v1/asset/bytes/x?id=1&api_key=secret")').fetchall()
```

generating these server logs confirming that Partial Access was invoked:

```
[531d1958246ae636] 127.0.0.1:59854 (public) - "HEAD /api/v1/asset/bytes/x?id=1&api_key=secret HTTP/1.1" 405 Method Not Allowed
[1e2951b2eaa982dd] 127.0.0.1:59854 (admin) - "GET /api/v1/asset/bytes/x?id=1&api_key=secret HTTP/1.1" 206 Partial Content
[6a5b9f06acf0a7e5] 127.0.0.1:59854 (admin) - "GET /api/v1/asset/bytes/x?id=1&api_key=secret HTTP/1.1" 206 Partial Content
[79d13fc32a84235e] 127.0.0.1:59854 (admin) - "GET /api/v1/asset/bytes/x?id=1&api_key=secret HTTP/1.1" 206 Partial Content
[90ebff815c556fa1] 127.0.0.1:59866 (admin) - "GET /api/v1/asset/bytes/x?id=1&api_key=secret HTTP/1.1" 206 Partial Content
```

### Checklist
- [x] Add a Changelog entry

Closes #734

### DIFF
--- a/tiled/server/file_response_with_range.py
+++ b/tiled/server/file_response_with_range.py
@@ -27,7 +27,9 @@ class FileResponseWithRange(FileResponse):
         **kwargs,
     ):
         if (range is not None) and (status_code != HTTP_206_PARTIAL_CONTENT):
-            raise RuntimeError(f"Range requests must have a {HTTP_206_PARTIAL_CONTENT} status code.")
+            raise RuntimeError(
+                f"Range requests must have a {HTTP_206_PARTIAL_CONTENT} status code."
+            )
         self.range = range
         super().__init__(path, status_code, *args, **kwargs)
 
--- a/tiled/server/file_response_with_range.py
+++ b/tiled/server/file_response_with_range.py
@@ -1,6 +1,12 @@
-# This is adapted from a closed PR in starlette which was reviewed by
-# a core maintainer but put aside for now in favor of other priorities.
-# https://github.com/encode/starlette/pull/1999
+# This is a variation on starlette's FileRespones that adds support for the
+# 'Range' HTTP header.
+
+# It is adapted from a closed PR in starlette which was reviewed by a core
+# starlette maintainer but put aside for now in favor of other priorities in
+# starlette development. Thus, we implement it here in tiled. If in the future
+# starlette adds support upstream, we should consider refactoring to use that.
+
+# Ref: https://github.com/encode/starlette/pull/1999
 import os
 import stat
 import typing
--- a/tiled/server/file_response_with_range.py
+++ b/tiled/server/file_response_with_range.py
@@ -14,19 +14,20 @@ from starlette.responses import (
     formatdate,
     md5_hexdigest,
 )
+from starlette.status import HTTP_200_OK, HTTP_206_PARTIAL_CONTENT
 
 
 class FileResponseWithRange(FileResponse):
     def __init__(
         self,
         path: typing.Union[str, "os.PathLike[str]"],
-        status_code: int = 200,
+        status_code: int = HTTP_200_OK,
         *args,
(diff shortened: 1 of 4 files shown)


tiled/server/file_response_with_range.py
```diff
@@ -0,0 +1,96 @@
+# This is adapted from a closed PR in starlette which was reviewed by
+# a core maintainer but put aside for now in favor of other priorities.
```
hyperrealist, line 2: Perhaps not put aside anymore?

tiled/server/file_response_with_range.py
```diff
@@ -0,0 +1,96 @@
...
+        if self.stat_result is None:
+            try:
+                stat_result = await anyio.to_thread.run_sync(os.stat, self.path)
+                self.set_stat_headers(stat_result)
+            except FileNotFoundError:
+                raise RuntimeError(f"File at path {self.path} does not exist.")
+            else:
+                mode = stat_result.st_mode
+                if not stat.S_ISREG(mode):
+                    raise RuntimeError(f"File at path {self.path} is not a file.")
```
hyperrealist, line 63: I wouldn't raise this error message verbatim with a straight face 😄

hyperrealist (approved):
Looks good to me, with the caveat that I could only do a shallow review of `file_response_with_range.py` given my inexperience with starlette. But the logic checks out.

I edited `file_response_with_range.py` to replace literal http status codes with `HTTP_200__OK` et. al. from starlette.status.

Also there are few minor comments in the files.

tiled/server/file_response_with_range.py
danielballan: Ah, I meant put aside by starlette's core devs, not by us. I revised the comment to clarify and add a little more  context.

PR #768, opened 2024-07-15
Fix bug with pickling `Context`

danielballan:
Unit test reproduces reported bug:

```
E       AttributeError: 'Context' object has no attribute '_verify'
```

Next commit fixes bug.

### Checklist
- [x] Add a Changelog entry


danielballan:
The user who reported this issue checked out this PR branch and confirmed that it fully resolved their problem.

Issue #648, opened 2024-02-06
Healthchecks and containers


dylanmcreynolds:
Without other configuration, container systems determine that tiled is healthy by waiting on the entrypoint/command to exit. If you're using docker-compose or k8s, you probably want to be more specific.  Your probably want to send http requests to the server instead. 

For k8s, I think you can configure this within the manifest by configuring what url to [fetch from](https://www.paulsblog.dev/how-to-successfully-implement-a-healthcheck-in-docker-compose/).  For docker-compose, I think you need to configure a script to run in the container. The simplest example I can think see is in the [docker-compose](https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck) documentation.

So, I have two ideas in this area for tiled to consider:

1. Incur the massive overhead of installing `curl` in the tiled docker image so make this easier.
2. Add a `healthz` endpoint to the tiled server so that when we see these health requests, we know why, and we can code it to do as little work as we can get away with.

danielballan:
I would merge PRs doing both.

PR #770, opened 2024-07-17
💉 Add health check for hosting containerized tiled instances

Kezzsim:
Kubernetes and other container orchestration platforms require pods that expose themselves as services to prove that they are alive and ready to handle HTTP responses, while this response can be baked into deployment yml, it's considered best practice to actually have the service return something indicating that it is healthy.

This implements the spec defined in the [Kubernetes Readyness](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) Probe section.

This resolves part of issue #648

### DIFF
--- a/tiled/_tests/test_routes.py
+++ b/tiled/_tests/test_routes.py
@@ -5,7 +5,7 @@ from starlette.status import HTTP_200_OK
 from ..server.app import build_app
 
 
-@pytest.mark.parametrize("path", ["/", "/docs"])
+@pytest.mark.parametrize("path", ["/", "/docs", "/healthz"])
 @pytest.mark.asyncio
 async def test_meta_routes(path):
     app = build_app({})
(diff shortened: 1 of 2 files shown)


dylanmcreynolds:
This is great. 

Is there any configuration you have to do in k8s to invoke this, or is it automagic?

There is a `docker-compose.yml` in the root of the tiled repo. Perhaps this is a future PR, but it might be nice to add the health check to it, now that it's possible. Here's an [example](https://medium.com/@saklani1408/configuring-healthcheck-in-docker-compose-3fa6439ee280)

Kezzsim:
Good idea!
Yeah we rushed this PR through on a call so we could deploy to our Rancher instance, the `/healthz` endpoint check is standard in spec and seems to work with any Kubernetes instance out of the box.

dylanmcreynolds (approved):

Issue #733, opened 2024-05-01
Let `Container.write_dataframe` accept a dict


danielballan:
It is convenient that `Container.write_array` accepts a `list`, like:

```
$ tiled serve catalog --temp --api-key=secret
```

```py
In [1]: from tiled.client import from_uri

In [2]: c = from_uri('http://localhost:8000', api_key='secret')

In [3]: c
Out[3]: <Container {}>

In [4]: c.write_array([1,2,3], key='x')
Out[4]: <ArrayClient shape=(3,) chunks=((3,),) dtype=int64>
```

It would be convenient if

```py
c.write_dataframe({'a': [1,2,3], 'b':, [4,5,6]}, key='y')
```

were likewise acceptable. Currently, a `pandas.DataFrame` or `dask.dataframe` object must be passed.

danielballan:
This needs two additions:

1. Add a branch to this conditional that checks for `isinstance(dataframe, dict)` and, if so, attempts `dataframe = pandas.DataFrame(dataframe)`.

https://github.com/bluesky/tiled/blob/07c8925746e977352606432743534158972e6da0/tiled/client/container.py#L943-L946

2. Add a variation on this test that passes the dict `data` to `write_dataframe` instead of the `pandas.DataFrame` `df`.

https://github.com/bluesky/tiled/blob/07c8925746e977352606432743534158972e6da0/tiled/_tests/test_writing.py#L125-L147

PR #771, opened 2024-07-17
Accept Python dict from client "write_dataframe" and TableAdapter

nmaytan:
In ref of #733. 

Both cases insist upon a dataframe-compatible shape. For `write_dataframe`:
```
d = {'a': [1,2,3], 'b': [4,5,6,7]}
c.write_dataframe(d, key='n')
[...]
ArrowInvalid: Column 1 named b expected length 3 but got length 4
```
`pyarrow.lib.Table.validate()` is in the stack trace, and docs say that a pyarrow Table is:

> A collection of top-level named, **equal length** Arrow arrays.

For TableAdapter:
```
d = {'a': [1,2,3,4,5], 'b': [4,5,6,7,8,9]}
tdf = DataFrameAdapter.from_pydict(d, npartitions=1)
[...]
ValueError: An error occurred while calling the from_dict method registered to the pandas backend.
Original Message: All arrays must be of the same length
```
This comes from `dask.dataframe.from_dict`.

Lastly, Dan pointed out that this dict support let us slightly simplify `generated_minimal.py`, and I've confirmed that a simple dict replicates [the example ](https://blueskyproject.io/tiled/how-to/direct-client.html#from-a-service-side-tree-instance)identically.

```
client['C'].read()
Out[4]:
      x    y    z
0   1.0  2.0  3.0
1   1.0  2.0  3.0
2   1.0  2.0  3.0
3   1.0  2.0  3.0
4   1.0  2.0  3.0
..  ...  ...  ...
95  1.0  2.0  3.0
96  1.0  2.0  3.0
97  1.0  2.0  3.0
98  1.0  2.0  3.0
99  1.0  2.0  3.0

[100 rows x 3 columns]
```

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/_tests/test_writing.py
+++ b/tiled/_tests/test_writing.py
@@ -174,6 +174,32 @@ def test_write_dataframe_partitioned(tree):
         assert result.specs == specs
 
 
+# @pytest.mark.filterwarnings(f"ignore:{WARNING_PANDAS_BLOCKS}:DeprecationWarning")
+def test_write_dataframe_dict(tree):
+    with Context.from_app(
+        build_app(tree, validation_registry=validation_registry)
+    ) as context:
+        client = from_context(context)
+
+        data = {f"Column{i}": (1 + i) * numpy.ones(5) for i in range(5)}
+        df = pandas.DataFrame(data)
+        metadata = {"scan_id": 1, "method": "A"}
+        specs = [Spec("SomeSpec")]
+
+        with record_history() as history:
+            client.write_dataframe(data, metadata=metadata, specs=specs)
+        # one request for metadata, one for data
+        assert len(history.requests) == 1 + 1
+
+        results = client.search(Key("scan_id") == 1)
+        result = results.values().first()
+        result_dataframe = result.read()
+
+        pandas.testing.assert_frame_equal(result_dataframe, df)
+        assert result.metadata == metadata
+        assert result.specs == specs
+
+
 @pytest.mark.parametrize(
     "coo",
     [
--- a/tiled/_tests/test_writing.py
+++ b/tiled/_tests/test_writing.py
@@ -174,7 +174,6 @@ def test_write_dataframe_partitioned(tree):
         assert result.specs == specs
 
 
-# @pytest.mark.filterwarnings(f"ignore:{WARNING_PANDAS_BLOCKS}:DeprecationWarning")
 def test_write_dataframe_dict(tree):
     with Context.from_app(
         build_app(tree, validation_registry=validation_registry)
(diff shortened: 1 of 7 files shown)


tiled/adapters/table.py
```diff
@@ -57,6 +57,38 @@ def from_pandas(
             ddf, metadata=metadata, specs=specs, access_policy=access_policy
         )
 
+    @classmethod
+    def from_pydict(
```
danielballan, line 61: Little naming thing: the method name `from_dict` is way more common. (See `awkward`, `pandas`, `polars`.) Arrow's choice of `from_pydict` is a bit quirky and maybe reflects Arrow's perspective as a C++ library that it is interfacing with Python here.

tiled/structures/table.py
```diff
@@ -47,6 +47,15 @@ def from_pandas(cls, df):
         data_uri = B64_ENCODED_PREFIX + schema_b64
         return cls(arrow_schema=data_uri, npartitions=1, columns=list(df.columns))
 
+    @classmethod
+    def from_pydict(cls, d):
```
danielballan, line 51: Same as above: I suggest considering `from_dict` here.

tiled/_tests/test_writing.py
```diff
@@ -174,6 +174,32 @@ def test_write_dataframe_partitioned(tree):
         assert result.specs == specs
 
 
+# @pytest.mark.filterwarnings(f"ignore:{WARNING_PANDAS_BLOCKS}:DeprecationWarning")
```
danielballan, line 177: This should probably be deleted. Was this here to silence an unrelated DeprecationWarning that we should address in a separate PR?

padraic-shafer: I agree. This can be deleted.

This is a remnant of some changes I made a while ago. The DeprecationWarning ended up getting handled in the (deprecated) object cache.  This alternate approach seems to have been accidentally left asa comment.

https://github.com/bluesky/tiled/commit/5aae35a7d3b6a3959263b1df656aeb794bd688c5

nmaytan: I take it we should remove all 4 of these commented items in this file? Want that in a separate PR or here?

tiled/adapters/table.py
nmaytan: Changed and tested in incoming PR

tiled/structures/table.py
nmaytan: Changed and tested in incoming PR

tiled/_tests/test_writing.py
danielballan: Either works.

danielballan: I made a new Issue.

danielballan: https://github.com/bluesky/tiled/issues/772

PR #775, opened 2024-08-08
Rename path argument to "dataset" in hdf5_lookup

jwlodek:
### Checklist
- [X] Add a Changelog entry

Recent ophyd_async change renamed the internal HDF path parameter passed to the stream resource to `dataset` from `path`.

Required to merge this in sync with https://github.com/bluesky/bluesky/pull/1790. 

See: https://github.com/bluesky/ophyd-async/blob/5ac94e283b170b603df42a7d5fe33fd7f2786524/src/ophyd_async/core/_hdf_dataset.py#L64

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Write the date in place of the "Unreleased" in the case a new version is release
 - Make `tiled.client` accept a Python dictionary when fed to `write_dataframe()`.
 - The `generated_minimal` example no longer requires pandas and instead uses a Python dict.
 - Remove unused pytest-warning ignores from `test_writing.py`.
+- Rename argument in `hdf5_lookup` function from `path` to `dataset` to reflect change in `ophyd_async`
 
 ### Fixed
 - A bug in `Context.__getstate__` caused picking to fail if applied twice.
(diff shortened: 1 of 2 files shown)


genematx (approved):
Looks good to me, thanks Jakub! I'd still wait for a comment from @danielballan before merging it in case `path` needs to be changed in any other places that I'm not aware of.

danielballan:
I agree the old name was confusing. This needs a SQL migration to update existing DataSource documents.

danielballan:
To create a migration script:

```
$ python -m tiled.catalog revision -m "Change 'path' to 'dataset' in HDF5 assets"
$ git status
On branch main
Your branch is up to date with 'origin/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

       tiled/catalog/migrations/versions/9cfb45fa2c30_change_path_to_dataset_in_hdf5_assets.py
```

add to the top of this list:

https://github.com/bluesky/tiled/blob/15abbc98df9bfd25f8c713656f664a558136e26f/tiled/catalog/core.py#L7-L17

In the migration file created above, you need to run an `UPDATE` command. Example from another migration:

https://github.com/bluesky/tiled/blob/15abbc98df9bfd25f8c713656f664a558136e26f/tiled/catalog/migrations/versions/e756b9381c14_add_structure_family_to_data_sources.py#L49-L53

danielballan (approved):

PR #777, opened 2024-08-09
ENH: add alembic migration

genematx:
This PR adds an alembic catalog migration script that searches for `path` parameter in Tiled nodes corresponding to hdf5 arrays and changes the key to `dataset`.

This is related to PR #775 and recent changes in the [ophyd-async schema](https://github.com/bluesky/ophyd-async/blob/5ac94e283b170b603df42a7d5fe33fd7f2786524/src/ophyd_async/core/_hdf_dataset.py#L64).

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/catalog/migrations/versions/562203c724c7_change_path_to_dataset_in_hdf5_assets.py
+++ b/tiled/catalog/migrations/versions/562203c724c7_change_path_to_dataset_in_hdf5_assets.py
@@ -1,0 +1,39 @@
+"""Change 'path' to 'dataset' in HDF5 assets
+
+Revision ID: 562203c724c7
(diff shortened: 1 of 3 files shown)


tiled/catalog/migrations/versions/562203c724c7_change_path_to_dataset_in_hdf5_assets.py
```diff
@@ -0,0 +1,47 @@
+"""Change 'path' to 'dataset' in HDF5 assets
+
+Revision ID: 562203c724c7
+Revises: ed3a4223a600
```
danielballan, line 4: This will need an update.

PR #779, opened 2024-08-22
add SQL adapter

skarakuzu:
preliminary start of sql adapter. to be continued ...

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ tiled = "tiled.commandline.main:main"
 all = [
     "adbc_driver_manager",
     "adbc_driver_postgresql",
+    "adbc_driver_sqlite",
     "aiofiles",
     "aiosqlite",
     "alembic",
@@ -201,6 +202,7 @@ minimal-server = [
 server = [
     "adbc_driver_manager",
     "adbc_driver_postgresql",
+    "adbc_driver_sqlite",
     "aiofiles",
     "aiosqlite",
     "alembic",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,9 @@ tiled = "tiled.commandline.main:main"
 
 # This is the union of all optional dependencies.
 all = [
+    "adbc_driver_manager",
+    "adbc_driver_sqlite",
+    "adbc_driver_postgresql",
     "aiofiles",
     "aiosqlite",
     "alembic",
@@ -196,6 +199,9 @@ minimal-server = [
 ]
 # This is the "kichen sink" fully-featured server dependency set.
 server = [
+    "adbc_driver_manager",
+    "adbc_driver_sqlite",
+    "adbc_driver_postgresql",
     "aiofiles",
     "aiosqlite",
     "alembic",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,6 @@ tiled = "tiled.commandline.main:main"
 # This is the union of all optional dependencies.
 all = [
     "adbc_driver_manager",
-    "adbc_driver_sqlite",
     "adbc_driver_postgresql",
     "aiofiles",
     "aiosqlite",
@@ -71,6 +70,7 @@ all = [
     "dask",
     "dask[array]",
     "dask[dataframe]",
+    "duckdb",
     "entrypoints",
     "fastapi",
     "h5netcdf",
@@ -200,7 +200,6 @@ minimal-server = [
 # This is the "kichen sink" fully-featured server dependency set.
 server = [
     "adbc_driver_manager",
-    "adbc_driver_sqlite",
     "adbc_driver_postgresql",
     "aiofiles",
     "aiosqlite",
@@ -215,6 +214,7 @@ server = [
     "dask",
     "dask[array]",
     "dask[dataframe]",
+    "duckdb",
     "fastapi",
     "h5netcdf",
     "h5py",
--- a/tiled/_tests/adapters/test_arrow.py
+++ b/tiled/_tests/adapters/test_arrow.py
@@ -4,6 +4,8 @@ import pyarrow as pa
 import pytest
 
 from tiled.adapters.arrow import ArrowAdapter
+from tiled.structures.core import StructureFamily
+from tiled.structures.data_source import DataSource, Management, Storage
 from tiled.structures.table import TableStructure
 
 names = ["f0", "f1", "f2"]
@@ -26,11 +28,29 @@ data_uri = "file://localhost/" + tempfile.gettempdir()
 
 
 @pytest.fixture
-def adapter() -> ArrowAdapter:
+def data_source_from_init_storage() -> DataSource[TableStructure]:
     table = pa.Table.from_arrays(data0, names)
     structure = TableStructure.from_arrow_table(table, npartitions=3)
-    assets = ArrowAdapter.init_storage(data_uri, structure=structure)
-    return ArrowAdapter([asset.data_uri for asset in assets], structure=structure)
+    data_source = DataSource(
+        management=Management.writable,
+        mimetype="application/vnd.apache.arrow.file",
+        structure_family=StructureFamily.table,
+        structure=structure,
+        assets=[],
+    )
+    storage = Storage(filesystem=data_uri, sql=None)
+    return ArrowAdapter.init_storage(
+        data_source=data_source, storage=storage, path_parts=[]
+    )
+
+
+@pytest.fixture
+def adapter(data_source_from_init_storage: DataSource[TableStructure]) -> ArrowAdapter:
+    data_source = data_source_from_init_storage
+    return ArrowAdapter(
+        [asset.data_uri for asset in data_source.assets],
+        data_source.structure,
+    )
 
 
 def test_attributes(adapter: ArrowAdapter) -> None:
--- a/tiled/_tests/adapters/test_sql.py
+++ b/tiled/_tests/adapters/test_sql.py
@@ -299,6 +299,14 @@ def test_write_read_one_batch_many_part(
         result_read_partition
     )
 
+    # read a specific field
+    result_read = adapter.read_partition(0, fields="f1")
+    assert [*data0[1].tolist(), *data2[1].tolist()] == result_read.tolist()
+    result_read = adapter.read_partition(1, fields="f0")
+    assert [*data1[0].tolist(), *data0[0].tolist()] == result_read.tolist()
+    result_read = adapter.read_partition(2, fields="f2")
+    assert [*data2[2].tolist(), *data1[2].tolist()] == result_read.tolist()
+
 
 @pytest.mark.parametrize(
     "table_name, expected",
--- a/tiled/_tests/adapters/test_sql.py
+++ b/tiled/_tests/adapters/test_sql.py
@@ -122,14 +122,16 @@ async def postgres_uri() -> AsyncGenerator[str, None]:
 def adapter_psql_one_partition(
     data_source_from_init_storage: Callable[[str, int], DataSource[TableStructure]],
     postgres_uri: str,
-) -> SQLAdapter:
+) -> Generator[SQLAdapter, None, None]:
     data_source = data_source_from_init_storage(postgres_uri, 1)
-    return SQLAdapter(
+    adapter = SQLAdapter(
         postgres_uri,
         data_source.structure,
         data_source.parameters["table_name"],
         data_source.parameters["dataset_id"],
     )
+    yield adapter
+    adapter.close()
 
 
 @pytest.fixture
--- a/tiled/_tests/adapters/test_sql.py
+++ b/tiled/_tests/adapters/test_sql.py
@@ -1,6 +1,6 @@
 import os
 from pathlib import Path
-from typing import Generator
+from typing import Callable, Generator
 
 import adbc_driver_sqlite
 import pyarrow as pa
@@ -30,8 +30,8 @@ batch2 = pa.record_batch(data2, names=names)
 
 
 @pytest.fixture
-def data_source_from_init_storage() -> DataSource[TableStructure]:
-    def _data_source_from_init_storage(data_uri):
+def data_source_from_init_storage() -> Callable[[str], DataSource[TableStructure]]:
+    def _data_source_from_init_storage(data_uri: str) -> DataSource[TableStructure]:
         table = pa.Table.from_arrays(data0, names)
         structure = TableStructure.from_arrow_table(table, npartitions=1)
         data_source = DataSource(
@@ -46,12 +46,14 @@ def data_source_from_init_storage() -> DataSource[TableStructure]:
         return SQLAdapter.init_storage(
             data_source=data_source, storage=storage, path_parts=[]
         )
+
     return _data_source_from_init_storage
+
 
 @pytest.fixture
 def adapter_sql(
     tmp_path: Path,
-    data_source_from_init_storage: DataSource[TableStructure],
+    data_source_from_init_storage: Callable[[str], DataSource[TableStructure]],
 ) -> Generator[SQLAdapter, None, None]:
     data_uri = f"sqlite:///{tmp_path}/test.db"
     data_source = data_source_from_init_storage(data_uri)
@@ -114,7 +116,8 @@ def postgres_uri() -> str:
 
 @pytest.fixture
 def adapter_psql(
-    data_source_from_init_storage: DataSource[TableStructure], postgres_uri: str
+    data_source_from_init_storage: Callable[[str], DataSource[TableStructure]],
+    postgres_uri: str,
 ) -> SQLAdapter:
     data_source = data_source_from_init_storage(postgres_uri)
     return SQLAdapter(
--- a/tiled/_tests/adapters/test_sql.py
+++ b/tiled/_tests/adapters/test_sql.py
@@ -1,12 +1,12 @@
 import os
 from pathlib import Path
-from typing import Callable, Generator
+from typing import Any, Callable, Generator, Union
 
 import adbc_driver_sqlite
 import pyarrow as pa
 import pytest
 
-from tiled.adapters.sql import SQLAdapter
+from tiled.adapters.sql import SQLAdapter, check_table_name
 from tiled.structures.core import StructureFamily
 from tiled.structures.data_source import DataSource, Management, Storage
 from tiled.structures.table import TableStructure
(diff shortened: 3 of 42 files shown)


pyproject.toml
```diff
@@ -44,6 +44,9 @@ tiled = "tiled.commandline.main:main"
 
 # This is the union of all optional dependencies.
 all = [
+    "adbc_driver_manager",
```
danielballan, line 47: This section is used when tiled is installed like `pip install "tiled[all]"`. These three should also be added to the section `server`, below, so that they are included when tiled is installed like `pip install "tiled[server]"  # server only`.

danielballan:
Lifecycle:

1. Client declares that it wants to create a new tabular dataset, via a request `POST /api/v1/metadata/my_table`.
2. In "catalog" SQL database, the server adds a row to the `nodes` table with any metadata about this table. _This_ is how the new table is connected to any overall dataset, like Bluesky scan and its Scan ID.
3. Also in the "catalog" SQL database, the server adds a row each to the `data_sources` table and the `assets` table. Together, they describe how to locate where the new data will be saved. The Asset part is very locked down. It has room for the URI of the tabular SQL database: `postgresql://...` and some boilerplate. The DataSource has a freeform area called `parameters`, which can fit any JSON. We can use this to put in dataset-specific details, like the name of the SQL table (`table_name`)---derived from the Arrow schema in this case---and a means of selecting the rows of interest for this new dataset (`dataset_id`).
4. When data is written or read, a `SQLAdapter` object is instantiated inside the server. It is passed information extracted from this DataSource and Asset. So, it can know the `table_name` and the `dataset_id`.

danielballan:
Test script:

```py
import pandas
from tiled.structures.data_source import Asset, DataSource, Management
from tiled.structures.table import TableStructure

client = from_uri("http://localhost:8000", api_key="secret")

df = pandas.DataFrame({"a": [1, 2, 3], "b": [1., 2., 3.]})
structure = TableStructure.from_pandas(df)

x = client.new(
    structure_family=StructureFamily.table,
    data_sources=[
        DataSource(
            management=Management.writable,
            mimetype="application/x-tiled-sql-table",
            structure_family=StructureFamily.table,
	        structure=structure,
            assets=[],
        ),
    ],
    metadata={},
    specs=[],
    key="x",
)
x.write(df)
x.append_partition(df, 0)

# This does not work yet
# x.read()  # calls /table/partition/x?partition=0 adapter.read_partition()
```

tiled/adapters/sql.py
```diff
@@ -0,0 +1,732 @@
...
+        """
+
+        query = f"SELECT * FROM {self.table_name} WHERE dataset_id={self.dataset_id}"
+        self.cur.execute(query)
+        data = self.cur.fetch_arrow_table()
+        self.conn.commit()
+
+        table = data.to_pandas()
+        if fields is not None:
+            return table[fields]
```
genematx, line 255: Would it be more performant to include `fields` in the sql query? Or is it actually the other way around?

tiled/adapters/sql.py
```diff
@@ -0,0 +1,732 @@
...
+        if Path(filepath).is_file() and (not os.access(filepath, os.W_OK)):
+            raise ValueError(f"The path {filepath} exists but is not writable.")
+    else:
+        raise ValueError(f"The directory {directory} does not exist.")
+    return filepath
+
+
+def add_dataset_column(table: pyarrow.Table, dataset_id: int) -> pyarrow.Table:
+    column = dataset_id * numpy.ones(len(table), dtype=numpy.int64)
+    return table.add_column(0, pyarrow.field("dataset_id", pyarrow.int64()), [column])
```
genematx, line 316: Is `0` the number of partition here or something else? If so, should it be passed in the signature?

Also, perhaps, there is a way to generalize this function to something like `add_column` to make it more usable. Not sure if we'll need it, though.
Perhaps, in this specific case, there's no need for a dedicated function at all and this (highly specialized) logic can be just implemented in the calling method (although, having functions is nice in general).

tiled/adapters/sql.py
```diff
@@ -0,0 +1,732 @@
...
+        """
+        yield from (
+            (key, ArrayAdapter.from_array(self.read([key])[key].values))
+            for key in self._structure.columns
+        )
+
+    def append_partition(
+        self,
+        data: Union[List[pyarrow.record_batch], pyarrow.record_batch, pandas.DataFrame],
+        partition: int,
```
genematx, line 213: `partition` is not used

tiled/adapters/arrow.py
```diff
@@ -61,18 +63,28 @@ def metadata(self) -> JSON:
         return self._metadata
 
     @classmethod
-    def init_storage(cls, data_uri: str, structure: TableStructure) -> List[Asset]:
+    def init_storage(
+        cls,
+        storage: Storage,
+        data_source: DataSource[TableStructure],
```
genematx, line 69: Related to my comment on `init_storage` in `SQLAdapter`, I find it a bit surprising that this method both accepts and returns a data_source...

tiled/_tests/adapters/test_sql.py
```diff
@@ -0,0 +1,285 @@
...
+from tiled.adapters.sql import SQLAdapter, check_table_name
+from tiled.structures.core import StructureFamily
+from tiled.structures.data_source import DataSource, Management, Storage
+from tiled.structures.table import TableStructure
+
+names = ["f0", "f1", "f2"]
+data0 = [
+    pa.array([1, 2, 3, 4, 5]),
+    pa.array(["foo0", "bar0", "baz0", None, "goo0"]),
+    pa.array([True, None, False, True, None]),
```
genematx, line 18: Would be nice to see a floating-point type here too.
Do integer arrays allow for missing values?

tiled/_tests/adapters/test_sql.py
```diff
@@ -0,0 +1,285 @@
...
+]
+data2 = [pa.array([13, 14]), pa.array(["foo2", "baz2"]), pa.array([False, None])]
+
+batch0 = pa.record_batch(data0, names=names)
+batch1 = pa.record_batch(data1, names=names)
+batch2 = pa.record_batch(data2, names=names)
+
+
+@pytest.fixture
+def data_source_from_init_storage() -> Callable[[str], DataSource[TableStructure]]:
```
genematx, line 33: Looks like you are defining a (sub-)function in the scofe of the fixture in order to be able to parameterize it. It might be possible to do the same easier with `request.param`, also see [here](https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments:~:text=resource%2Dbased%20ordering.-,Deferring%20the%20setup%20of%20parametrized%20resources,-%C2%B6).

tiled/_tests/adapters/test_sql.py
```diff
@@ -0,0 +1,285 @@
...
+
+def test_write_read_sql_one(adapter_sql: SQLAdapter) -> None:
+    # test writing and reading it
+    adapter_sql.append_partition(batch0, 0)
+    result = adapter_sql.read()
+    # the pandas dataframe gives the last column of the data as 0 and 1 since SQL does not save boolean
+    # so we explicitely convert the last column to boolean for testing purposes
+    result["f2"] = result["f2"].astype("boolean")
+
+    assert pa.Table.from_arrays(data0, names) == pa.Table.from_pandas(result)
```
genematx, line 82: Do you need to convert them to `pa.Table` or could you just compare the arrays themselves?

tiled/_tests/adapters/test_sql.py
```diff
@@ -0,0 +1,285 @@
...
+    data_source = data_source_from_init_storage(postgres_uri)
+    return SQLAdapter(
+        postgres_uri,
+        data_source.structure,
+        data_source.parameters["table_name"],
+        data_source.parameters["dataset_id"],
+    )
+
+
+def test_psql(postgres_uri: str, adapter_psql: SQLAdapter) -> None:
```
genematx, line 131: `postgres_uri` is not used

genematx:
This is great! Thank you @skarakuzu and @danielballan!
Perhaps, adding a few lines in the docs would also be helpful. Specifically, on how would one start a local `--temp` Tiled server backed by an SQLite or DuckDB tabular storage. And the writing examples above are also very useful and could be elevated to the docs, imho.

tiled/client/container.py
```diff
@@ -926,13 +991,50 @@ def write_dataframe(
...
+        """
+        return self._write_dataframe(
+            dataframe,
+            None,  # default format, not appendable
+            key=key,
+            metadata=metadata,
+            specs=specs,
+        )
+
+    def _write_dataframe(
```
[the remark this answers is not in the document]
danielballan: Yes, vestigial.

tiled/client/container.py
```diff
@@ -926,13 +991,50 @@ def write_dataframe(
...
+        metadata : dict, optional
+            User metadata. May be nested. Must contain only basic types
+            (e.g. numbers, strings, lists, dicts) that are JSON-serializable.
+        specs : List[Spec], optional
+            List of names that are used to label that the data and/or metadata
+            conform to some named standard specification.
+
+        See Also
+        --------
+        write_appendable_dataframe
```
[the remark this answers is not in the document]
danielballan: Yes. Should be updated to `create_appendable_table`.

tiled/adapters/sql.py
danielballan: The `0` here means, "Insert this column before first row." As this was only used in one place, it probalby does make sense to just move it inline.

tiled/adapters/sql.py
danielballan: We should validate that this is `0` and raise a `ValueError` if not. (This may be extended in the future once we decide what to do about partitions with SQL storage.)

tiled/adapters/sql.py
```diff
@@ -0,0 +1,732 @@
...
+        """
+        The metadata representing the actual data.
+        Returns
+        -------
+        The metadata representing the actual data.
+        """
+        return self._metadata
+
+    @classmethod
+    def init_storage(
```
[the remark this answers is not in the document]
danielballan: The optional `table_name` is tucked inside `data_source.parameters`. By accepting and then returning (a copy of) the `data_source`, this Adapter method has the opportunity to inject Adapter-specific defaults and normalization before the `data_source` is saved in the catalog database. In this case, it adds a default table name if one is not provided.

tiled/adapters/sql.py
```diff
@@ -0,0 +1,732 @@
...
+    def metadata(self) -> JSON:
+        """
+        The metadata representing the actual data.
+        Returns
+        -------
+        The metadata representing the actual data.
+        """
+        return self._metadata
+
+    @classmethod
```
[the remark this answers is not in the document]
danielballan: I am very much open to exploring other APIs here, but I this this pattern is serving us well. See my answer to

> Should init_storage also accept an optional table_name?

I see "classmethod constructors" as one but not the only good use for `classmethod` in Python.

But again, happy to dream up alternatives.

tiled/structures/data_source.py
```diff
@@ -21,10 +21,13 @@ class Asset:
     id: Optional[int] = None
 
 
+StructureT = TypeVar("StructureT")
+
+
 @dataclasses.dataclass
-class DataSource:
+class DataSource(Generic[StructureT]):
```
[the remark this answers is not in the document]
skarakuzu: typing for `Datasource`  have been implemented on top of your changes due to mypy requirements in the PR and `DataSource[None]` cases handled in each relevant file.

tiled/adapters/sql.py
```diff
@@ -0,0 +1,732 @@
...
+    def init_storage(
+        cls,
+        storage: Storage,
+        data_source: DataSource[TableStructure],
+        path_parts: List[str],
```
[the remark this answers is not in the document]
skarakuzu: I imagine @danielballan can explain this better. I imagine it is a place holder for future multi partitioning plan.

tiled/adapters/sql.py
```diff
@@ -0,0 +1,732 @@
...
+            raise ValueError("The SQL adapter must have only 1 partition")
+
+        schema = (
+            data_source.structure.arrow_schema_decoded
+        )  # based on hash of Arrow schema
+        encoded = schema.serialize()
+
+        default_table_name = "table_" + hashlib.md5(encoded).hexdigest()
+        table_name = data_source.parameters.setdefault("table_name", default_table_name)
+        check_table_name(table_name)
```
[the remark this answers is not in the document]
skarakuzu: Sorry I can not see the difference apart from some space which has not been addressed/complained by pre-commit. I am blind maybe? :D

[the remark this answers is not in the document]
genematx: No difference :)
Just moved the comment and removed linebreaks

tiled/adapters/sql.py
```diff
@@ -0,0 +1,732 @@
...
+        key : a string to indicate which column to be retrieved
+        Returns
+        -------
+        The column for the associated key.
+        """
+        if key not in self.structure().columns:
+            return None
+        return ArrayAdapter.from_array(self.read([key])[key].values)
+
+    def __getitem__(self, key: str) -> ArrayAdapter:
```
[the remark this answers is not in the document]
skarakuzu: You mean inheriting from `TableAdapter` ? I do not see advantage as each adapter has its own personality at this point . If we enter into that business, we need to talk about many changes which may affect all adapters that support Table style. I think it is a design decision which this PR do not need to deal with.

tiled/adapters/sql.py
```diff
@@ -0,0 +1,732 @@
...
+        ----------
+        data : data to append into the database. Can be a list of record batch, or pandas dataframe.
+        partition : the partition index to write.
+        Returns
+        -------
+        """
+        if isinstance(data, pandas.DataFrame):
+            table = pyarrow.Table.from_pandas(data)
+        else:
+            if not isinstance(data, list):
```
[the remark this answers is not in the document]
skarakuzu: Why would it the data be tuple? We enforce it `Union[List[pyarrow.record_batch], pyarrow.record_batch, pandas.DataFrame]`

tiled/adapters/sql.py
skarakuzu: I am not SQL query expert. If you know how to do it with an optional parameters, we can try :)

tiled/_tests/adapters/test_arrow.py
```diff
@@ -26,11 +28,29 @@
...
+    )
+    storage = Storage(filesystem=data_uri, sql=None)
+    return ArrowAdapter.init_storage(
+        data_source=data_source, storage=storage, path_parts=[]
+    )
+
+
+@pytest.fixture
+def adapter(data_source_from_init_storage: DataSource[TableStructure]) -> ArrowAdapter:
+    data_source = data_source_from_init_storage
```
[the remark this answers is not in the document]
skarakuzu: There is not a fixture named `data_source` , I do not get what you mean.

tiled/_tests/adapters/test_sql.py
skarakuzu: `read` method returns the table in the form of pandas. We convert both to pa.Table to check equivalency.

genematx: yeah, I understand. I was just wondering whether converting it to `pa.Tables` might be an overkill when you can just pull the column of interest from pandas as an array and compare arrays, although it probably wouldn't work if you have columns with different dtypes. This is not critical, of course.

tiled/adapters/sql.py
danielballan:
```py
query = "SELECT {', '.join(fields) if fields is not None else '*'} ..."
```

danielballan: Probably should first do...

```py
if not set(fields or []).issubset(self.structure().columns):
    raise ValueError("...")
```

...and think about SQL injection attack protection for maliciously named fields....maybe this is for a follow-up PR. :-)

tiled/adapters/sql.py
danielballan: Done.

tiled/adapters/sql.py
danielballan: Done.

tiled/client/container.py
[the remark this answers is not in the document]
danielballan: In fact it's worse: there was unreachable code here due to a tired but abandoned approach. Conslidated.

tiled/client/container.py
```diff
@@ -942,6 +1044,14 @@ def write_dataframe(
             List of names that are used to label that the data and/or metadata
             conform to some named standard specification.
         """
+        parameters = {}
+        if table_name is not None:
+            if mimetype != "application/x-tiled-sql-table":
+                raise ValueError(
```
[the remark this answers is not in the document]
danielballan: Code was unreachable anyway. Moot.

tiled/client/container.py
```diff
@@ -952,13 +1062,20 @@ def write_dataframe(
...
+
         client = self.new(
             StructureFamily.table,
             [
-                DataSource(structure=structure, structure_family=StructureFamily.table),
+                DataSource(
+                    structure=structure,
+                    structure_family=StructureFamily.table,
+                    mimetype=mimetype,
+                    parameters=parameters,
```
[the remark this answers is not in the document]
danielballan: Done

tiled/adapters/sql.py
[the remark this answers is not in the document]
danielballan: This is how the service communicates to the adapter the path/name of the dataset it is initializing. File-based adapters use this to name the file(s) they create. This adapter ignores it. But it should be present in the signature so that the service can call `init_storage` generically.

tiled/adapters/sql.py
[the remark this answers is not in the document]
genematx: I guess, all I wanted to say, `data` _could_ be anything -- the function itself can not control what's being passed to it. If, by any chance, it's a `tuple` or other iterable, I'd imagine, it shouldn't break, so it would make sense to make this check a bit more general, imho.

tiled/_tests/adapters/test_arrow.py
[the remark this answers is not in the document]
genematx: You can call the fixture `data_source`, instead of calling it `data_source_from_init_storage ` and then immediately assigning `data_source = data_source_from_init_storage`

tiled/_tests/adapters/test_sql.py
danielballan: No.

tiled/adapters/sql.py
```diff
@@ -0,0 +1,797 @@
...
+        Returns
+        -------
+        An iterator for the data in the associated database.
+        """
+        yield from (
+            (key, ArrayAdapter.from_array(self.read([key])[key].values))
+            for key in self._structure.columns
+        )
+
+    def append_partition(
```
genematx, line 239: Just trying to use this adapter to write tabular data. Is there a reason it only has the `append_partition` method, not `write_partition` or `write` like the rest of writable adapters (at least CSV and Parquet adapters seem to have them). I think it would be nice to unify and standardize the interfaces of all writable adapters, similarly to how we have done this with `.from_catalog` and `.from_uris` methods to make it less confusing for the user. @skarakuzu

skarakuzu: It was implemented initially but removed later. If I remember correctly, it is because `write` or `write_partition` means deleting the existing table and starting a new one but we do not want to delete the tables.

genematx: Ah, I see. Thank you for the explaining this, Seher

PR #781, opened 2024-09-04
Address bit rot

danielballan:
Several issues piled up on `main` in a short time, breaking the nightly CI builds:

1. A change in starlette (more below)
2. Compression test failed, due to fragility, easily resolved
3. Container no longer builds due to issue with upgrading pip
4. Container-canary check was flaky because it was not waiting long enough for container to start

Tiled has `PatchedResponse` and `PatchedStreamingResponse`, which subclass `Response` and `StreamingResponse` from starlette to accept `memoryview` as well as `bytes`. This enables numpy buffers to pass into responses zero-copy.

In [Starlette 0.38.0](https://www.starlette.io/release-notes/#0380-july-20-2024) Starlette applied these patches upstream, so the patches are no longer needed.

For reasons I have not dug into, our patch interacts badly with the upstream change, failing a slew of tests in CI:

```
AttributeError: 'PatchedStreamingResponse' object has no attribute 'background'
```

Removing our patch resolves the problem.

### Checklist
- [x] Add a Changelog entry


danielballan:
I am confused by what is going wrong with the container-canary CI workflow. It passes the build phase but fails on validation. Locally, validation is fine:

```
./canary_linux_amd64 validate --debug --file canary-validator.yml abb2a2d2379bb168dcae63819f0bb49e224eb3262b746e986907826c6213510d
Validating abb2a2d2379bb168dcae63819f0bb49e224eb3262b746e986907826c6213510d against tiled
Running container with command 'docker run -d --name canary-runner-72068eac -e TILED_SINGLE_USER_API_KEY=secret -p 8000:8000/TCP abb2a2d2379bb168dcae63819f0bb49e224eb3262b746e986907826c6213510d'
 Exposes an HTTP interface on port 8000             [passed]
 Serves a REST API with an authenticated endpoint /api/v1/search [passed]
 Serves a REST API with a public endpoint /api/v1   [passed]
 Serves a UI at /ui/browse                          [passed]
validation passed
```

CHANGELOG.md
```diff
@@ -3,19 +3,34 @@ Write the date in place of the "Unreleased" in the case a new version is release
...
 - Added an `Arrow` adapter which supports reading/writing arrow tables via `RecordBatchFileReader`/`RecordBatchFileWriter`.
 
 ### Changed
+
 - Make `tiled.client` accept a Python dictionary when fed to `write_dataframe()`.
 - The `generated_minimal` example no longer requires pandas and instead uses a Python dict.
 - Remove unused pytest-warning ignores from `test_writing.py`.
 - Rename argument in `hdf5_lookup` function from `path` to `dataset` to reflect change in `ophyd_async`
 
+## v0.1.0b6 (2024-07-17)
```
danielballan, line 26: Releases for b6 and b7 were made on GH, but the CHANGELOG was not updated.

The release b8 will go out today (I intend) to release the fixes in this PR.

Dockerfile
```diff
@@ -26,7 +26,8 @@ RUN python3 -m venv $VIRTUAL_ENV
 # in ENTRYPOINT or CMD.
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 
-RUN pip install --upgrade --no-cache-dir "cython<3" pip wheel
```
danielballan, line 29: Rationale:
- We needed to pin cython back right after the cython 3.0.0 release because some dependencies had not caught up. This is no longer needed.
- I used to think we should upgrade to latest `pip` and `wheel`, but now I think it's better take whatever the base image gives us. Upgrading was running into weird issues, and I think it's a sign that I made the wrong choice.

genematx (approved):

danielballan:
@Wiebke pin-pointed the starlette PR that created the regression: https://github.com/encode/starlette/pull/2620/files

PR #782, opened 2024-09-05
Datetime units

genematx:
This expands `BuiltinDtype` by adding explicit units for numpy `datetime64` and `timedelta64` dtypes, e.g. `datetime64[ns]`. This is required, for unambiguous representation of 8-byte datetime objects, specifically with zarr.

The `BuiltinDtype` class now defines an additional attribute, `units`, which is `""` by default. The units are extracted from (and added to) the string representation of `numpy.dtype`, e.g. `<M8[ns]`.

More information of supported units can be found [here](https://numpy.org/doc/stable/reference/arrays.datetime.html#:~:text=the%20rule%20%27same_kind%27-,Datetime%20units,-%23).

### Checklist
- [x] Add a Changelog entry


danielballan:
@joshmoore @MSanKeys963 If you have a moment I wonder if you could comment on plans for datetime data types in Zarr v3. At SciPy 2023 I remarked to Josh that they appeared to be missing from the [spec](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#core-data-types) and I think he suggested that they might be reintroduced. It does seem to be common in xarray examples, particularly climate examples.

The term `datetime` appears on that linked page once, but in the context of a "(currently made up) extension data type".

joshmoore:
@danielballan: thanks for the ping. Unfortunately, I don't have any updates on `datetime` in Zarr v3. As you can imagine, the focus is still heavily on getting zarr-python v3 out the door, but I imagine there would be a lot of support for moving it forward. (I pointed to this PR and https://github.com/bluesky/tiled/pull/774 at a recent dev meeting.)

tiled/server/pydantic_array.py
```diff
@@ -39,10 +41,18 @@ class BuiltinDtype(BaseModel):
 
     @classmethod
     def from_numpy_dtype(cls, dtype) -> "BuiltinDtype":
+        # Extract datetime units from the dtype string representation,
+        # e.g. `'<M8[ns]'` has `units = 'ns'`.
+        units = ""
+        if dtype.kind in ("m", "M"):
+            if res := re.search(r"\[(.+)\]$", dtype.str):
```
tacaswell, line 48: Surely there is a better way than running a regex on the string version of the dtype ?!

tiled/server/pydantic_array.py
```diff
@@ -27,6 +28,7 @@ class BuiltinDtype(BaseModel):
     endianness: Endianness
     kind: Kind
     itemsize: int
+    units: str = ""
```
tacaswell, line 31: Would this be better done as `None` as the missing sentinel?  I worry about the in-band encoding of "no units because no one has said anything about what my units are, be warned they cloud be ANYTHING" as an empty string vs "I have units and the are an empty string because I am a dimensions value like the Rayleigh number" as an empty string.

tiled/server/pydantic_array.py
```diff
@@ -60,14 +70,16 @@ def to_numpy_str(self):
...
+        units = f"[{self.units}]" if self.units else ""
+        return f"{endianness}{self.kind.value}{size}{units}"
 
     @classmethod
     def from_json(cls, structure):
         return cls(
             kind=Kind(structure["kind"]),
             itemsize=structure["itemsize"],
             endianness=Endianness(structure["endianness"]),
+            units=structure.get("units", "s"),
```
tacaswell, line 82: defaulting to seconds?

genematx: this must be a weird typo, as I was just copy-pasting this part `structures/array.py`. Changing it to empty string; thanks for catching this!

tiled/server/pydantic_array.py
genematx: Thank you, I really like the idea! Changed the default to `None`.

tiled/server/pydantic_array.py
genematx: Changed the default units to `None`, so this is not relevant anymore.

tiled/server/pydantic_array.py
tacaswell: If we can not dig this out, I think it would be better to go with `units = {dtype_obj: 'string', ..}[dtype]` where we directly map the numpy dtype to the string we want for it.  A bit of boilerplate for us, but it is clear what is going on and protects us from most of the things I can think of that numpy could change that would mess with us (and it does not user their str repr as a critcal API).

tacaswell: This would mean we could also pick some units _not_ to support https://numpy.org/doc/stable/reference/arrays.datetime.html#datetime-units

genematx: Turns out, there is a helper [utility function](https://numpy.org/doc/stable/reference/generated/numpy.datetime_data.html) in numpy to extract the units from `datetime` dtypes, and the units can be even more complicated. Thank you, @danielballan for pointing this out! This is probably the cleanest solution, and I hope numpy wouldn't change this api. I don't think we need to keep the units (strings) and counts (int) separately, though -- a compound attribute is enough -- but keen to hear your thoughts on this.

tacaswell: That is something I would expect them to keep stable.

tacaswell (approved):

danielballan (approved):

PR #783, opened 2024-09-06
ENH: Add data type to sparse

genematx:
### Checklist
- [ ] Add a Changelog entry


danielballan (approved):
I had to think about how this ever worked in the first place! Since we move the data in Arrow IPC format by default, the data types come free. Clearly you are correct: we should be writing down the data type.

Let's merge after some third person has chance to kick the tires and confirm interactively that there are no surprises here.

PR #786, opened 2024-09-18
🔑->🗑️ Enable Tiled Admins to delete API Keys belonging to other user's principals

### LINKED ISSUE #766: Add route for Tiled admins to revoke API key
We have `POST /api/v1/auth/principal/{uuid}/apikey`, guarded with `admin:apikeys` scope, enabling Tiled admins to mint API keys on behalf of others users and service principals. We should have corresponding `DELETE` route that Tiled admins can use to revoke any API key.

Kezzsim:
### Checklist
- [ ] Add a Changelog entry

Resolves #766 enhancement request by providing a way for admins to delete keys belonging to other users via principal, through the `api/v1/auth/principal/{principal_uuid}/apikey?first_eight={first_eight}'` endpoint.

### DIFF
--- a/tiled/_tests/test_authentication.py
+++ b/tiled/_tests/test_authentication.py
@@ -655,30 +655,32 @@ def test_api_key_bypass_scopes(enter_password, principals_context):
                     ).raise_for_status()
 
 
-@pytest.mark.parametrize(
-    "username, scopes, resource",
-    (
-        ("alice", ["read:principals"], "/api/v1/auth/principal"),
-        ("bob", ["read:data"], "/api/v1/array/full/A1"),
-    ),
-)
 def test_admin_delete_principal_apikey(
-    enter_password, principals_context, username, scopes, resource
+    enter_password,
+    principals_context,
 ):
     """
     Admin can delete API keys for any prinicipal, revoking access.
     """
     with principals_context["context"] as context:
-        # Log in as Alice
+        # Log in as Bob (Ordinary user)
+        with enter_password("secret2"):
+            context.authenticate(username="bob")
+
+        # Create an ordinary user API Key
+        principal_uuid = principals_context["uuid"]["bob"]
+        api_key_info = context.create_api_key(scopes=["read:data"])
+        context.logout()
+
+        # Log in as Alice (Admin)
         with enter_password("secret1"):
             context.authenticate(username="alice")
 
-        # Create the API Key
-        principal_uuid = principals_context["uuid"][username]
-        api_key_info = context.admin.create_api_key(principal_uuid, scopes=scopes)
         # Delete the created API Key via service principal
-        context.admin.revoke_api_key(
-            principal_uuid, api_key_info["first_eight"]
-        )
+        context.admin.revoke_api_key(principal_uuid, api_key_info["first_eight"])
+        context.logout()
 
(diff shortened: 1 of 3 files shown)


tiled/client/context.py
```diff
@@ -837,6 +837,25 @@ def create_service_principal(
             )
         ).json()
 
+    def delete_service_principal_apikey(self, uuid, first_eight=None):
```
danielballan, line 840: We have:

```py
c.context.create_api_key  # Create API key for currently-authenticated Principal
c.context.admin.create_api_key  # Create API key for any Principal (requires `admin:apikeys` scope)
```

The `admin` namespace is just an "accessor", a usability tweak to put methods that require admin scopes out of the path of ordinary users so that they do not clutter `c.context.<TAB>`. It's implemented as an `Admin` class defined in this module.

We also have:

```py
c.context.revoke_api_key  # Remove API key for currently-authenticated Principal
```

and I propose we add this method to the `Admin` class, similarly, name it `revoke_api_key`, so we have

```py
c.context.revoke_api_key  # Remove API key for currently-authenticated Principal
c.context.admin.revoke_api_key  # Remove API key for any Principal (requires `admin:apikeys` scope)
```

The docstring should also be broadened: this method and associated HTTP endpoint will in fact allow revoking _any_ API key. It is not restricted to Service Principals.

tiled/server/authentication.py
```diff
@@ -912,6 +912,35 @@ async def principal(
     )
 
 
+@base_authentication_router.delete(
+    "/principal/{uuid}/apikey",
+    response_model=schemas.Principal,
+)
+async def delete_service_principal_apikey(
```
danielballan, line 919: We have methods named:

````
new_apikey
revoke_apikey
apikey_for_principal
```

so maybe this should be `revoke_apikey_for_principal`? Certainly the word `service` shouldn't be in here because it can be used to revoke other users' API keys as well, as the docstring correctly says.

tiled/_tests/test_authentication.py
```diff
@@ -653,3 +653,32 @@ def test_api_key_bypass_scopes(enter_password, principals_context):
...
+    with principals_context["context"] as context:
+        # Log in as Alice
+        with enter_password("secret1"):
+            context.authenticate(username="alice")
+
+        # Create the API Key
+        principal_uuid = principals_context["uuid"][username]
+        api_key_info = context.admin.create_api_key(principal_uuid, scopes=scopes)
+        # Delete the created API Key via service principal
+        context.admin.delete_service_principal_apikey(
```
danielballan, line 680: Here, `alice` is deleting her own API key. A fuller test would do:

1. Bob, a user, creates an API key. It works.
2. Alice, an admin, revokes Bob's API key.
3. Bob tries to use the API key. It fails.

danielballan (changes requested):
I have some naming suggestions, for consistency with existing methods, and some thoughts on expanding the test.

tiled/client/context.py
Kezzsim: Resolved in db0084ad67b89cfde25b2733e00980fbe321be27
The admin `revoke_api_key` command has been refactored for easier legibility

danielballan (approved):

PR #794, opened 2024-10-08
🥌 Add cURL to dockerfile: good for healthchecks

Kezzsim:
Quick and easy last thing to wrap up #648 , a neat thing to have in the container as pointed out by @dylanmcreynolds

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,9 @@ FROM python:3.12-slim as builder
 # We need gcc to compile thriftpy2, a secondary dependency.
 RUN apt-get -y update && apt-get install -y git gcc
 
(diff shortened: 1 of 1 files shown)


danielballan:
While we're here, would you add `httpie` too? Nicer usage for exploring modern apps. But I think it's worth keeping `curl` as well because everyone knows it.

danielballan:
The docs CI failure is, as you'd imagine, unrelated. Fix: https://github.com/bluesky/tiled/pull/795

danielballan:
Power-cycled to re-run CI against fixed `main` branch.

danielballan:
A new docs-building failure! Unrelated to PR. Merging.

PR #796, opened 2024-10-15
Add jpeg adapter, support for image/jpeg mimetype

jwlodek:
### Checklist
- [x] Add a Changelog entry
 
Sister PR to https://github.com/bluesky/bluesky/pull/1816. Tested in the lab with a real camera. ~~Still needs added tests~~, and currently requires source install of ophyd-async branch that includes AD JPEG writer.

Edit: Tests added and all passing now.

### DIFF
--- a/tiled/adapters/jpeg.py
+++ b/tiled/adapters/jpeg.py
@@ -1,0 +1,293 @@
+import builtins
+from typing import Any, Dict, List, Optional, Tuple, cast
+
+import numpy as np
+from numpy._typing import NDArray
+
+from ..structures.array import ArrayStructure, BuiltinDtype
+from ..structures.core import Spec, StructureFamily
+from ..utils import path_from_uri
+from .protocols import AccessPolicy
+from .resource_cache import with_resource_cache
+from .type_alliases import JSON, NDSlice
+from PIL import Image
+
+
+class JPEGAdapter:
+    """
+    Read a JPEG file.
+
+    Examples
+    --------
+
+    >>> JPEGAdapter("path/to/file.jpeg")
+    """
+
+    structure_family = StructureFamily.array
+
+    def __init__(
+        self,
+        data_uri: str,
+        *,
+        structure: Optional[ArrayStructure] = None,
+        metadata: Optional[JSON] = None,
+        specs: Optional[List[Spec]] = None,
+        access_policy: Optional[AccessPolicy] = None,
+    ) -> None:
+        """
+
+        Parameters
+        ----------
+        data_uri :
+        structure :
+        metadata :
+        specs :
+        access_policy :
+        """
+        if not isinstance(data_uri, str):
+            raise Exception
+        filepath = path_from_uri(data_uri)
+        cache_key = (Image.open, filepath)
+        self._file = with_resource_cache(cache_key, Image.open, filepath)
+        self.specs = specs or []
+        self._provided_metadata = metadata or {}
+        self.access_policy = access_policy
+        if structure is None:
+            if self._file.is_shaped:
+                from_file: Tuple[Dict[str, Any], ...] = cast(
+                    Tuple[Dict[str, Any], ...], self._file.shaped_metadata
+                )
+                shape = tuple(from_file[0]["shape"])
+            else:
+                arr = np.asarray(self._file)
+                shape = arr.shape
+            structure = ArrayStructure(
+                shape=shape,
+                chunks=tuple((dim,) for dim in shape),
+                data_type=BuiltinDtype.from_numpy_dtype(self._file.series[0].dtype),
+            )
+        self._structure = structure
+
+    def metadata(self) -> JSON:
+        """
+
+        Returns
+        -------
+
+        """
+        # This contains some enums, but Python's built-in JSON serializer
+        # handles them fine (converting  to str or int as appropriate).
+        d = {tag.name: tag.value for tag in self._file.pages[0].tags.values()}
+        d.update(self._provided_metadata)
+        return d
+
+    def read(self, slice: Optional[NDSlice] = None) -> NDArray[Any]:
+        """
+
+        Parameters
+        ----------
+        slice :
+
+        Returns
+        -------
+
+        """
+        # TODO Is there support for reading less than the whole array
+        # if we only want a slice? I do not think that is possible with a
+        # single-page TIFF but I'm not sure. Certainly it *is* possible for
+        # multi-page TIFFs.
+        arr = np.asarray(self._file)
+        if slice is not None:
+            arr = arr[slice]
+        return arr
+
+    def read_block(
+        self, block: Tuple[int, ...], slice: Optional[slice] = None
+    ) -> NDArray[Any]:
+        """
+
+        Parameters
+        ----------
+        block :
+        slice :
+
+        Returns
+        -------
+
+        """
+        # For simplicity, this adapter always treat a single TIFF file as one
+        # chunk. This could be relaxed in the future.
+        if sum(block) != 0:
+            raise IndexError(block)
+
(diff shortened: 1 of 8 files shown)


jwlodek:
```python
In [25]: run['primary']['external']['manta-cam1'].metadata
Out[25]: 
{'data_key': 'manta-cam1',
 'mimetype': 'multipart/related;type=image/jpeg',
 'parameters': {'chunk_shape': [1, 544, 728],
                'template': 'proj_6ffba08d-9416-472c-974d-fbe5a1b1fcdd_{:06d}.jpeg'},
 'run_start': '0d1ab7fc-3e95-40ac-96df-0dd4820af8a7',
 'uid': '986d0ee3-1528-4e75-86c3-14f4feaba9fe',
 'uri': 'file://localhost/nsls2/data/tst/legacy/mock-proposals/2024-1/pass-000000/assets/'}

In [26]: run['primary']['external']['manta-cam1'][7]
Out[26]: 
array([[2, 2, 2, ..., 0, 0, 0],
       [2, 2, 2, ..., 0, 0, 0],
       [2, 2, 2, ..., 0, 0, 0],
       ...,
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0]], dtype=uint8)

In [27]:
```

tiled/client/register.py
```diff
@@ -340,11 +340,23 @@ async def register_single_item(
...
-async def group_tiff_sequences(
+IMG_SEQUENCE_STEM_PATTERNS = {
+    ".tif": re.compile(r"^(.*?)(\d+)\.(?:tif|tiff)$"),
+    ".tiff": re.compile(r"^(.*?)(\d+)\.(?:tif|tiff)$"),
+    ".jpg": re.compile(r"^(.*?)(\d+)\.(?:jpg|jpeg)$"),
+    ".jpeg": re.compile(r"^(.*?)(\d+)\.(?:jpg|jpeg)$"),
+}
+IMG_SEQUENCE_EMPTY_NAME_ROOT = "_unnamed"
+
+IMG_SEQUENCE_MIMETYPES = {
```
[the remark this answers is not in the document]
jwlodek: I am not aware of any of our beamlines writing pngs, I think a compressed format but not as compressed as jpeg is probably of limited use - if they want actual data they'd use tiff, if they want a snapshot where they don't care if data is compressed, jpeg is probably fine

tiled/adapters/jpeg.py
```diff
@@ -0,0 +1,279 @@
...
+
+from ..structures.array import ArrayStructure, BuiltinDtype
+from ..structures.core import Spec, StructureFamily
+from ..utils import path_from_uri
+from .protocols import AccessPolicy
+from .resource_cache import with_resource_cache
+from .type_alliases import JSON, NDSlice
+
+
+class JPEGAdapter:
```
genematx, line 17: I wonder if there's a chance to define base classes, something like `ImageAdapter` and `ImageSequenceAdapter` and then subclass them for jpeg and tiff? What we will need to specify when subclassing are the supported mimetypes, and file reading functions (for a single file and for the sequence) -- otherwise the code looks almost identical.
Happy to help with restructuring it!

tiled/adapters/jpeg.py
```diff
@@ -0,0 +1,279 @@
...
+        specs :
+        access_policy :
+        """
+        self._seq = seq
+        # TODO Check shape, chunks against reality.
+        self.specs = specs or []
+        self._provided_metadata = metadata or {}
+        self.access_policy = access_policy
+        if structure is None:
+            shape = (len(self._seq), *self.read(slice=0).shape)
```
genematx, line 184: I ran into this problem yesterday, and it took me some time to figure out what was going on... During the init time, `read()` method may still be partially undefined, for example if it accesses `self.structure()`. I know it's not the case right now, but this limits our ability to extend it later (unfortunately, in a very non-obvious way).

tiled/adapters/jpeg.py
genematx: Yes, that was what I thought too, first; for tiffs and jpegs alone, just passing a callable should be enough, but this may not extend to any other formats (or if there's need for some custom logic) in the future (this is partly why I was asking about pngs). I think, I'd be in favor of keeping seperate `JPEGAdapter` and `TiffAdapter` classes, even if they are very short.

tiled/adapters/jpeg.py
```diff
@@ -0,0 +1,279 @@
...
+        self._structure = structure
+
+    def metadata(self) -> JSON:
+        """
+
+        Returns
+        -------
+
+        """
+        return {}
```
genematx, line 72:
```suggestion
        return self._provided_metadata.copy()
```

tiled/adapters/jpeg.py
```diff
@@ -0,0 +1,279 @@
...
+        self._seq = seq
+        # TODO Check shape, chunks against reality.
+        self.specs = specs or []
+        self._provided_metadata = metadata or {}
+        self.access_policy = access_policy
+        if structure is None:
+            shape = (len(self._seq), *self.read(slice=0).shape)
+            structure = ArrayStructure(
+                shape=shape,
+                # one chunks per underlying TIFF file
```
genematx, line 187:
```suggestion
                # one chunks per underlying JPEG file
```
(although, unnecessary if we move this part to `FileSequenceAdapter`)

tiled/adapters/sequence.py
```diff
@@ -0,0 +1,200 @@
...
+
+        Returns
+        -------
+            Return a numpy array
+        """
+        if slice is Ellipsis:
+            arr = self._load_from_files()
+        elif isinstance(slice, int):
+            # e.g. read(slice=0) -- return an entire image (drop 0th dimension of the stack)
+            arr = self._load_from_files(slice)[-1]
```
danielballan, line 147:
```suggestion
            arr = numpy.squeeze(self._load_from_files(slice), 0)
```

Using `squeeze` is slightly better because, if something unexpected happens, it will not silently drop data.

tiled/adapters/sequence.py
```diff
@@ -0,0 +1,200 @@
...
+            if len(slice) == 0:
+                arr = self._load_from_files()
+            elif len(slice) == 1:
+                arr = self.read(slice=slice[0])
+            else:
+                left_axis, *the_rest = slice
+                # Could be int or slice (i, ...) or (slice(...), ...); the_rest is converted to a list
+                if isinstance(left_axis, int):
+                    # e.g. read(slice=(0, ....))
+                    arr = self._load_from_files(left_axis)[-1]
```
danielballan, line 161:
```suggestion
                    arr = numpy.squeeze(self._load_from_files(slice), 0)
```

Same as above.

tiled/adapters/tiff.py
```diff
@@ -135,160 +134,6 @@ def structure(self) -> ArrayStructure:
...
-    def structure(self) -> ArrayStructure:
-        """
-
-        Returns
-        -------
-
-        """
-        return self._structure
+class TiffSequenceAdapter(FileSequenceAdapter):
+    def _load_from_files(self, slc: Union[slice, int] = slice(None)) -> NDArray[Any]:
```
danielballan, line 138: It is convention to use `slice` as the parameter name. We use `builtins.slice` to avoid a clash. Arguable, but might as well be consistent.

danielballan (approved):
I suggested some minor refactors, but this is great @jwlodek! Really nice extension of existing patterns. Thanks for guiding this along, @genematx.

tiled/adapters/tiff.py
genematx: Fixed!

PR #797, opened 2024-10-16
Force reshape tiff data by the adapter

genematx:
### Context
Inconsistency between the expected (by the client) and the actual data shapes is a common cause of server-side errors. Barring the cases of actually erroneous metadata, the shape of the array can often be corrected by presenting a reshaped numpy view, i.e. without any modification or copying of the underlying data.

For example, multichannel images may have the dimensionality of `n_channels x height x width`, which for monochrome images reduced to `1 x height x width`; the extra singleton dimension is specified in the Bluesky descriptor document is expected by the client. However, the data may be read from the external file by the I/O library as a 2D array, `height x width`, therefore resulting in a shape mismatch. Currently this raises a 500 Server Error.

While the descriptors can be modified for any acquisition in the future, changing the existing documents in the database is undesired.

### Solution
This PR adds shape validation to the corresponding Adapters on the server side, which have the information about the expected shape in their `._structure().shape` attribute (so far, only for `TiffSequenceAdapter`, but it can be extended to other classes too). If the expected and actual shape differ only in missing or extra singleton dimensions, the returned array is reshaped accordingly. In future, this can be also modified to handle cases of swapped or transposed dimensions, but these will require more rigorous checks.

### Reference
This modification is necessitated by the ongoing migration of existing Bluesky documents from Mongo storage to an SQL catalog.

### Checklist
- [x] Add a Changelog entry


tiled/adapters/utils.py
```diff
@@ -1,15 +1,83 @@
...
 from ..utils import node_repr as tree_repr  # noqa: F401
+from .type_alliases import NDSlice
 
 _MESSAGE = (
     "Instead of {name}_indexer[...] use {name}()[...]. "
     "The {name}_indexer accessor is deprecated."
 )
 
 
+def sliced_shape(shp: Tuple[int, ...], slc: Optional[NDSlice] = ...) -> Tuple[int, ...]:
```
danielballan, line 17:
```py
In [1]: from ndindex import ndindex

In [2]: sliced_shape = ndindex(slice(0, 5)).newshape((10, 10))
```

danielballan: Note: we aleady have an `ndindex` dependency as we use it for exactly this situation, so it is a good bet.

danielballan (commented):
This can go in once the suggested refactor is implemented. Nice improvement!

jwlodek (approved):

tiled/adapters/utils.py
genematx: Fixed!

danielballan (approved):

danielballan (approved):

PR #800, opened 2024-10-25
Issue799

### LINKED ISSUE #799: tiled[minimal-server] import issue
For testing code that uses a tiled client, I use a pytest fixture that build a server app.

For my tests, I import tiled[minimal-server].

For the code
```python
@pytest.fixture
def app(catalog):
    app = build_app(catalog)
    yield app

```

I get the error:

```
src/_test/conftest.py:21: in <module>
    from tiled.server.app import build_app
/usr/local/Caskroom/miniconda/base/envs/arroyo/lib/python3.11/site-packages/tiled/server/app.py:37: in <module>
    from ..authenticators import Mode
/usr/local/Caskroom/miniconda/base/envs/arroyo/lib/python3.11/site-packages/tiled/authenticators.py:14: in <module>
    from .server.authentication import Mode
/usr/local/Caskroom/miniconda/base/envs/arroyo/lib/python3.11/site-packages/tiled/server/authentication.py:64: in <module>
    from .core import DEFAULT_PAGE_SIZE, MAX_PAGE_SIZE, json_or_msgpack
/usr/local/Caskroom/miniconda/base/envs/arroyo/lib/python3.11/site-packages/tiled/server/core.py:42: in <module>
    register_builtin_serializers()
/usr/local/Caskroom/miniconda/base/envs/arroyo/lib/python3.11/site-packages/tiled/serialization/__init__.py:26: in register_builtin_serializers
    from ..serialization import sparse as _sparse  # noqa: F401
/usr/local/Caskroom/miniconda/base/envs/arroyo/lib/python3.11/site-packages/tiled/serialization/sparse.py:27: in <module>
    from .table import (
E   ImportError: cannot import name 'serialize_excel' from 'tiled.serialization.table' (/usr/local/Caskroom/miniconda/base/envs/arroyo/lib/python3.11/site-packages/tiled/serialization/table.py)
```

I looked into this and I see the problem. 

[sparse.py](https://github.com/bluesky/tiled/blob/915b6292333cd791d6805875a52624d198e10277/tiled/serialization/sparse.py#L24) will import `serialize_excel` if `pandas` and `pyarow` are installed.

However, [table.py](https://github.com/bluesky/tiled/blob/915b6292333cd791d6805875a52624d198e10277/tiled/serialization/table.py#L70) will only define `serialize_excel` if `openpyxl` and `pandas` are installed.

If `pandas` and `pyarraow` are installed but not `openpyxl`, then`serialize_excel` never got defined. This is the case for me, having [installed](https://github.com/bluesky/tiled/blob/915b6292333cd791d6805875a52624d198e10277/pyproject.toml#L192_)`tiled[minimal-server]'

I see two possible ways to fix this, I'm not sure what's best:

1. Add `openpyxl` to minimal-server, making is slightly less minimal.
2. Change `sparse.py' to add 'openpyxl' to it's check. Something like this?

```
if modules_available("pandas", "pyarrow"):
    import pandas

    from .table import (
        APACHE_ARROW_FILE_MIME_TYPE,
        XLSX_MIME_TYPE,
        deserialize_arrow,
        serialize_arrow,
        serialize_csv,
        # serialize_excel.    <----  import below with additional check
        serialize_html,
        serialize_parquet,
    )
    if modules_avaialble("openpyxl"):
       from .table import serialize_excel

```

dylanmcreynolds:
### Checklist
- [x ] Add a Changelog entry
- [x ] Add the ticket number which this PR closes to the comment section

Fixes #799 

Add a check for the `openpyxcl` module when importing the excel serializer.

This was found when trying to run the `tiled[minimal-server]` installation. A true test of this would  involve creating an environment with just this setup. I think this is something that pixi makes easier, but would be a big change to current CI. 

A valid question might be: is `minimal-server` ever used and could it be excised?


tiled/catalog/adapter.py
```diff
@@ -1330,6 +1330,7 @@ def in_memory(
         access_policy=access_policy,
         writable_storage=writable_storage,
         readable_storage=readable_storage,
+        init_if_not_exists=True,
```
danielballan, line 1333: This is not necessary. When the database is in memory, the tables are _always_ initialized because by definition this is a new database in process memory. This keyword argument has the effect of spawning a subprocess to initialize a database, which cannot help in this case.

danielballan (commented):
Stray commit should be removed (and I think is not helpful anyway).

Yes, it might make sense to remove `minimal-server`. Worth polling some users.

tiled/catalog/adapter.py
dylanmcreynolds: Removed. Oops.

danielballan (approved):

PR #803, opened 2024-10-29
MNT: Refactoring changes to CSV adapter + CSVArrayAdapter

genematx:
* Allow `CSVAdapter` to accept kwargs for `pd.read_csv` (e.g. separator)
*  Remove `dataframe_adapter` property from `CSVAdapter`
* Allow simple instantiation with either single uri or a list of uris
* Specify `multipart/related;type=text/csv` mimetype
* Experimental: Add a new `CSVArrayAdaper` backed by an `ArrayAdapter` (instead of `TableAdapter`). It can be used to load homogeneous numerical arrays stored as scv files. The distinction between the two is intended to be done by the mimetype: "text/csv;header=present" -- for tables, and "text/csv;header=absent" -- for arrays.
* Unify the classmethod interfaces across the internal adapters (with `from_assets` and `from_uris` being the two primary methods).

### Checklist
- [x] Add a Changelog entry


danielballan:
I see two structural changes here:

1. In `main`, we have registries that map mimetypes to any callable that returns an Adapter. In this PR, the registries map must map to an Adapter class. These Adapter classes adhere to an interface with several methods, used for various functions:
    i. "I want to create a new empty dataset of this format, to write into it." (`init_storage`)
    ii. "I found some files in this format that I want introspect, to register them in my catalog database." (`from_uris`)
    iii. "I previously introspected or wrote some files, and now I want to access them based on this information from my catalog database." (`from_assets`)
  As before, `__init__`, the "single privileged entrypoint", expects only exactly what it needs, and its signature varies from class to class.

2. There is an additional change involved in (iii), `from_assets`. In this PR, it expects a flat list of Asset objects, which may point to:
    i. Files of mixed types and roles, such as TIFFs in an ordered sequence mixed with "sidecar" JSON or YAML metadata files
    ii. Files that must be tracked by the database for accounting purposes but are actually not used directly by the Adapter (e.g. HDF5 "data" files in a virtual dataset)

Some time ago, we encountered these examples and some others in the wild. We did not like the look of these Adapters:
(a) Their signatures did not explain to the caller what was expected: just "list of URIs".
(b) The Adapter code had to assign roles, distinguish scalars for lists, potentially sort list. It seemed wasteful and error-prone. When we introspect the files (ii) we _know_ this information. We should write it down at that time and use it!

Thus, we added additional columns to the Asset--DataSource many-to-many relation table which encode which _role_ and _order_ each Asset plays. With each row in the many-to-many relation table we also store, in addition to `asset_id` and and `data_source_id`:

parameter | num | What to do
-- | -- | --
`<name>` | `NULL` | `Adapter.from_assets(<name>=asset.data_uri)`
`<name>` | `<int>` | `Adapter.from_assets(<name>=[asset.data_uri, ...])`
`NULL` | `NULL` | `Adapter.from_assets(...)`  # do not include asset in parameters
`NULL` | `<int>` | Disallowed by database triggers

---

I **really like** having standard separate paths for introspection (ii) and construction from DataSource/Asset(iii). That's a huge improvement.

I see in the implementation of `from_assets` a regression that (a) puts us back to homogenous signatures (`assets: list[Assets]`) that make diverse hard requirements and (b) leaves the Adapters redoing work (and perhaps making mistakes of missing errors) that was already done and recorded when the data was registered or written, as the case may be.

---

_Now on shakier ground, just brainstorming..._

Can we get the best of both worlds? As a starting point, convenience function like this might work:

```py
def from_node_and_data_source(adapter_cls: type[Adapter], node: Node, data_source: DataSource) -> Adapter
    "Usage example: from_data_source(CSVAdapter, node, data_source)"
    adapter_cls = adapter_cls or ADAPTERS_BY_MIMETYPE
    parameters = defaultdict(list)
    for asset in data_source.assets:
        if asset.num is None:
            # This asset is associated with a parameter that takes a single URI.
            parameters[asset.parameter] = asset.data_uri
        else:
            # This asset is associated with a parameter that takes a list of URIs.
            parameters[asset.parameter].append(asset.data_uri)
    return adapter_cls(metadata=node.metadata, specs=node.specs, structure=data_source.structure, **parameters)
```

(A further convenience could wrap this and figure out the `adapter_cls` automatically by looking up `data_source.mimetype` in an `adapter_by_mimetype` registry, i.e. `f(node, data_source) -> Adapter`.)

Thus, we would drop the `from_assets` constructor in favor of using the explicit `cls.__init__`, which declares in its signature precisely what it expects: whether it be one URI, a list of URIs, or multiple different URIs/lists with different formats and/or roles to play.

We would retain two great aspects of this PR:
- an easy way to construct Adapters from data_sources and their assets
- a clear separation between between the introspection/registration path and the "from assets" path

There may be better implementations of these goals available, but I hope this is a promising example.

danielballan:
Two minor weak points in both the implementation on `main` and my sketch above:

1. The namespace of `parameters` is shared with the namespace of Tiled-reserved terms like `metadata`, `structure`, and `specs`. Echoing @tacaswell from the other data, when "user" (user = Adapter author) and framework names share a namespace, that's often painful. OTOH, having clean function signatures that describe their expected inputs is nice! And the set of reserved terms is unlikely to change much.
2. The names we use for parameters right now---often `data_uri` or `data_uris`---are not very descriptive. Their utility becomes more obvious in these special cases like HDF5 with "master" and "data" files or arrangements of mixed formats, like TIFF+YAML. Maybe we should migrate to more specific names for parameters, like `"image_file"`?

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ## Unreleased
 
+### Added
+
+- Refactor CSVAdapter to allow pd.read_csv kwargs
```
danielballan, line 10: It may be worth mentioning standardizing the Adapter interface and adding `from_catalog`.

tiled/catalog/adapter.py
```diff
@@ -480,18 +476,13 @@ async def get_adapter(self):
...
-        adapter_kwargs = dict(parameters)
-        adapter_kwargs.update(data_source.parameters)
-        adapter_kwargs["specs"] = self.node.specs
-        adapter_kwargs["metadata"] = self.node.metadata_
-        adapter_kwargs["structure"] = data_source.structure
-        adapter_kwargs["access_policy"] = self.access_policy
         adapter = await anyio.to_thread.run_sync(
-            partial(adapter_factory, **adapter_kwargs)
+            partial(
+                adapter_class.from_catalog,
```
danielballan, line 481: To cleanly separate the `parameters` namespace from arguments used by Tiled itself, should we consider using [positional-only parameters](https://peps.python.org/pep-0570/) in the `init_adapter_from_catalog` and the `from_catalog` constructors?

```py
def init_adapter_from_catalog(cls, data_source, node, /, **kwargs):
    ...
```

I don't actually foresee collisions here, but I somewhat like the idea of keeping the namespaces formally separate.

We have yet to introduce these in Tiled (or in any Bluesky project AFAIK) but this is exactly the use case. They are available in all supported versions of Python.

tiled/client/register.py
```diff
@@ -296,10 +296,12 @@ async def register_single_item(
         )
         unhandled_items.append(item)
         return
-    adapter_factory = settings.adapters_by_mimetype[mimetype]
+    adapter_class = settings.adapters_by_mimetype[mimetype]
     logger.info("    Resolved mimetype '%s' with adapter for '%s'", mimetype, item)
     try:
-        adapter = await anyio.to_thread.run_sync(adapter_factory, ensure_uri(item))
+        adapter = await anyio.to_thread.run_sync(
+            adapter_class.from_uris, [ensure_uri(item)]
```
danielballan, line 303: I recall we floated the idea of using variadic args here, so that the usage is consistent for one URI or several, but we can avoid the slightly awkward boxing of a single item in a list.

```py
@classmethod
def from_uris(cls, data_uri, **kwargs):
    ...
```

and

```py
@classmethod
def from_uris(cls, *data_uris, **kwargs):
    ...
```

What do you think?

danielballan (commented):
It's coming together! A couple comments on fine-tuning the interfaces.

CHANGELOG.md
genematx: Done

tiled/catalog/adapter.py
genematx: I think this makes total sense. I don't think I ever use them as `from_catalog(data_source=data_source, node=node)`, but I'll doublecheck.
Do you think `data_source` should take precedence? Or `node`?

tiled/client/register.py
genematx: I like it; it's so much cleaner. I need to think of any potential edge cases though.

genematx: The only difficulty that I see with this is that in `from_catalog` we pass either `str` or `list[str]` for `data_uri(s)`. In `from_uris`, it would be just a sequence of `str` arguments, which is a bit of inconsistency.

genematx: Done

tiled/catalog/adapter.py
danielballan: What do you mean by precedence?

genematx: I just meant `from_catalog(data_source, node, /, **kwargs)` or `from_catalog(node, data_source, /, **kwargs)`. Which one makes more sense? It seemed to me that `data_source` is more "important", because node just supplies metadata and specs, but I'm happy to switch the order.

genematx: Done. Made `data_source` and `node` positional only.

danielballan: Gotcha. Yes, I agree with that reasoning.

danielballan (approved):
I took this for a spin and failed to find any regressions.

I have so minor cleanup for the Zarr adapter (no fixes or changes in behavior, just tightening up) that I 'll put up in a separate PR.

PR #810, opened 2024-11-04
Do not require SQL URIs to be prefixed with SQLAlchemy driver

danielballan:
In many places, we require SQL database URIs to include the "driver" like:

```
postgresql+asyncpg://user:password@localhost:5432/database
sqlite+aiosqlite:////test.db
```

This is a SQLAlchemy idiom that is not standard for SQL database URIs generally. In this PR, Tiled now accepts standard URIs like:

```
postgresql://user:password@localhost:5432/database
sqlite:////test.db
```

It still accepts URIs with a driver specified.

These commits are cherry-picked from #779 into a separate PR.
### Checklist
- [x] Add a Changelog entry
- [ ] ~Add the ticket number which this PR closes to the comment section~

### DIFF
--- a/tiled/utils.py
+++ b/tiled/utils.py
@@ -740,7 +740,7 @@ def ensure_specified_sql_driver(uri: str) -> str:
     'postgresql+my_custom_driver://...' -> 'postgresql+my_custom_driver://...'
     '/path/to/file.db' -> 'sqlite+aiosqlite:////path/to/file.db'
     """
-    if not SCHEME_PATTERN.match(uri):
+    if not SCHEME_PATTERN.match(str(uri)):
         # Interpret URI as filepath.
         uri = f"sqlite+aiosqlite:///{Path(uri)}"
     scheme, rest = uri.split(":", 1)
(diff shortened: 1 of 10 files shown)


tiled/utils.py
```diff
@@ -721,6 +721,33 @@ def ensure_uri(uri_or_path) -> str:
...
+    'postgresql://...' -> 'postgresql+asynpg://...'
+    'sqlite://...' -> 'sqlite+aiosqlite://...'
+    'postgresql+asyncpg://...' -> 'postgresql+asynpg://...'
+    'postgresql+my_custom_driver://...' -> 'postgresql+my_custom_driver://...'
+    '/path/to/file.db' -> 'sqlite+aiosqlite:////path/to/file.db'
+    """
+    if not SCHEME_PATTERN.match(str(uri)):
+        # Interpret URI as filepath.
+        uri = f"sqlite+aiosqlite:///{Path(uri)}"
+    scheme, rest = uri.split(":", 1)
```
dylanmcreynolds, line 746: Did you not use `urllib` to parse because in theory this could be URI that's not a URL?

dylanmcreynolds (approved):

tiled/utils.py
danielballan: I remember reaching for `urllib` first. I believe the issue was that certain SQLite URIs do not round-trip quite right.

PR #816, opened 2024-11-20
feat: tty free login

### LINKED ISSUE #815: Make authentication prompt hookable
How we can able GUIs to use the Tiled Python client and authenticate with HTTP basic auth into OAuth2? Currently, Tiled prompts for a username and password using TTY input.

These lines

https://github.com/bluesky/tiled/blob/2459624dcf53e87e94299b439cf10b1d009f4e0a/tiled/client/context.py#L558-L559

can be factored into a global function:

```py
def prompt_for_credentials(username=None) -> (username, password):
     "Accept optional default username, return username and password."
    ...
```

The `Context.__init__` can grow an optional argument, default `None` that enables overriding this function. The constructor functions `from_uri` and `from_profile` can grow the same argument and pass it through.

We should also consider what might be needed to support login via a browser, where Tiled is providing a URL and polling (device code flow) for the user to complete login and authorization. This does not need to be solved in the same PR, but we have a sense of what the interface will be so that the API proposed above is consistent with that.

maffettone:
Closes #815 

## Adds
Hook to send password directly to authenticate. Not added to `login` since `context.authenticate` is meant for programatic use, and we probably shouldn't encourage general users to type their password without `getpass` protections. Tests are added to mirror auth tests. 
`prompt_for_reauthentication` is expected as True for the hook to function, and the `CannotPrompt` error is updated to reflect that. Alternatively, this could be updated to use `if not prompt_for_reauthentication and password is not UNSET`. 

## Fixes
Idempotency in context manager used by tests to avoid TTY input for context manager. (It was used in all tests previously, so idempotentcy was not relevant).

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/_tests/test_authentication.py
+++ b/tiled/_tests/test_authentication.py
@@ -96,37 +96,39 @@ def test_password_auth(enter_username_password, config):
 
 def test_password_auth_hook(config):
     with Context.from_app(build_app_from_config(config)) as context:
-        # Passing prompt_for_reauthentication=False should raise
-        with pytest.raises(CannotPrompt):
-            context.authenticate(
-                username="alice", password="secret1", prompt_for_reauthentication=False
-            )
-
         # Log in as Alice.
-        context.authenticate(
-            username="alice", password="secret1", prompt_for_reauthentication=True
-        )
+        context.authenticate(username="alice", password="secret1")
         assert "authenticated as 'alice'" in repr(context)
-        context.logout()
+
+        # Attempting to reauth without a prompt hook should raise.
+        with pytest.raises(CannotPrompt):
+            context.http_client.auth.sync_clear_token("refresh_token")
+            context.http_client.auth.sync_clear_token("access_token")
+            context.authenticate(username="alice")
 
         # Log in as Bob.
-        context.authenticate(
-            username="bob", password="secret2", prompt_for_reauthentication=True
-        )
+        context.authenticate(username="bob", password="secret2")
         assert "authenticated as 'bob'" in repr(context)
         context.logout()
 
         # Bob's password should not work for Alice.
         with fail_with_status_code(HTTP_401_UNAUTHORIZED):
-            context.authenticate(
-                username="alice", password="secret2", prompt_for_reauthentication=True
-            )
+            context.authenticate(username="alice", password="secret2")
 
         # Empty password should not work.
         with fail_with_status_code(HTTP_401_UNAUTHORIZED):
-            context.authenticate(
-                username="alice", password="", prompt_for_reauthentication=True
-            )
+            context.authenticate(username="alice", password="")
+
+        # Hook for reauthenticating as Alice should succeeed
+        context.authenticate(username="alice", password="secret1")
+        assert "authenticated as 'alice'" in repr(context)
+        context.http_client.auth.sync_clear_token("refresh_token")
+        context.http_client.auth.sync_clear_token("access_token")
+        context.authenticate(
+            username="alice",
+            prompt_for_reauthentication=lambda u, p: ("alice", "secret1"),
+        )
+        assert "authenticated as 'alice'" in repr(context)
 
 
 def test_logout(enter_username_password, config, tmpdir):
--- a/tiled/_tests/test_authentication.py
+++ b/tiled/_tests/test_authentication.py
@@ -95,6 +95,7 @@ def test_password_auth(enter_username_password, config):
 
 
 def test_password_auth_hook(config):
+    """Verify behavior with user-defined 'prompt_for_reauthentication' hook."""
     with Context.from_app(build_app_from_config(config)) as context:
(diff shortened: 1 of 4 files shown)


tiled/_tests/test_authentication.py
```diff
@@ -94,6 +94,41 @@ def test_password_auth(enter_username_password, config):
                 from_context(context, username="alice")
 
 
+def test_password_auth_hook(config):
+    with Context.from_app(build_app_from_config(config)) as context:
+        # Passing prompt_for_reauthentication=False should riase an
```
danielballan, line 99:
```suggestion
        # Passing prompt_for_reauthentication=False should raise
```

danielballan: Unrelated to my copy-editing above, I have questions about the desired behavior.

I had in mind that `context.authenticate(username, password, prompt_for_reauthication=False)` should _not_ raise immediately, but rather configure the `context` to raise _later_ if the session expires and fresh credentials are required. The idea was that the graphical application would catch this error.

Reflecting on this now, I am questioning whether exceptions are the best signaling mechanism. Philosophically, is not an "exceptional" situation: it's a very foreseeable one. And pragmatically, forcing the application to carry every client interaction inside a `try:` is inelegant.

What if we had these options:
- `True` -> current behavior
- `False` -> raise `CannotPrompt` if creds expire
- `callable` -> callback hook that graphical application can use to pop up a dialog or something for new creds

tiled/_tests/utils.py
```diff
@@ -55,22 +54,22 @@ async def temp_postgres(uri):
...
-    original_getusername = context.prompt_for_username
-    original_getpass = getpass.getpass
-    context.PROMPT_FOR_REAUTHENTICATION = True
-    context.prompt_for_username = lambda u: username
-    setattr(getpass, "getpass", lambda: password)
-    yield
-    setattr(getpass, "getpass", original_getpass)
-    context.PROMPT_FOR_REAUTHENTICATION = original_prompt
-    context.prompt_for_username = original_getusername
+    try:
```
danielballan, line 64: Good catch!

tiled/_tests/test_authentication.py
maffettone: This makes more sense. I'll implement something to that effect. I'll want to chat through the workflow a little bit to understand when/how reauth is normally triggered, but I think I can get the right bones in place.

maffettone: I followed through on this in https://github.com/bluesky/tiled/pull/816/commits/a9e97c0eca17d0cf89ece5925173b3cb90a79631.


And updated the tests in 
https://github.com/bluesky/tiled/pull/816/commits/877ff19b335d0b88ca34c9812adc4151af6220e4

Worth noting that this removes the raising of `CannotRefreshAuthentication`, just removes the `http_client.auth`, and allows that to be caught at `CannotPrompt`.

maffettone:
@danielballan green and ready for a another pass

danielballan (approved):

danielballan:
Looks good to me but let's get one more pair of eyes on this -- requested review from @genematx

tiled/client/context.py
```diff
@@ -23,17 +24,21 @@
...
 
 
-def prompt_for_username(username):
+def prompt_for_credentials(username, password):
     """
     Utility function that displays a username prompt.
     """
-    if username:
+    if username is not None and password is not None:
+        # If both are provided, return them as-is, without prompting.
```
padraic-shafer, line 32:
```suggestion
        # If both are provided, return them as-is, without prompting.
        # This is particularly useful for GUI clients without a TTY Console.
```

tiled/_tests/test_authentication.py
```diff
@@ -94,6 +94,43 @@ def test_password_auth(enter_username_password, config):
                 from_context(context, username="alice")
 
 
+def test_password_auth_hook(config):
+    with Context.from_app(build_app_from_config(config)) as context:
```
padraic-shafer, line 98:
```suggestion
def test_password_auth_hook(config):
    """Verify behavior with user-defined 'prompt_for_reauthentication' hook."""
    with Context.from_app(build_app_from_config(config)) as context:
```

tiled/_tests/utils.py
```diff
@@ -55,22 +54,22 @@ async def temp_postgres(uri):
...
+        original_prompt = context.PROMPT_FOR_REAUTHENTICATION
+        original_credentials = context.prompt_for_credentials
```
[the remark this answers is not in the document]
maffettone: Sorted in https://github.com/bluesky/tiled/pull/816/commits/39f1d0deee4a2f18756df1a6155e3b8c632f054e
Went a step further to remove anything from the try block beyond the yield, making it clear that the only raise should come from the calling routine.

genematx (approved):

PR #817, opened 2024-11-21
📂🆙 Switch from appdirs to platformdirs

### LINKED ISSUE #743: Switch from `appdirs` to `platformdirs`
See https://github.com/intake/intake/issues/806 which brought this to my attention.

Kezzsim:
Resolves #743 which is of the same name. A self explanatory upgrade to get around the depreciation of `appdirs`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,6 @@ all = [
     "aiosqlite",
     "alembic",
     "anyio",
-    "platformdirs",
     "asgi-correlation-id",
     "asyncpg",
     "awkward >=2.4.3",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -46,7 +46,7 @@ all = [
     "aiosqlite",
     "alembic",
     "anyio",
-    "appdirs",
+    "platformdirs",
```
danielballan, line 49: The deps are sorted alphabetically in each section. Would you move these accordingly?

danielballan:
I confirmed that `appdirs` is scrubbed from the codebase:

```
$ git grep appdirs
$
```

danielballan (commented):
One minor bit of housekeeping, then this is good to go.

danielballan (approved):

PR #821, opened 2024-11-22
Move curl and httpie to final layer in dockerfile

dylanmcreynolds:
### Checklist
- [x] Add a Changelog entry

#794 attempted to add `curl` and `httpie` to the docker image as a tool to be used in health checks. However, it was added to the `builder` layer, not the `runner` layer and hence not available at run time. This PR moves this installation.

Before this change:

```
docker exec -it ta991 bash
root@16a473d8c0e1:/deploy# curl
bash: curl: command not found
root@16a473d8c0e1:/deploy# 
```

After:
```
docker exec -it a901 bash
root@a90185b67888:/deploy# curl
curl: try 'curl --help' or 'curl --manual' for more information
root@a90185b67888:/deploy# 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ## Unreleased
 
+- Fix curl and httpie installation in docker image.
+
 ### Added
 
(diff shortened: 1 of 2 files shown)


danielballan:
I should have tested that before merging. Heads up @Kezzsim.

PR #822, opened 2024-12-02
Feat: client driven timeout

### LINKED ISSUE #819: Enable client to specify shorter token expiry
Quoting the docs:

> You may configure:
> - The maximum time to keep an inactive session alive (default is one week)
> - The maximum session lifetime (default is unlimited)
> - Access token lifetime. This is internal from the user’s point of view It should be short. (Default 15 minutes.)

This are global server-side configuration settings. Should we enable the client to request a session with different parameters? The use case is, on my personal laptop, I want Tiled to me logged in so I am not re-prompted every time I restart a Jupyter kernel or an IPython session. On a beamline machine, I might want Tiled to quickly time out my session if I step away. This should be specified at time of login, when the credentials are presented.

maffettone:
Give the client a chance to manually reduce their refresh token lifetime. Example being, I want to auth but know I will only be active for 2 hours before leaving the beamline; this provides a secondary safety mechanism should I forget to logout. 

Closes #819 

### Checklist
- [ ] Add a Changelog entry
- [ x ] Add the ticket number which this PR closes to the comment section

### DIFF
--- a/tiled/_tests/test_authentication.py
+++ b/tiled/_tests/test_authentication.py
@@ -7,6 +7,7 @@ import time
 
 import numpy
 import pytest
+from click import prompt
 from starlette.status import HTTP_400_BAD_REQUEST, HTTP_401_UNAUTHORIZED
 
 from ..adapters.array import ArrayAdapter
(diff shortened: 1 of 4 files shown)


danielballan:
@Kezzsim Would you review and test-drive this?

Kezzsim:
This works with a few caveats:

1. I've had to increase the wait time in the tests in commit 149efe2 because they fail when run together on CI
2. Using the `toy_authentication` config with an ipython environment, this is not intuitive to test. Unlike in the [Tiled example docs](https://blueskyproject.io/tiled/tutorials/login.html#login-tutorial), `from_uri` prompts the user for username and password synchronously instead of being able to invoke `c.login(username, password, refresh_token_max_age)` after declaration. There is no way that I know of to pass the `refresh_token_max_age` to `from_uri` and test it.
3. Might it be more useful to also have this set some sort of timeout server side to match? I can still use
```
curl -H "Content-type: multipart/form-data" \                                  
     -F username=alice \
     -F password=secret \
     -F refresh_token_max_age="1" \                                                
     -X POST \
     http://localhost:8000/api/v1/auth/provider/toy/token
```
to log in and then 
```
curl -i http://localhost:8000/api/v1/metadata/ \                               
  -H "Authorization: Basic eyJhbGc..."
```
to get data in perpituity.

PR #823, opened 2024-12-02
Fix: Regression due to httpx updates

genematx:
Updates of `httpx` to version 0.28.0 break the existing pagination handling code (and likely other functionality too) where a url is constructed using query parameters that are passed as both, explicitly as part of the url and as kwargs to the `httpx.client`. This patch fixes this by refactoring the query parameters only as kwargs.

Furthermore, `httpx.AsyncClient` no longer accepts `app` as a kwarg; instead, the application needs to be wrapped into an `ASGITransport` object explicitly before being passed it to `AsyncClient`.

This behavior has been observed when running a CI/CD pipeline in the main Bluesky [repo](https://github.com/bluesky/bluesky/pull/1830#issuecomment-2512061914). After the fix, the previously failing tests have passed (in a local environment).

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/client/array.py
+++ b/tiled/client/array.py
@@ -176,7 +176,7 @@ class _DaskArrayClient(BaseClient):
 
     def write_block(self, array, block, slice=...):
         url_path = self.item["links"]["block"].format(*block)
-        query_params = {
+        params = {
             **parse_qs(urlparse(url_path).query),
             **params_from_slice(slice),
         }
@@ -185,7 +185,7 @@ class _DaskArrayClient(BaseClient):
                 url_path,
                 content=array.tobytes(),
                 headers={"Content-Type": "application/octet-stream"},
-                params=query_params,
+                params=params,
             )
         )
 
--- a/tiled/client/array.py
+++ b/tiled/client/array.py
@@ -1,5 +1,6 @@
 import itertools
 from typing import Union
+from urllib.parse import parse_qs, urlparse
 
 import dask
 import dask.array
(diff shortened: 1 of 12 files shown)


danielballan:
This was overdue for a cleanup. Thanks. Looks good once CI is happy.

genematx:
There are still a few bits that need to be cleaned up (CICD goes into an infinite loop now...) -- working on it. Perhaps this is a reintroduced [bug in httpx](https://github.com/encode/httpx/issues/652), but probably still good to systematize our use of query params.

tiled/client/array.py
```diff
@@ -92,11 +93,13 @@ def _get_block(self, block, dtype, shape, slice=None):
             expected_shape = ",".join(map(str, shape))
         else:
             expected_shape = "scalar"
+        url_path = self.item["links"]["block"]
         content = handle_error(
             self.context.http_client.get(
-                self.item["links"]["block"],
+                url_path,
```
danielballan, line 99: I confirmed that in current httpx any query params here are ignored in `params` is passed.

```py
In [6]: httpx.get("http://example.com?a=1&b=3", params={'a': 2, 'c': 3}).request.url
Out[6]: URL('http://example.com?a=2&c=3')

In [7]: httpx.get("http://example.com?a=1&b=3").request.url
Out[7]: URL('http://example.com?a=1&b=3')
```

danielballan (approved):

PR #827, opened 2024-12-12
Add helm

### LINKED ISSUE #595: Helm Chart
We have a tiled container image and a docker-compose file that deploys it alongside Postgres, with Grafana+Prometheus for monitoring. We had discussed making a Helm Chart, but we have not yet had an immediate user for it. (NSLS2 does not have a production k8s cluster, yet.)

@gordonwatts has a use case. Let’s make a Helm chart.

DiamondJoseph:
### Checklist
- [x] Add a Changelog entry

Closes #595


helm/tiled/values.yaml
```diff
@@ -0,0 +1,143 @@
...
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+# This is to setup the liveness and readiness probes more information can be found here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+livenessProbe:
```
danielballan, line 100: There is a `/healthz` endpoint designed to meet k8s' expectations as described in https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

helm/tiled/values.yaml
```diff
@@ -0,0 +1,143 @@
...
+
+tolerations: []
+
+affinity: {}
+
+
+# Additional envVars to mount to the pod
+extraEnvVars:
+- name: TILED_SINGLE_USER_API_KEY
+  value: "7d0adb8a42634d14ae213c6a5b422e84"
```
danielballan, line 137: I guess this is an example key. Is there a way to avoid hard-coding one, to force an actual secret to be provided?

danielballan (commented):
Awesome! Two quick comments.

helm/tiled/values.yaml
DiamondJoseph: Yep! I just wanted to get it working today.

DiamondJoseph: Made the path of least resistance making a Secret and added docs on creating a SealedSecret.

danielballan (approved):

danielballan:
Thanks for contributing this!

Issue #814, opened 2024-11-19
Refine AccessPolicy API


danielballan:
Tersely capturing a design discussion with @nmaytan (to elaborate later)...

The AccessPolicy API may be extended from 

```py
def filter(self, node, principal, scopes)
    ...

def allowed_scopes(self, node, principal):
    ...
```

to

```py
async def filter(self, node, principal, scopes, segments=None)
    ...

async def allowed_scopes(self, node, principal, segments=None):
    ...
```

Then the enforcement code in the server will identify the first node in the tree that has a not-`None` `access_policy`, and it will pass that node into `filter` along with the list of segments (i.e. URL path given as list of strings). Within `filter`, the access policy can do any additional I/O that it wants in order to decide the access allowed on the target.

nmaytan:
Closed by #829 - a follow-up body of work will further refine `SecureEntry`

PR #829, opened 2024-12-13
Use last policy

nmaytan:
### Checklist
- [x] Add a Changelog entry

In ref of #814 - though implementation was adjusted to preserve existing behavior (i.e. subtrees can have differing access policies, still). This fixes an issue where the intended access policy as defined for a tree in a Tiled config is not properly applied to resources that are children in that tree, since the children _themselves_ did not explicitly have an access policy attribute.

The behavior is now:
* Starting at the root and descending, filter the entry according to the access policy of each node (if the node has one)
* When retrieving the requested resource, use the last access policy encountered while descending the tree
* If the target resource is inside a catalog adapter, the access policy used during the request is whatever access policy the catalog adapter node has (even if `None`)

`filters` and `alllowed_scopes` now also takes a `path_parts` parameter which should be the list of segments from the current node until the target resource - in case the access policy needs this information to perform an access check (such as pulling information from a database).

### DIFF
--- a/tiled/_tests/test_access_control.py
+++ b/tiled/_tests/test_access_control.py
@@ -48,10 +48,10 @@ def context(tmpdir_module):
         "access_control": {
             "access_policy": "tiled.access_policies:SimpleAccessPolicy",
             "args": {
-                "access_lists": {"alice": ["a", "c", "d", "e"]},
+                "access_lists": {"alice": ["a", "c", "d", "e", "g"]},
                 "provider": "toy",
                 "admins": ["admin"],
-                "public": ["f"],
+                "public": ["f", "g"],
             },
         },
         "trees": [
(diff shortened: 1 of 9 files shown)


danielballan:
I think the last piece we need here is a new test that uses a mock access policy which fetches metadata from the node (the `entry_with_access_policy`) to make its determination.

danielballan:
...Just to state it plainly: that will validate this design and make it much easier to reproduce any future issues with this.

danielballan:
Next we want an example access policy that makes use of `path_parts`. In `SimpleAccessPolicy`, if the filter lets a user see a given direct child of the `entry_with_access_policy`, then the user has the same scopes (`self.scopes`) on every (nested) child therein. What if an access policy wants to confer `write:data` on some sub-children but not others?

For example, what if some node was structured like:

```
my_collection/my_dataset/parameters_table
my_collection/my_dataset/results_array
```

Perhaps, based on the metadata in those items, the AccessPolicy would confer some data processing pipeline `write:data` on `results_array` but only `read:data` on `parameters_data`?

danielballan:
I'm not sure that's the best use case I can come up with. Maybe we should think a little more about this and get crisper use cases. But it seems important for `allowed_ scopes` to know which node we are actually trying to access, not just broadly `entry_with_access_policy`.

tiled/_tests/test_access_control.py
```diff
@@ -2,18 +2,66 @@
...
 from ..server.app import build_app_from_config
+from ..server.core import NoEntry
 from .utils import enter_username_password, fail_with_status_code
 
 arr = numpy.ones((5, 5))
+arr_zeros = numpy.zeros((5, 5))
 arr_ad = ArrayAdapter.from_array(arr)
 
 
+class EntryBasedAccessPolicy(SimpleAccessPolicy):
```
danielballan, line 27: A docsting here would be useful:

> This is access policy demonstrates how the metadata on some nested child node can be efficiently consulted and incorporated in logic that determines access scopes. In this test example, the metadata on the node quite literally lists some scopes that it should not allow. In realistic examples it could be incorporated in site-specific logic.

danielballan:
CI failures are apparently due to a GHA+npm networking problem that's been going on for the last hour or so. Not related to the content of the PR.

PR #835, opened 2024-12-19
Fix critical bug in delete_tree

danielballan:
The `delete_tree` deletes nodes recursively. It was added specifically to support `tiled serve directory --watch ...`, which blows and rebuilds the database when files change (as an initial simple/safe implementation).

Crucially, it can only ever delete _registered_ data, never user-uploaded data. Which is a good thing because, as implemented, it always deletes the entire tree---not from a prescribed node downward but always from the root. Bad bug!

This bug was masking other bugs, also fixed here. Needs tests, but here's a demo of the intended behavior working as planned on this PR branch:

```py
In [1]: from tiled.client import from_uri, show_logs, tree

In [2]: c = from_uri('http://localhost:8000', api_key='secret')

In [3]: c.create_container(key='A')
   ...: c.create_container(key='B')
Out[3]: <Container {}>

In [4]: c['A'].create_container('a1')
Out[4]: <Container {}>

In [5]: tree(c)
├── A
│   └── a1
└── B

In [6]: c['A'].delete_tree()

In [7]: tree(c)
├── A
└── B

In [8]: c.delete('A')

In [9]: tree(c)
└── B
```

### Checklist
- [ ] Add a Changelog entry


tiled/catalog/adapter.py
```diff
@@ -921,26 +926,27 @@ async def delete_tree(self, external_only=True):
...
-                        orm.DataSource.management != Management.external
-                    )
-                )
-                for condition in conditions:
-                    sel_int_asset_statement.filter(condition)
-                int_assets = (await db.execute(sel_int_asset_statement)).all()
-                for data_uri, is_directory in int_assets:
-                    delete_asset(data_uri, is_directory)
+                raise NotImplementedError
             # TODO Deal with Assets belonging to multiple DataSources.
```
danielballan, line 930: I spent some time working through this, and it's quite tricky. Since this has no use cases for us yet, I propose to defer this (and open an issue) without letting it holdup the fix for the common case.

PR #836, opened 2025-01-03
Clean up Python client auth

### LINKED ISSUE #820: Re-prompt if password is rejected
Currently if you misstype you password in Tiled's `getpass()` prompt, Tiled raises an exception. Instead it should prompt again (a reasonable number of times). An empty password (i.e. just pressing Enter) should escape the retry loop.

### LINKED ISSUE #831: Python API fails connect with private Tiled servers
Conditions:

- Client / Server: 0.1.0b12
- Auth: ORCiD
- AccessPolicy: DummyAccessPolicy
- allow_anonymous_access: false

Python API fails connect with private Tiled servers (specifically with `allow_anonymous_access: false`). The entire authentication process through ORCID completes correctly but, the `tiled.client.from_uri` fails to return and raises an error. Notably, changing `allow_anonymous_access: true` allows things to work.

Example:

```
c = from_uri("https://127.0.0.1:8443") 
```

Resulting Error (subset of a long stack trace):

> ClientError: 401: Not enough permissions. Requires scopes ['read:metadata']. Request had scopes [] https://127.0.0.1:8443/api/v1/metadata/?include_data_sources=false

### LINKED ISSUE #833: Tiled CLI "forgets" that is logged in for Public/Private Tiled Servers
Tiled CLI does not seem to remember that it successfully logged in after `tiled login`. Future calls to CLI do not use the acquired tokens. Maybe related to #831 ? This is fixed by reverting client to v0.1.0b11 or by using a fully private server (ie `allow_anonymous_access: false`)

Conditions:

- Client / Server: 0.1.0b12
- Auth: ORCiD
- AccessPolicy: DummyAccessPolicy
- allow_anonymous_access: true

Example: 

```
% tiled login                     

You have 15 minutes visit this URL

  https://orcid.org/oauth/authorize?client_id=APP-AXDIPEVR6IMA0RL7&response_type=code&scope=openid&redirect_uri=https%3A%2F%2F127.0.0.1%3A8443%2Fapi%2Fv1%2Fauth%2Fprovider%2Forcid%2Fdevice_code

and enter the code:

  017E-D806


Waiting........
You have logged in with ORCID as 0000-0003-4736-0743.

% tiled tree                      
├── A
├── B
├── C
└── D
    ├── temperature
    └── time
```

But this `tiled tree` request does not use the login information, and the uvicorn log shows that is requesting via public access:
```
[3141081c5a25237c] 192.168.65.1:48548 (public) - "GET /api/v1/ HTTP/1.1" 200 OK
```

danielballan:
The goal of this PR is to improve the user experience and simplify the implementation for authentication in the Python client. There are no changes to the server in this PR.The new docs page in this PR, `docs/source/how-to/authentication.md`, is a good place to start. I recommend reading that before reading the summary below.

## Regressions fixed

The PR addresses a critical regressions, both released in v0.1.0b12

A [trailing-comma-as-tuple bug](https://github.com/bluesky/tiled/pull/823/files#diff-d951a6f56304c48bd302a77ab64c9e290913abf861ad83af86f21c427a7cd751R157) in #823 which manifests as

```
>>> from_uri("https://tiled.nsls2.bnl.gov")
...
KeyError: 0
```

and a deeper bug in #816 that prevents login:

```py
>>> from_uri("https://tiled.nsls2.bnl.gov")
ClientError: 401: Incorrect username or password https://tiled.nsls2.bnl.gov/api/v1/auth/provider/pam/token
```

## Design changes 

- The high-level convenience wrappers no longer collect an optional username. The `username` kwarg in `from_uri(..., username="...")` and `from_profile(..., username="...")` is now completely ignored (with a warning to the user). It will later be removed.
    - The `username` kwarg never made sense for browser-based (device code grant) auth anyway. (In that case, the username goes to the browser, not Tiled.) It always was a bit of wart.
    - The `username` kwarg has been the source of a lot of confusion over the years... with unintuitive behavior related to `username=None` and lately `username=UNSET`. I take this as a sign that it should not have been merged into the `from_uri` and `from_profile` convenience functions.
    - A username prompt like `Username [dallan]:` can be a "password trap". Tiled is trying to collect your username, but you might reflexively type your password onto the screen in clear text. If we don't have a valid session ready to reuse, we should just ask the user to type their username with plain old `Username:`.
- This PR enables users to opt out of stashing tokens with a new keyword argument: `from_uri("https://...", remember_me=False)`. The tokens will be stored only in memory. This is equivalent of logging into GMail in an Incognito Browser.
- For a given token directory---typically `$XDG_CACHE_DIR/tiled/tokens/`---Tiled only stashes tokens for one user at a time (per Tiled server, as defined by its `netloc`).
    - It seems like a mistake to have ever supported stashing more than one. Storing two distinct principals' tokens with the same file permissions invites leakage.
    - The use case I originally had in mind was two processes running in a shared account (e.g. an NSLS2 Operator Account) authenticated as differnet users. But this rather niche use case can be addressed in many other, simpler ways:
        - `remember_me=False` "incognito mode"
        - setting custom token cache locations using `TILED_CACHE_DIR`
        - using an API key
    - This removes the notion of "default identity" and simplifies the state on disk. Instead of deeply nested directories like `~/.cache/tiled/{netloc}/{idP}/{username}` we now just have one directory per tiled server: `~/.cache/tiled/{netloc}`.
    - There is no need for `logout(clear_default=True)`. Now `logout()` is all you need.

Importantly, it is still straightforward to authenticate to multiple servers in one session because the tokens are still scoped by the server's netloc:

```py
aps_client  = from_uri("https://tiled.aps.anl.gov")  # an aspirational URL
nsls2_client = from_uri("https://tield.nsls2.bnl.gov")
```

Finally, as mentioned in the docs, this provides a path for applications without access to an interactive terminal. The approach outlined there (which could be fleshed out in the future with a fuller example) will work for both password grant and device code grant modes, as well as for Tiled servers that offer multiple authentication providers. Authentication code has been factored out in to public functions `password_grant` and `device_code_grant`, suitable for reuse.

Closes #820 
Closes #831 
Closes #833 

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/docs/source/how-to/authentication.md
+++ b/docs/source/how-to/authentication.md
@@ -1,0 +1,149 @@
+# Python Client Authentication
+
+This covers authentication from the user (client) perspective. To learn how to
+_deploy_ authenticated Tiled servers, see {doc}`../explanations/security`.
+
+## Interactive Login
+
+Some Tiled servers are configured to let users connect anonymously without
+authenticating.
+
+```py
+>>> from tiled.client import from_uri
+>>> client = from_uri("https://...")
+>>> <Container ...>
+```
+
+Logging in may enable you to see more datasets that may not be public.
+Log in works in one of two ways, depending on the server.
+
+1. Username and password ("OAuth2 password grant")
+
+   ```py
+   >>> client.login()
+   Username: ...
+   Password:
+   ```
+
+2. Via a web browser ("OAuth2 device code grant")
+
+   ```py
+   >>> client.login()
+   You have 15 minutes visit this URL
+
+   https://...
+
+   and enter the code: XXXX-XXXX
+   ```
+
+In the future, Tiled will log you into this server automatically, without
+re-prompting for credentials, until your session expires.
+
+   ```py
+   >>> from tiled.client import from_uri
+   >>> client = from_uri("https://...")
+   # Automatically logged in!
+
+   # This is a quick way to verify whether you are already logged in
+   >>> client.context
+   <Context authenticated as '...'>
+   ```
+
+To opt out of this, set `remember_me=False`:
+
+```py
+>>> from tiled.client import from_uri
+>>> client = from_uri("https://...", remember_me=False)
+```
+
+```{note}
+Tiled stores OAuth2 tokens (it _never_ stores your password) in files
+with properly restricted permissions under `$XDG_CACHE_DIR/tiled/tokens`,
+typically `~/.config/tiled/tokens` on Linux and MacOS.
+
+To customize the location of this storage, set the environment variable
+`TILED_CACHE_DIR`.
+```
+
+Some Tiled servers are configured to always require login, disallowing any
+anonymous access. For those, the client will prompt immediately, such as:
+
+   >>> from tiled.client import from_uri
+   >>> client = from_uri("https://...")
+   Username:
+   ```
+
+## Noninteractive Authentication (API keys)
+
+There are environments where logging in interactively is not possible,
+such as running a batch script. For these applications, we recommend
+using an API key. These can be created from the CLI:
+
+```sh
+$ tiled login
+$ tiled api_key create
+```
+
+or from an interactive Python session:
+
+```py
+>>> client = from_uri("https://...")
+>>> client.login()
+>>> client.create_api_key()  # {"secret": ...}
+```
+
+The best way to provide an API key is to set the environment variable
+`TILED_API_KEY`. A script like this:
+
+```py
+from tiled.client import from_uri
+
+client = from_uri("https://....")
+```
+
+will detect that `TILED_API_KEY` is set and use that API key for
+authentication with Tiled. This is equivalent to:
+
+```py
+import os
+from tiled.client import from_uri
+
+client = from_uri("https://....", api_key=os.environ["TILED_API_KEY"])
+```
+
+Avoid typing the API key in to the code:
+
+```py
+from_uri("https://...", api_key="secret!")  # DON'T
+```
+
+as it is easy to accidentally share or leak.
+
+## Custom Applications
+
+Custom applications, such as a graphical interfaces that wrap Tiled,
+may not be able to use Tiled commandline-based prompts. They may
+provide their own mechanism for collection credentials
+(for password grants) or launching a browser and waiting for the
+user to authorize a session (for device code grants).
+
+Custom applications should avoid using the convenience functions
+`tiled.client.construtors.from_uri` and
+`tiled.client.construtors.from_profile`. Instead, follow this pattern:
+
+```py
+from tiled.client import Context
+
+def my_prompt(http_client, providers):
+    ...
+    return tokens
+
+
+URI = "https://..."
+context, node_path_parts = Context.from_any_uri(URI)
+context.authenticate(prompt_for_reauthentication=my_prompt)
+client = from_context(context, node_path_parts=node_path_parts)
+```
+
+Note that `my_prompt` will be called immediately _and may also be called again
+later_ if the active sessions expires or is revoked.
--- a/docs/source/how-to/authentication.md
+++ b/docs/source/how-to/authentication.md
@@ -127,29 +127,32 @@ as it is easy to accidentally share or leak.
 
 ## Custom Applications
 
-Custom applications, such as a graphical interfaces that wrap Tiled,
-may not be able to use Tiled commandline-based prompts. They may
-provide their own mechanism for collection credentials
-(for password grants) or launching a browser and waiting for the
-user to authorize a session (for device code grants).
+Custom applications, such as a graphical interfaces that wrap Tiled, may not be
+able to use Tiled commandline-based prompts. They should avoid using the
+convenience functions `tiled.client.construtors.from_uri` and
+`tiled.client.construtors.from_profile`.
 
-Custom applications should avoid using the convenience functions
-`tiled.client.construtors.from_uri` and
-`tiled.client.construtors.from_profile`. Instead, follow this pattern:
+They may implement their own interfaces for collecting credentials (for
+password grants) or launching a browser and waiting for the user to authorize a
+session (for device code grants). The functions
+`tiled.client.context.password_grant` and
+`tiled.client.context.device_code_grant` may be useful building blocks. The
+tokens obtained from this process may then be passed directly in to the Tiled
+client like so.
+
 
 ```py
 from tiled.client import Context
 
-def my_prompt(http_client, providers):
-    ...
-    return tokens
-
-
 URI = "https://..."
 context, node_path_parts = Context.from_any_uri(URI)
-context.authenticate(prompt_for_reauthentication=my_prompt)
+tokens, remember_me = launch_custom_interface()
+context.configure_auth(tokens, remember_me=remember_me)
 client = from_context(context, node_path_parts=node_path_parts)
 ```
 
-Note that `my_prompt` will be called immediately _and may also be called again
-later_ if the active sessions expires or is revoked.
+The client will transparently handle OAuth2 refresh flow. If the session is
+revoked or expires, and an attempt at refreshing the tokens is thus rejected
+by the server, the exception `tiled.client.auth.CannotRefreshAuthentication`
+will be raised. The application should be prepared to catch that exception
+and reinitiate authentication.
--- a/docs/source/how-to/authentication.md
+++ b/docs/source/how-to/authentication.md
@@ -81,7 +81,7 @@ using an API key. These can be created from the CLI:
 
 ```sh
 $ tiled login
-$ tiled api_key create
+$ tiled api_key create --expires-in 7d --note "for this week's experiment"
 ```
 
 or from an interactive Python session:
@@ -89,7 +89,13 @@ or from an interactive Python session:
 ```py
 >>> client = from_uri("https://...")
 >>> client.login()
->>> client.create_api_key()  # {"secret": ...}
+>>> client.create_api_key(expires_in="7d", note="for this week's experiment")
+{"secret": ...}
+ ```
+
+The expiration and note are optional, but recommended. Expiration can be given
+in units of years `y`, days `d`, hours `h`, minutes `m`, or seconds `s`.
+
 ```
 
 The best way to provide an API key is to set the environment variable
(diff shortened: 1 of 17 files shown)


docs/source/how-to/authentication.md
```diff
@@ -0,0 +1,163 @@
...
+>>> client.context
+<Context authenticated as '...'>
+```
+
+To opt out of this, set `remember_me=False`:
+
+```py
+>>> from tiled.client import from_uri
+>>> client = from_uri("https://...", remember_me=False)
+```
```
padraic-shafer, line 75: If valid tokens already exist, they will be used.  `remember_me` gets applied to subsequent connections.  I wonder whether existing tokens should get expired/ignored in the current connection being requested when `remember_me=False` is passed?

```suggestion
To opt out of this for subsequent connections, set `remember_me=False`:

```py
>>> from tiled.client import from_uri
>>> client = from_uri("https://...", remember_me=False)
```
```

File: docs/source/how-to/authentication.md [hunk 2]
```diff
@@ -0,0 +1,163 @@
...
+or from an interactive Python session:
+
+```py
+>>> client = from_uri("https://...")
+>>> client.login()
+>>> client.create_api_key()  # {"secret": ...}
+```
+
+The best way to provide an API key is to set the environment variable
+`TILED_API_KEY`. A script like this:
```
padraic-shafer, line 110: Would this be a good place to also recommend setting an appropriately "short" lifetime for the token, when it used on shared workstations?

docs/source/how-to/authentication.md
```diff
@@ -0,0 +1,163 @@
...
+
+Custom applications, such as a graphical interfaces that wrap Tiled,
+may not be able to use Tiled commandline-based prompts. They may
+provide their own mechanism for collection credentials
+(for password grants) or launching a browser and waiting for the
+user to authorize a session (for device code grants).
+
+Custom applications should avoid using the convenience functions
+`tiled.client.construtors.from_uri` and
+`tiled.client.construtors.from_profile`. Instead, follow this pattern:
```
padraic-shafer, line 146:
```suggestion
`tiled.client.constructors.from_uri` and
`tiled.client.constructors.from_profile`. Instead, follow this pattern:
```

tiled/client/context.py
```diff
@@ -934,40 +854,60 @@ def _can_prompt():
...
+
+
+def device_code_grant(http_client, auth_endpoint):
+    verification_response = http_client.post(auth_endpoint, json={}, auth=None)
+    handle_error(verification_response)
+    verification = verification_response.json()
+    authorization_uri = verification["authorization_uri"]
+    print(
+        f"""
+You have {int(verification['expires_in']) // 60} minutes visit this URL
```
padraic-shafer, line 875:
```suggestion
You have {int(verification['expires_in']) // 60} minutes to visit this URL
```

tiled/client/context.py
```diff
@@ -934,40 +854,60 @@ def _can_prompt():
...
+            },
+            auth=None,
+        )
+        if (access_response.status_code == httpx.codes.BAD_REQUEST) and (
+            access_response.json()["detail"]["error"] == "authorization_pending"
+        ):
+            print(".", end="", flush=True)
+            continue
+        handle_error(access_response)
+        print("")
```
padraic-shafer, line 910: Should it display a "success" message here, or does the equivalent happen in an outer function?

```suggestion
        print("\nSUCCESS: Authenticated")
```

tiled/client/context.py
```diff
@@ -22,25 +21,78 @@
...
+    if len(providers) == 1:
+        # There is only one choice, so no need to prompt the user.
+        (spec,) = providers
+    else:
+        while True:
+            print("Authenticaiton providers:")
+            for i, spec in enumerate(providers, start=1):
+                print(f"{i} - {spec['provider']}")
+            raw_choice = input(
+                "Choose an authentication provider (or press Enter to escape): "
```
padraic-shafer, line 36:
```suggestion
                "Choose an authentication provider (or press Enter to cancel): "
```

tiled/client/context.py
```diff
@@ -22,25 +21,78 @@
...
-        username_reprompt = input(f"Username [{username}]: ")
-        if len(username_reprompt.strip()) != 0:
-            username = username_reprompt
-    else:
+    spec = _choose_identity_provider(providers)
+    auth_endpoint = spec["links"]["auth_endpoint"]
+    provider = spec["provider"]
+    mode = spec["mode"]
+    if mode == "password":
+        # Prompt for username, password at terminal.
```
padraic-shafer, line 64: Should external clients also prompt for the username every time, rather than store the name of the last user as a convenience?

In other words, since Tiled python client has made this change, is it a recommendation for all Tiled clients?

tiled/client/context.py
```diff
@@ -22,25 +21,78 @@
...
+        for _attempt in range(PASSWORD_ATTEMPTS):
+            password = getpass.getpass()
+            try:
+                tokens = password_grant(
+                    http_client, auth_endpoint, provider, username, password
+                )
+            except httpx.HTTPStatusError as err:
+                if err.response.status_code == httpx.codes.UNAUTHORIZED:
+                    print(
+                        "Username or password not recognized. Retry, or press Enter to escape."
```
padraic-shafer, line 76:
```suggestion
                        "Username or password not recognized. Retry, or press Enter to cancel."
```

padraic-shafer (commented):
Overall looks good.  I've added some questions/suggestions in the comments.

docs/source/how-to/authentication.md
danielballan: Yes, done. While I'm at it, I added support for expressions with units like `"1d"`.

tiled/client/context.py
danielballan: The equivalent happens in the caller (`prompt_for_credentials`) where we print a message that indicates not only "success" but specifically the username that you have just successfully logged in with. The exact message is customizable by the server, so we can comply with rules like the ORCID Terms of Service (which requires a confirmation that includes the ORCID digits) or BNL Cyber rules (which require the boilerplate about accessing a Federal system).

tiled/client/context.py
danielballan: What do you have in mind when you say "external clients"?

tiled/client/auth.py
```diff
@@ -69,17 +72,19 @@ def sync_set_token(self, key, value):
...
 
     def sync_clear_token(self, key):
         with self._sync_lock:
-            self.tokens.pop(key, None)
-            filepath = self.token_directory / key
-            filepath.unlink(missing_ok=False)
+            if self.token_directory is not None:
+                self.tokens.pop(key, None)
+                filepath = self.token_directory / key
+                filepath.unlink(missing_ok=False)
```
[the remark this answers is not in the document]
danielballan: The `key` must be a string; it's either `"access_token"` or `"refresh_token"`, specifying which token the caller intends to clear. The corresponding file is deleted here.

If key were `None` / empty, that would be a bug in the call and would likely raise on Line 86 here.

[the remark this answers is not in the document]
padraic-shafer: It looks like it would remove the directory if it were called with `key=""`.  But I take your point, that would be a bug in the call; the call is used infrequently in the code base; the function is not really intended for general users.

**End result: It's a non-issue.**

```python3
In [1]: from pathlib import Path

In [2]: base = Path("mydir")

In [3]: base / "file"
Out[3]: PosixPath('mydir/file')

In [4]: base / ""
Out[4]: PosixPath('mydir')

In [5]: base /
  Cell In[5], line 1
    base /
           ^
SyntaxError: invalid syntax


In [6]: base / None
TypeError                                 Traceback (most recent call last)
Cell In[6], line 1
----> 1 base / None

TypeError: unsupported operand type(s) for /: 'PosixPath' and 'NoneType'
```

tiled/client/context.py
padraic-shafer: I was thinking of general GUI clients (Qt, JS, etc.) and also had not seen the nice summary in "Design changes" when I wrote that question...so I was guessing at the reasons for removing `username` from the function args.

I see now that the reasons were partly about confusion in the function call and partly about UX (whoops, I just typed my plaintext password into the username field).

padraic-shafer: I suppose I was asking: _Is there a security concern for caching the username as a convenience for the user?_

It looks the answer is that it may be a behavioral concern (did the user realize that it's not the right place to enter their password) but not generally a "leakage" concern of _Who was the last user before me?_

My takeaway is that an app _could_ store the username but that would be up to the app designer, and if so then it should be clear where the password gets entered.

danielballan: Right. Fully on the same page, just reiterating:

The change in behavior in Tiled here is: we used to cache the username ("default identity") and now we only cache tokens. So, you either have a valid session of you don't, and if you don't Tiled itself will not retain any state related to some previously-used identity or session. That would be up to applications to manage as they wish, as you said.

danielballan:
By the way, regarding the CI failures here, I haven't even bothered running `pytest` locally yet. I'm sure there is a lot to fix related to API changes. Just looking for feedback on the user interface and the approach. Then I'll tighten it up.

docs/source/how-to/authentication.md
```diff
@@ -0,0 +1,168 @@
...
+Custom applications, such as a graphical interfaces that wrap Tiled,
+may not be able to use Tiled commandline-based prompts. They may
+provide their own mechanism for collection credentials
+(for password grants) or launching a browser and waiting for the
+user to authorize a session (for device code grants).
+
+Custom applications should avoid using the convenience functions
+`tiled.client.constructors.from_uri` and
+`tiled.client.constructors.from_profile`. Instead, follow this pattern:
+
```
maffettone, line 155: This might be a good place to link: https://github.com/bluesky/tiled/blob/799f7ba163a9424a44a366e232fc995854d12470/tiled/client/context.py#L55-L95

as a sensible default.

maffettone (commented):
Conceptually this makes a lot of sense and clears some clutter that could become problematic. 
- I really like the `parse_time_string` changes. That could (for expediency and clarity) be its own atomic PR, I think. 
- The caching changes make a lot of sense for locking this down. The use case of "two processes running in a shared account" should be accomplished by on-behalf-of flow. 
- This will probably break some of the testing and applications designed around here (https://github.com/bluesky/tiled/pull/816/files), but rightfully so. And that is a recent enough change that it won't be a strong dependency of many. When we are at a confident state, I can update https://github.com/NSLS2/tiled-qt-tools to serve as an on site example of the interaction.

jwlodek:
Looks like the correct approach to me as well. I think caching username is a nice convenience, but probably not really needed, and can be done at the application level.

PR #837, opened 2025-01-07
Support expiration times given as a number with units.

danielballan:
Split out from #836 

### Checklist
- [X] Add a Changelog entry
- [ ] ~Add the ticket number which this PR closes to the comment section~


maffettone (approved):

maffettone:
`Union` seems to be breaking the testing....

PR #851, opened 2025-01-20
Add documentation and configuration for deploying Helm chart with oauth2-proxy

DiamondJoseph:
Explains how to add [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy) as a dependency of an umbrella Helm chart, allowing the web front end and API to be served behind a reverse proxy that redirects unauthenticated traffic to the auth provider.

According to Jake (Filik), passing the `Authorization` header into the application requires using the unstable alphaConfig. I'd like to find a nicer way of handling the oauth2-proxy -> tiled service name mapping.

Parking here momentarily until authentication can be revisited. 

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/docs/source/how-to/helm.md
+++ b/docs/source/how-to/helm.md
@@ -104,8 +104,24 @@ Deploying behind a reverse proxy that redirects unauthenticated requests to your
 Configure your OAuth2 client values, referencing back to the Quickstart guide for how to configure a SealedSecret in your templates directory.
 
 ```{note}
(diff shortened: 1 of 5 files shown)


danielballan:
Power-cycled to re-run CI against `main`.

DiamondJoseph:
I'd hold fire on this one for a day or two, I'm currently in progress deploying it at DLS and I've got some modifications to make

PR #856, opened 2025-01-23
Refactor Authenticators to make use of inheritance

DiamondJoseph:
Replaces the use of the Mode enum with inheritance from two abstract base classes that provide different authentication mechanisms. Added some type hints.

Replaced "password" in AboutAuthenticationProvider with "internal", which is then a matching pair with "external". 

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/authenticators.py
+++ b/tiled/authenticators.py
@@ -5,7 +5,7 @@ import logging
 import re
 import secrets
 from collections.abc import Iterable
-from typing import Any, cast
+from typing import Any, Mapping, Optional, cast
 
 import httpx
 from fastapi import APIRouter, Request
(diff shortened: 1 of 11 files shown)


tiled/authenticators.py
```diff
@@ -98,20 +93,18 @@ def __init__(self, service="login", confirmation_message=""):
         self.confirmation_message = confirmation_message
         # TODO Try to open a PAM session.
 
-    async def authenticate(self, username: str, password: str) -> UserSessionState:
+    async def authenticate(self, username: str, password: str) -> UserSessionState | None:
         import pamela
 
         try:
             pamela.authenticate(username, password, service=self.service)
+            return UserSessionState(username, {})
```
DiamondJoseph, line 101: IMO this is clearer than try/except/else: if pamela.authenticate doesn't throw an exception, we go from line 100 to line 101, rather than to line 110, and the happy path all lives together

tiled/client/context.py
```diff
@@ -6,11 +6,14 @@
 import urllib.parse
 import warnings
 from pathlib import Path
+from typing import List
 from urllib.parse import parse_qs, urlparse
 
 import httpx
 import platformdirs
 
+from tiled.server.schemas import AboutAuthenticationProvider
```
danielballan, line 15: This pulls in server-side deps. This keeps coming up; maybe we should move these schemas into `tiled.schemas`.

danielballan (commented):
I like this. One depedency-management issue noted in line.

DiamondJoseph:
I may rewrite history a little if this passes and move the schema stuff to a seperate PR.

DiamondJoseph:
This will be rebased after https://github.com/bluesky/tiled/pull/862 goes in

danielballan:
Ready for rebase. Please add a `CHANGELOG` entry also.

PR #862, opened 2025-01-24
Move Schemas for describing server configuration

DiamondJoseph:
Moves Schemas that define current server behaviour into common module, so that use can be made of it in the tiled client.

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 # Changelog
 
+## Unreleased
+
+### Maintenance
+
+- Make depedencies shared by client and server into core dependencies.
+- Use schemas for describing server configuration on the client side too.
+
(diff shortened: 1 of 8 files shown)


danielballan (commented):
We need to add `pydantic` to the `client` and `minimal-client` dependency groups in `pyproject.toml`.

DiamondJoseph:
> We need to add `pydantic` to the `client` and `minimal-client` dependency groups in `pyproject.toml`.

I would prefer for #863 to go in prior to this- looking at the changed files for that commit, it's so much easier to see what is going on maintaining 1 list of common dependencies, rather than replicating the same dependency (and needing to keep to equivalent versions) in 5-6 places, that are often long enough to not have all context shown in the files changes tab. Yes to keeping dependencies specific to one or the other separate, but there are common dependencies for e.g. serialisation and communication that should be kept in step.

danielballan:
Rebased on `main` and updated CHANGELOG.

PR #865, opened 2025-01-24
Allow reading string-valued arrays from DataFrames

genematx:
1. Added `.get` method to `TableAdapter` and `ParquetDatasetAdapter`
2. Allow representation of string-valued arrays derived from columns of data frames. The resulting arrays are cast as `<U{max_size}` type, where `max_size` is the maximum length among all strings in the array.

### Checklist
- [x] Add a Changelog entry
- [x] ~~Add the ticket number which this PR closes to the comment section~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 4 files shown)


danielballan (approved):
I rebased on `main` and took this for a test drive. Looks good.

PR #867, opened 2025-01-27
WIP: Array Views

genematx:
This adds a mechanism for creating views of arrays in Tiled. The source for a view could be either an array or a column of a table already registered in Tiled. the view is represented as its own DataSource with possibly several Assets pointing to a location of the source in the Tiled catalog tree, e.g. `tiled://X/Y/image` or `tiled://X/Y/table/column_1`. There is a possibility to use additional parameters to slice or reshape (TBD?) the original array and present only a subset of its values as a view. Likewise, multiple arrays (declared in separate assets) can be concatenated into a single view (TBD?).

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/adapters/array.py
+++ b/tiled/adapters/array.py
@@ -148,16 +148,6 @@ class ArrayAdapter:
         return array
 
 
-class ArrayViewAdapter(ArrayAdapter):
-
-    structure_family = StructureFamily.array
-
-    @classmethod
-    def from_catalog(cls, data_source, node, **kwargs):
-        breakpoint()
-        pass
-
-
 def slice_and_shape_from_block_and_chunks(
     block: Tuple[int, ...], chunks: Tuple[Tuple[int, ...], ...]
 ) -> Tuple[NDSlice, Tuple[int, ...]]:
--- a/tiled/adapters/array.py
+++ b/tiled/adapters/array.py
@@ -103,7 +103,7 @@ class ArrayAdapter:
         slice = data_source.parameters.get("slice")
         slice = NDSlice.from_json(slice) if slice is not None else None
         if isinstance(adapter, ArrayAdapter):
-            arr = adapter._array[slice] if slice else adapter._array
+            arr = adapter._array[tuple(slice)] if slice else adapter._array
         else:
             arr = adapter.read(slice) if slice else adapter.read()
 
@@ -129,7 +129,7 @@ class ArrayAdapter:
 
     def read(
         self,
-        slice: Union[NDSlice, EllipsisType] = ...,
+        slice: Optional[NDSlice] = None,
     ) -> NDArray[Any]:
         """
 
@@ -141,7 +141,7 @@ class ArrayAdapter:
         -------
 
         """
-        array = self._array[slice]
+        array = self._array[tuple(slice)] if slice else self._array
         if isinstance(self._array, dask.array.Array):
             return array.compute()
         return array
@@ -149,7 +149,7 @@ class ArrayAdapter:
     def read_block(
         self,
         block: Tuple[int, ...],
-        slice: Union[NDSlice, EllipsisType] = ...,
+        slice: Optional[NDSlice] = None,
     ) -> NDArray[Any]:
         """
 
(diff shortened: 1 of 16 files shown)


tiled/client/container.py
```diff
@@ -792,6 +796,56 @@ def write_block(x, block_id, client):
             # to bother with the return type?
             da.map_blocks(write_block, dtype=da.dtype, client=client).compute()
         return client
+    
+    def create_array_view(self, links: Iterable[str], key: str=None, slices: Optional[Iterable[NDSlice]] = None):
+
+        from ..structures.array import ArrayStructure
+
+        def serialize_ndslice(ndslice):
```
genematx, line 804: Same as above; to be moved to utils or an NDSlice method.

tiled/client/container.py
```diff
@@ -792,6 +796,56 @@ def write_block(x, block_id, client):
...
+            input_structure = document['data']['attributes']['structure']
+            out_shape = ndindex(slc).newshape(input_structure['shape'])
+            out_chunks = [[i] for i in out_shape]    # No chunking, TODO: more general case -- reassign chunks
+
+            substructures.append(copy.deepcopy(input_structure))
+            substructures[-1]['shape'] = out_shape
+            substructures[-1]['chunks'] = out_chunks
+            assets.append(Asset(data_uri=f"tiled://{links[0]}", is_directory=False, parameter='data_uri'))
+    
+        # TODO: combine substructures if tehre are multiple sub-arrays that are concatenated
```
genematx, line 832: Here, we can concatenate several subarrays into a single view. Not sure if it's needed or useful, but seems like a sensible generalization.

tiled/catalog/adapter.py
```diff
@@ -453,13 +455,48 @@ async def get_adapter(self):
...
+                    node = (await db.execute(statement)).scalar()
+                if node is None:
+                    raise RuntimeError("Can not find the root node for %s", asset.data_uri)
+                adapter = await STRUCTURES[node.structure_family](self.context, node).lookup_adapter(rest)
+                if isinstance(adapter, CatalogNodeAdapter):
+                    adapter = await adapter.get_adapter()
+
+                ndslice = deserialize_ndslice(data_source.parameters['slices'][0])
+
+                adapter = ArrayAdapter.from_array(adapter.read(slice=ndslice))
```
genematx, line 490: Possibly use `ArrayViewAdapter`, although it's probably unnecessary...

tiled/adapters/array.py
```diff
@@ -52,8 +54,8 @@ def from_array(
         cls,
         array: NDArray[Any],
         *,
-        shape: Optional[Tuple[int, ...]] = None,
-        chunks: Optional[Tuple[Tuple[int, ...], ...]] = None,
+        # shape: Optional[Tuple[int, ...]] = None,
```
genematx, line 57: we never use these...🤔

tiled/adapters/array.py
```diff
@@ -139,7 +163,7 @@ def read_block(
         """
         # Slice the whole array to get this block.
         slice_, _ = slice_and_shape_from_block_and_chunks(block, self._structure.chunks)
-        array = self._array[slice_]
+        array = self._array[tuple(slice_)]
```
genematx, line 166: `_array[...]` requires a tuple here; even though `NDSlice` is a subclass of tuple, it doesn't work here. Probably there's a better way to define `NDSlice` so it can be accepted right away (if we want to go this route at all, of course)

genematx:
@whs92

dylanmcreynolds:
This is really cool! 

I have a question. At what point do you see the view configured? After every scan? We might want to think about automating that at some point in bluesky.

genematx:
@dylanmcreynolds 
Yes, the original idea was to create the views in the Bluesky container after every scan, most likely somewhere in [TiledWriter](https://github.com/bluesky/bluesky/blob/main/src/bluesky/callbacks/tiled_writer.py), to define a flat namespace across table columns. We are now looking at a bit more performant way to do this, though, but views may still be useful in other scenarios.

PR #870, opened 2025-01-29
Add configuration and documentation for debugging remote instance in the cluster

DiamondJoseph:
I'm needing to debug the contents of the authorization headers that are being passed into my instance from the oauth2-proxy that it is deployed behind in the cluster: which requires that it is deployed with specific callback addresses that cannot easily be mocked from my machine. 

### Checklist
- [ ] Add a Changelog entry


helm/tiled/templates/deployment.yaml
```diff
@@ -32,6 +32,28 @@ spec:
...
+          - "-m"
+          args:
+          - "tiled"
+          - "serve"
+          - "config"
+          - "--host"
+          - "0.0.0.0"
+          - "--port"
+          - "8000"
+          - "--scalable"
```
DiamondJoseph, line 56: These duplicate the Dockerfile `CMD` arguments: Kubernetes uses `command` ~analogously to Docker `ENTRYPOINT` and `args` ~analogously to Docker `CMD`. Yes, it's frustrating. 

Unfortunately if you set `command`, `CMD` args are ignored, if they were picked up we wouldn't need to duplicate this here.

dylanmcreynolds (approved):
I learned some new stuff from the PR. I don't absolutely love distributing debugpy with all instance of the tiled image, but it's better than trying to maintain dual images.

PR #871, opened 2025-01-29
Ensure back-compat with old Python clients

danielballan:
### Checklist
- [x] Add a Changelog entry


danielballan:
Testing this branch against the `tiled.nsls2.bnl.gov`, it breaks on commit a8b05e5 but works on commit 04c68a1.

PR #873, opened 2025-01-29
Push image and chart only on tags

DiamondJoseph:
Prevents issues with the current CI pipeline failing on non-tagged commits: previously were building a container image on every push, with a generated tag. Now only publish the image and helm chart only when explicitly tagged, but continue to build on every commit.

### Checklist
- [ ] Add a Changelog entry


danielballan:
I rebased on `main` and moved the CHANGELOG entry up to the latest `Unreleased` section.

danielballan (approved):

DiamondJoseph:
```
Run docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671
Error: Input does not meet YAML 1.2 "Core Schema" specification: push
Support boolean input list: `true | True | TRUE | false | False | FALSE`
```

Looks like not quite right let me push another fix

PR #874, opened 2025-01-29
Use HTTPS if available

nmaytan:
This change makes a newly created Tiled client switch to HTTPS, if available, if constructed using an HTTP argument as parameter.

This serves two purposes:
* Prevents that client from hitting an HTTP->HTTPS redirect on every request
* Prevents certain requests from breaking for clients created using an HTTP URL (such as a POST when redirects are setup to give 301 responses)

For example, here is a POST request to logout resulting in only one POST request, instead of a POST followed by redirected GET as would happen if HTTP were used:
```
In [4]: c = from_uri("http://tiled.nsls2.bnl.gov")

[...]

In [5]: c.logout()
14:57:39.140 -> (204 B) POST 'https://tiled.nsls2.bnl.gov/api/v1/auth/session/revoke' 'host:tiled.nsls2.bnl.gov' 'accept:*/*' [...]
```

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -386,12 +386,13 @@ class Context:
         uri = httpx.URL(uri)
         node_path_parts = []
         # Ensure that HTTPS is used if available
-        # Rely on location header from HTTP->HTTPS redirect to determine if exists
+        # Logic will follow only one redirect, it is intended ONLY to toggle HTTPS.
+        # The redirect will be followed only if the netloc host is identical to the original.
         if uri.scheme == "http":
-            redirect_header = httpx.get(uri).headers.get("location", None)
-            if redirect_header is not None:
-                redirect_uri = httpx.URL(redirect_header)
-                if redirect_uri.scheme == "https":
+            response_from_http = httpx.get(uri)
+            if response_from_http.is_redirect:
+                redirect_uri = httpx.URL(response_from_http.headers["location"])
+                if redirect_uri.scheme == "https" and redirect_uri.host == uri.host:
                     uri = redirect_uri
         if "/metadata" in uri.path:
             api_path, _, node_path = uri.path.partition("/metadata")
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -385,6 +385,14 @@ class Context:
         """
         uri = httpx.URL(uri)
         node_path_parts = []
+        # Ensure that HTTPS is used if available
+        # Rely on location header from HTTP->HTTPS redirect to determine if exists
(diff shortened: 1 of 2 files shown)


tiled/client/context.py
```diff
@@ -385,6 +385,14 @@ def from_any_uri(
         """
         uri = httpx.URL(uri)
         node_path_parts = []
+        # Ensure that HTTPS is used if available
+        # Rely on location header from HTTP->HTTPS redirect to determine if exists
+        if uri.scheme == "http":
+            redirect_header = httpx.get(uri).headers.get("location", None)
```
danielballan, line 391: This seems like an appropriate place to use `response.is_redirect()` and/or `get(..., follow_redirects=True)`. It is not too unusual to encounter a chain of multiple redirects, and the latter would transparently deal with that (and appropriately error out if we're in a redirect loop / too many redirects).

nmaytan: Okay, I agree that's a good point. I suppose the question is how specific this should be to just converting to `https`. AFAICT http -> https redirect is typically just one hop. Perhaps more generally,  what we want is to consume any redirects up front - which includes https - but could actually be more. I think that's possibly convincing.

But are we concerned at all with blindly following redirects? Following a single redirect, limited to http->https schemes specifically, is at least quite targeted and inline with what seems to be standard practice for various tools. Following all redirects, some of which might be http->http redirects, is possibly less secure? Perhaps we insist that all redirects are https only?

Also, if we are to follow all redirects, this makes me think that perhaps the context should remember the original URL/chain of URLs for reference, to help avoid the "What? That's not the URL I typed!" reaction that I would myself have, eventually, if the redirect URL were significantly different.

danielballan: Notes from discussion:
- Use `is_redirect()` but not `follow_redirects`.
- Ensure domain is not changed.
- Look into HSTS in future PR.

danielballan:
Tested interactively.

Before:

```py
In [3]: c = from_uri("http://tiled.nsls2.bnl.gov")  # HTTP

In [6]: c.context.api_uri
Out[6]: URL('http://tiled.nsls2.bnl.gov/api/v1/')  # still HTTP
```

After:

```
In [2]: c = from_uri("http://tiled.nsls2.bnl.gov")  # HTTP

In [3]: c.context.api_uri
Out[3]: URL('https://tiled.nsls2.bnl.gov/api/v1/')  # HTTPS
```

danielballan (approved):

PR #875, opened 2025-01-31
Remove pydantic-based Definitions of Structures

genematx:
Since the [issue in pydantic](https://github.com/samuelcolvin/pydantic/issues/3251) has been fixed, we can now use dataclasses-based strcutures for arrays, tables, etc.

### Checklist
- [x] Add a Changelog entry
- [x] ~~Add the ticket number which this PR closes to the comment section~~


danielballan:
For posterity, this is the justification for maintaining pydantic copies of these classes, pulled from a module docstring, which no longer applies:


> The ArrayStructure class in tiled.structure.array is implemeneted with Python
built-in dataclasses. This is an implementation of the same structure in pydantic.
FastAPI requires pydantic models specifically, and will attempt to automatically
convert dataclasses to pydantic models. This normally works fine, but in the
case of ArrayStructure specifically it fails because of this issue:

> https://github.com/samuelcolvin/pydantic/issues/3251

> Once this problem is fixed in pydantic, this separate implementation in
ArrayStructure will no longer by needed. We will turn this module into a shim
that just imports ArrayStructure from tiled.structures.array and then deprecate
it.

PR #876, opened 2025-02-03
Remove branching imports for Python <3.9

DiamondJoseph:
As support for python < 3.9 has been dropped, removed the branching imports for versions < 3.9.

This allows defining a type hint for `tree` in `build_app`: of Mapping[str, Any]

It may also be possible to make the FastAPI instance app generic[T] on Mapping[str, T]?

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/__init__.py
+++ b/tiled/__init__.py
@@ -1,1 +1,6 @@
-from ._version import __version__, __version_tuple__  # noqa: F401
+from ._version import __version__, __version_tuple__
+
+__all__ = [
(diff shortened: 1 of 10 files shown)


danielballan (approved):

danielballan:
Let's follow up with typing the FastAPI app in the wake of the giant restructuring PR. I don't want to unnecessarily make a rebase mess. Merging this one as is.

PR #890, opened 2025-02-18
In AuthN database, use localized datetimes

danielballan:
In https://github.com/bluesky/tiled/pull/842 we followed Python's recommendation to use localized timestamps (`datetime.now(timezone.utc)`) instead of naive (`datetime.utcnow()`). This created an incompatibility with existing AuthN PG databases, when SQLAlchemy tried to insert localized datetimes into naive datetime columns. (The catalog databases are not affected by the change.)

After discussion with @dylanmcreynolds, @padraic-shafer, and @callumforrester, the consensus was that the right choice was to store localized timestamps. "Label your units!" is a good rule, timestamps are common source of confusion, and this could be helpful for developers inspecting the database. This PR:

- Changes the table definitions to use `DateTime(timezone=True)` in all cases. They formerly used `DateTime(timezone=False`) in all cases.
- Adds a migration script to update existing DATETIME columns from naive UTC to explicitly-localized UTC
- Maintains existing `NULL / NOT NULL` designations, except for `time_created`, which is switch from `NULL` to `NOT NULL`. (It always gets a server-provided default value at creation time.)

Note that, in SQLite, SQLAlchmy stores UTC-localized timestamps as naive strings, so the migration only affects SQLite

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,10 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ## Unreleased
 
-
 ### Added
 
-- Added `SQLAdapter` which can save and interact with table structured data in `sqlite` , `postgresql` and `duckdb` databases using `arrow-adbc` API calls.
+- Added `SQLAdapter` which can save and interact with table structured data in
+  `sqlite` , `postgresql` and `duckdb` databases using `arrow-adbc` API calls.
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


danielballan:
Tested upgrade and downgrade multiple times, with various spot checks. Example of upgraded API keys table:

```
database=# select time_created, time_updated, latest_activity, expiration_time from api_keys;
         time_created          |         time_updated          |    latest_activity     |    expiration_time
-------------------------------+-------------------------------+------------------------+------------------------
 2025-02-18 20:26:35.521412+00 | 2025-02-18 20:26:52.457578+00 | 2025-02-18 20:26:52+00 | 2025-02-20 00:13:15+00
```

Note `+00`.

danielballan:
Having cleared the general approach with others (tagged above) I'm going to self-merge this because I'm in a hurry to release and deploy during an NSLS-II maintenance window.

Issue #887, opened 2025-02-14
Epic: Refactor authentication and authorization


DiamondJoseph:
From discussion earlier in the week, I think our ideal state would be one where we have two protocols or abstract base classes that handle all of the logic currently expressed in various Dependencies on the FastAPI app. One Authenticators and one Authorizer, which are then configured in the Settings object on the server configuration, are passed into router construction as dependencies that will discover who a user is (if required) and what a user can do (if required).

```python
class Authenticator(ABC):
    @abstractmethod
    def authenticate(request: Request) -> UserSessionState:
        ...

class APIAuthenticator(Authenticator):
    def authenticate(request: Request) -> UserSessionState:
        api_key = get_api_key(request)
        if not api_key:
            raise NotAuthenticatedError("No API key in request")
        if api_Key not in known_users:
            raise NotAuthenticatedError("API key not recognised")
        return UserSessionState(known_users[api_key])

class OIDCAuthenticator(Authenticator):
    def authenticate(request: Request) -> UserSessionState:
        jwt = self.jwt_client.decode(request.headers["Authorization"])
        return UserSessionState(jwt["sub"])
```

```python
class Authorizer(ABC):
    @abstractmethod
    def authorize(<unknown signature>) -> bool:
        ...

class ExternalAuthorizer(Authorizer):
    def authorize(<unknown signature>) -> bool:
        return self.external_service.authorize(foo, bar)
```

danielballan:
I like this. Two thoughts:

1. We currently support multiple IdPs. Everyone who I've discussed it with seems to agree that reducing it to one (which may in turn support N if you like) is a good idea. But it's a sweeping change that will have to be done carefully.
2. We can "Authorizer" `AccesPolicy` and it's in `tiled.access_policy`. Work is in progress (by @nmaytan) to make these a bit nicer.

PR #877, opened 2025-02-03
Add Authenticator for when serving OIDC as a proxy

DiamondJoseph:
WIP: adds an Authenticator implementation for use when serving Tiled behind a proxy.
This is intended be used to define an alternative `decode_access_token`
In order to inject the desired behaviour for decode_access_token has required a fairly weighty refactor, inverting the creation order of a lot of router objects.

- Removes injection of password into security obj
- Removes use of dependency_override which is intended for use in tests

### Checklist
- [ ] Add a Changelog entry
- [ ] Add the ticket number which this PR closes to the comment section

PR #891, opened 2025-02-19
Refactor auth

DiamondJoseph:
This is a PR to pick apart the changes required before #877  into more managably reviewable commits.
This is a step towards closing #887

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -10,7 +10,7 @@ import warnings
 from contextlib import asynccontextmanager
 from functools import cache, partial
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 
 import anyio
 import packaging.version
@@ -34,15 +34,18 @@ from starlette.status import (
     HTTP_500_INTERNAL_SERVER_ERROR,
 )
 
+from tiled.query_registration import QueryRegistry
 from tiled.server.authentication import move_api_key
 from tiled.server.protocols import ExternalAuthenticator, InternalAuthenticator
 
 from ..config import construct_build_app_kwargs
 from ..media_type_registration import (
-    compression_registry as default_compression_registry,
+    CompressionRegistry,
+    SerializationRegistry,
+    default_compression_registry,
 )
 from ..utils import SHARE_TILED_PATH, Conflicts, SpecialUsers, UnsupportedQueryType
-from ..validation_registration import validation_registry as default_validation_registry
+from ..validation_registration import ValidationRegistry, default_validation_registry
 from . import schemas
 from .compression import CompressionMiddleware
 from .dependencies import (
@@ -113,10 +116,11 @@ def build_app(
     tree,
     authentication=None,
     server_settings=None,
-    query_registry=None,
-    serialization_registry=None,
-    compression_registry=None,
-    validation_registry=None,
+    query_registry: Optional[QueryRegistry] = None,
+    serialization_registry: Optional[SerializationRegistry] = None,
+    deserialization_registry: Optional[SerializationRegistry] = None,
+    compression_registry: Optional[CompressionRegistry] = None,
+    validation_registry: Optional[ValidationRegistry] = None,
     tasks=None,
     scalable=False,
 ):
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -16,7 +16,7 @@ import anyio
 import packaging.version
 import yaml
 from asgi_correlation_id import CorrelationIdMiddleware, correlation_id
-from fastapi import APIRouter, FastAPI, HTTPException, Request, Response, Security
+from fastapi import APIRouter, Depends, FastAPI, HTTPException, Request, Response
 from fastapi.exception_handlers import http_exception_handler
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.openapi.utils import get_openapi
@@ -34,6 +34,7 @@ from starlette.status import (
     HTTP_500_INTERNAL_SERVER_ERROR,
 )
 
+from tiled.server.authentication import move_api_key
 from tiled.server.protocols import ExternalAuthenticator, InternalAuthenticator
 
 from ..config import construct_build_app_kwargs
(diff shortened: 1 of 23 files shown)


tiled/server/app.py
```diff
@@ -113,10 +112,11 @@ def build_app(
     tree,
     authentication=None,
     server_settings=None,
-    query_registry=None,
-    serialization_registry=None,
-    compression_registry=None,
-    validation_registry=None,
+    query_registry: Optional[QueryRegistry] = None,
```
danielballan, line 115: I wonder if there's any reason not to make the default values the default registry objects, and drop the `Optional`.

tiled/server/authentication.py
```diff
@@ -93,33 +92,14 @@ class TokenData(BaseModel):
...
-        scheme_name: Optional[str] = None,
-        description: Optional[str] = None,
-    ):
-        self.model: APIKey = APIKey(
-            **{"in": APIKeyIn.header}, name=name, description=description
-        )
-        self.scheme_name = scheme_name or self.__class__.__name__
-
+# TODO: remove custom subclass https://github.com/bluesky/tiled/issues/921
+class StrictAPIKeyHeader(APIKeyHeader):
```
danielballan, line 96: A docstring to describe the purpose of this subclass (What makes it `Strict`?) would be helpful here.

tiled/server/authentication.py
```diff
@@ -233,6 +218,92 @@ async def get_session_state(decoded_access_token=Depends(get_decoded_access_toke
...
+        )
+
+
+async def get_scopes_from_api_key(
+    api_key: str, settings: Settings, authenticators, db
+) -> Sequence[str]:
+    if not authenticators:
+        # Tiled is in a "single user" mode with only one API key.
+        return (
+            {
```
danielballan, line 236: While we're here, I wonder if this set should be defined as a global variable, `SINGLE_USER_SCOPES`, to improve readability.

tiled/server/authentication.py
```diff
@@ -233,6 +218,92 @@ async def get_session_state(decoded_access_token=Depends(get_decoded_access_toke
...
+    settings: Settings = Depends(get_settings),
+    authenticators=Depends(get_authenticators),
+    db=Depends(get_database_session),
+) -> set[str]:
+    if api_key is not None:
+        return await get_scopes_from_api_key(api_key, settings, authenticators, db)
+    elif decoded_access_token is not None:
+        return decoded_access_token["scp"]
+    else:
+        return {"read:metadata", "read:data"} if settings.allow_anonymous_access else {}
```
danielballan, line 287: Likewise, maybe define this as `ANONYMOUS_USER_SCOPES` for additional clarity.

tiled/server/authentication.py
```diff
@@ -233,6 +218,92 @@ async def get_session_state(decoded_access_token=Depends(get_decoded_access_toke
...
+
+async def check_scopes(
+    request: Request,
+    security_scopes: SecurityScopes,
+    scopes: set[str] = Depends(get_current_scopes),
+) -> None:
+    if not set(security_scopes.scopes).issubset(scopes):
+        raise HTTPException(
+            status_code=HTTP_401_UNAUTHORIZED,
+            detail=(
```
danielballan, line 298: Maybe this should be a separate issue, but I've often wished that this message differentiated between an unauthenticated request ("Maybe you need to log in?") and an authenticated one (current message is fine).

tiled/server/router.py
```diff
@@ -67,357 +66,436 @@
...
+    query_registry: QueryRegistry,
+    serialization_registry: SerializationRegistry,
+    deserialization_registry: SerializationRegistry,
+    validation_registry: ValidationRegistry,
+) -> APIRouter:
+    router = APIRouter()
+
+    T = TypeVar("T")
+
+    def patch_route_signature(route: Callable[..., T]) -> Callable[..., T]:
```
danielballan, line 80: As written here, this closes over `query_registry`. Another option would be to define this function at global scope and make it a higher-order decorator that takes `query_registry` as an argument.

Either is fine with me, not sure which I prefer, just considering options out loud.

tiled/server/router.py
```diff
@@ -67,357 +66,436 @@
...
+        select_metadata: Optional[str] = Query(None),
+        offset: Optional[int] = Query(0, alias="page[offset]", ge=0),
+        limit: Optional[int] = Query(
+            DEFAULT_PAGE_SIZE, alias="page[limit]", ge=0, le=MAX_PAGE_SIZE
+        ),
+        sort: Optional[str] = Query(None),
+        max_depth: Optional[int] = Query(None, ge=0, le=DEPTH_LIMIT),
+        omit_links: bool = Query(False),
+        include_data_sources: bool = Query(False),
+        entry: MapAdapter = Security(get_entry(), scopes=["read:metadata"]),
```
danielballan, line 260: This is nice. :+1:

danielballan (commented):
First pass. Echoing my comments on Slack, I love this. It's great to have another pair of eyes on this.

tiled/server/app.py
DiamondJoseph: Is `build_app` ever called not with the result of `construct_build_app_kwargs`? Just thinking about where the defaults should be injected.

Although ideally I'd like to get rid of `construct_build_app_kwargs` and anything that uses kwargs...

tiled/server/authentication.py
DiamondJoseph: Add a USER_SCOPES, not attached to the name if you think it's important to call out this is for the TILED_SINGLE_USER_API_KEY case.

tiled/server/authentication.py
DiamondJoseph: Leaving for a separate issue as it will change test behaviour.

danielballan: https://github.com/bluesky/tiled/issues/925

tiled/server/app.py
danielballan: Yes, a whole bunch in the test suite. One example at random:

https://github.com/bluesky/tiled/blob/ed9033801890f3db735e76915bf6dec96726890a/tiled/_tests/test_distinct.py#L65

DiamondJoseph: I'm going to push back on changing it for now, because I'd like to rip out construct_build_app_kwargs in the longer run, make build_app take just tree and settings, and eventually settings gets created on startup exactly once.

danielballan (approved):

PR #892, opened 2025-02-19
Run authN tests against PG as well.

danielballan:
Why didn't the bug fixed in #890 fail our test suite? The reason is, we were testing auth against SQLite only, and the issue affected PostgreSQL only. This PR parameterizes the core `config` fixture used by these tests to run against both.

### Checklist
- [x] Add a Changelog entry


padraic-shafer (approved):
Good reuse of test fixtures and patterns to guard against regression.

PR #895, opened 2025-02-20
feat: add option for additional default in json serializer

maffettone:
This addresses the change in `orjson` that no longer serializes NamedTuples by default to dictionaries. This is potentially the correct behavior, so the approach taken was to give an option for additional defaults in the serializer (which already does some work for numpy).
A related [PR in databroker ](https://github.com/bluesky/databroker/pull/837) takes advantage of this. 

### Checklist
- [x] Add a Changelog entry
- [x] Add the ticket number which this PR closes to the comment section (None present)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -389,6 +389,7 @@ Write the date in place of the "Unreleased" in the case a new version is release
 - A new method `DataFrameClient.append_partition`.
 - Support for registering Groups and Datasets _within_ an HDF5 file
 - Tiled version is logged by server at startup.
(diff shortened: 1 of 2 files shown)


maffettone:
@ambarb for visibility

danielballan:
As discussed in the [companion PR to databroker](https://github.com/bluesky/databroker/pull/837) @maffettone and I agree the best option is to fix this in the device, to use a serializable type.

maffettone:
Yes, I thought I put an issue there, but on double checking it was a different issue. I'll add something this morning.

PR #897, opened 2025-02-22
ENH: Refactor HDF5 Adapter

genematx:
This PR refactors the HDF5Adaptor to allow reading datsets split across multiple HDF5 files. The file structure is parsed on init and stored as a nested dictionary in the class atribute `self._tree`, which avoids keeping the HDF5 open while Adapter is in use. When needed, the arrays are read using `dask.delayed`.

Furthermore, added the ability to traverse HDF5 files with broken external and internal (soft) links without raising the error until the broken link/node is requested.

### Checklist
- [ ] Add a Changelog entry


tiled/adapters/hdf5.py
```diff
@@ -13,25 +20,303 @@
...
+    swmr: bool = SWMR_DEFAULT,
+    libver: str = "latest",
+    **kwargs: Optional[Any],
+) -> JSON:
+    """Get attributes of an HDF5 dataset"""
+    file_path = path_from_uri(file_uri)
+    with h5py.File(file_path, "r", swmr=swmr, libver=libver, **kwargs) as _file:
+        node = _file[dataset] if dataset else _file
+        d = dict(getattr(node, "attrs", {}))
+        for k, v in list(d.items()):
```
danielballan, line 108: I think this `list` call would unnecessarily traverse this an extra time.

```suggestion
        for k, v in d.items():
```

tiled/adapters/hdf5.py
```diff
@@ -13,25 +20,303 @@
...
 from .array import ArrayAdapter
-from .resource_cache import with_resource_cache
 
 SWMR_DEFAULT = bool(int(os.getenv("TILED_HDF5_SWMR_DEFAULT", "0")))
 INLINED_DEPTH = int(os.getenv("TILED_HDF5_INLINED_CONTENTS_MAX_DEPTH", "7"))
 
 
-def from_dataset(dataset: NDArray[Any]) -> ArrayAdapter:
-    return ArrayAdapter.from_array(dataset, metadata=getattr(dataset, "attrs", {}))
+def ndslice_from_string(
```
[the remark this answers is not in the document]
genematx: I actually have an entire `NDSlice` class for serializing slices in one of [other PRs](https://github.com/bluesky/tiled/blob/1584cfeb0b17453c84ce53c4788a0573adb7ea38/tiled/type_aliases.py#L21). Do you think it would be worth bringing it here (with a few changes)?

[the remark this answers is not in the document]
genematx: Done

genematx:
Sounds great, @toqduj, thanks! I've put together a few checks for broken links in HDF5 files: now it should return 410 if an external file has been removed or a soft link has broke _after_ an adapter has been instantiated. Otherwise -- it will raise the usual KeyError if a user tries to access a non-existing dataset.

tiled/_tests/test_access_control.py
```diff
@@ -43,9 +43,6 @@ async def allowed_scopes(self, node, principal, path_parts):
                 for i, segment in enumerate(path_parts):
                     if hasattr(node, "lookup_adapter"):
                         node = await node.lookup_adapter(path_parts[i:])
```
genematx, line 45: `lookup_adapter` will now either return a valid Adapter or raise a KeyError (NoEntry or BrokenLink)

tiled/catalog/adapter.py
```diff
@@ -450,7 +450,7 @@ async def lookup_adapter(
                     for segment in segments[i:]:
                         adapter = await anyio.to_thread.run_sync(adapter.get, segment)
```
genematx, line 451: Some downstream adapters may still return `None`, but HDF5 will either raise KeyError or BrokenLink if `segment` is not found.

toqduj:
It seems you have to change the requirements for the task package. It currently installs 2024.something, but you are relying on a later version. I get things to work with dask >= 2025.2.0

toqduj:
That said, tree(client) does not work on either plain Eiger files, nor my files. So perhaps 2025.2.0 is not right either.. 
```bash
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bpauw/Code/Julias_camonitor_log/.venv/lib/python3.12/site-packages/dask/array/core.py", line 4476, in concatenate
    meta = _concatenate(seq_metas, axis=axis)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: zero-dimensional arrays cannot be concatenated
```

tiled/_tests/test_array.py
```diff
@@ -170,3 +171,35 @@ def test_array_interface(context):
...
     assert len(output) == len(bytes(output))
+
+
+slice_cases = [
+    ("[0:5:2]", (slice(0, 5, 2),), [{"start": 0, "stop": 5, "step": 2}]),
+    ("(1:3)", (slice(1, 3),), [{"start": 1, "stop": 3}]),
+    ("1:", (slice(1, None),), [{"start": 1}]),
+    ("[0:3]", (slice(0, 3),), [{"start": 0, "stop": 3}]),
+    ("[:]", (slice(None, None, None),), [{}]),
+    ("[...]", (Ellipsis,), ["ellipsis"]),
```
danielballan, line 182: Should `[:]` and `[...]` render equivalent JSON, `{}`? I cannot think of a semantic distinction between the two.

tiled/adapters/array.py
```diff
@@ -104,7 +118,7 @@ def structure(self) -> ArrayStructure:
 
     def read(
         self,
-        slice: NDSlice = ...,
+        slice: Optional[NDSlice] = None,
```
danielballan, line 121: We originally used `None` as the default slice but switched to using `...`. Using `None` as a sentinel for "no slicing" is potentially confusing, because slicing with `None` actually means something else in numpy:

```py
In [6]: a = np.ones(3)

In [7]: a
Out[7]: array([1., 1., 1.])

In [8]: a[None]
Out[8]: array([[1., 1., 1.]])
```

Is there a new consideration that should prompt us to considering using `None` instead of `...`?

tiled/adapters/sequence.py
```diff
@@ -199,7 +201,7 @@ def read(self, slice: Optional[NDSlice] = ...) -> NDArray[Any]:
         return force_reshape(arr, sliced_shape)
 
     def read_block(
-        self, block: Tuple[int, ...], slice: Optional[NDSlice] = ...
+        self, block: Tuple[int, ...], slice: Optional[NDSlice] = None
```
danielballan, line 204: As above, we should consider keeping `...`.

tiled/adapters/sparse_blocks_parquet.py
```diff
@@ -129,9 +129,9 @@ def write_block(
         self,
         data: Union[dask.dataframe.DataFrame, pandas.DataFrame],
         block: Tuple[int, ...],
-        slice: NDSlice = ...,
+        slice: Optional[NDSlice] = None,
```
danielballan, line 132: As above, we should consider keeping `...`.

tiled/adapters/zarr.py
```diff
@@ -76,7 +76,7 @@ def _stencil(self) -> Tuple[slice, ...]:
 
     def read(
         self,
-        slice: NDSlice = ...,
+        slice: Optional[NDSlice] = None,
```
danielballan, line 79: As above, we should consider keeping `...`.

tiled/type_aliases.py
```diff
@@ -10,15 +12,132 @@
...
-from typing import Any, Dict, List, Set, Tuple, Union
+from typing import Any, Dict, List, Set, Union
 
 JSON = Dict[str, Union[str, int, float, bool, Dict[str, "JSON"], List["JSON"]]]
-NDSlice = Union[
-    int, slice, Tuple[Union[int, slice, EllipsisType], ...], EllipsisType
-]  # TODO Replace this with our Union for a slice/tuple/.../etc.
+
+
+class NDSlice(tuple):
```
danielballan, line 20: This is not a type alias anymore; maybe it should go in a new module.

I wonder what the trade-offs would be if this were implemented as functions that call [`ndindex`](https://quansight-labs.github.io/ndindex/api/ndindex.html) internally.

tiled/adapters/hdf5.py
```diff
@@ -11,28 +17,313 @@
...
+        dataset: Optional[str] = None,
+        slice: Optional[Union[str, NDSlice]] = None,
+        squeeze: Optional[bool] = False,
+        swmr: bool = SWMR_DEFAULT,
+        libver: str = "latest",
+        **kwargs: Optional[Any],
+    ) -> "HDF5ArrayAdapter":
+        structure = data_source.structure
+        assets = data_source.assets
+        data_uris = [ast.data_uri for ast in assets if ast.parameter == "data_uri"] or [
```
danielballan, line 226:
```suggestion
        data_uris = [ast.data_uri for ast in assets if ast.parameter == "data_uris"] or [
```

Right? To match the tests?

danielballan (commented):
First detailed pass. I need to test-drive it too.

tiled/_tests/test_array.py
danielballan: If we ever add schemas (pydantic or JSON schema) to these, an array of objects with three optional keys (start, stop, step) is a bit nicer than an array of items that may be objects or may be the literal string `ellipsis`.

tiled/adapters/array.py
genematx: True. I was bitten by slicing with `None` before.
I think, changing `...` to `None` was due to mypy complaints. I'm sure there's another workaround; I'll have a look.

tiled/_tests/test_array.py
genematx: There's a difference between `[:]` and `[...]` when you slice the middle part of a (very) high-dimensional array, for example:
```
In [18]: x.shape
Out[18]: (3, 4, 5, 6, 7)

In [20]: x[:2, :, :2].shape
Out[20]: (2, 4, 2, 6, 7)

In [21]: x[:2, ..., :2].shape
Out[21]: (2, 4, 5, 6, 2)
```

danielballan: Ah, I see. I wonder if as part of the JSON encoding we should normalize to an expanded notation where the length of the JSON array always matches the number of dimensions, so the above example would normalize to `[{"stop": 2}, {}, {}, {}, {"stop": 2}]`. A nice feature of this is that a given slice has a unique representation.

This is less compact, but we don't need to be concerned about the dimensionality getting _too_ high. HDF5 has a hard limit at 64 dimensions. (And Tiled could too, if we like.)

genematx: I like your idea, thank you! So much better than having a `Literal` string in json, even if less compact.

genematx: On the other hand, does it mean, we'll always need to pass the number of dimensions to `NDSlice` (or at least when converting from a string that uses Ellipsis)? I'd rather avoid this. I definitely see the issue with the literal and will try to figure something out.

genematx: A possible workaround, we could encode `...` in JSON as `{"step":0}`: `builtins.slice(None, None, 0)` or `::0` are not allowed, so it would be just an internal representation to keep the json schema simple. A bit weird, but otoh $\frac{stop-start}{step=0}\rightarrow\infty, which is almost what `Ellipsis` means.

danielballan: That's an interesting argument. Maybe a bit too clever, but maybe not.

What are the trade-offs of making `NDSlice` take the shape? Could it be useful to have it verify that the slice is valid for the shape of the array, as well as providing an explicit JSON?

danielballan: Yes, sorry, my comment was not quite clear. Yes, I agree we don't want to make it different than Python/numpy, and your use case is a good one: we might not know the (variable) shape. Passing a `shape` or at least an `ndim` to `to_json` seems right.

tiled/adapters/zarr.py
genematx: fixed

tiled/adapters/sparse_blocks_parquet.py
genematx: fixed

tiled/_tests/test_array.py
genematx: Done.

tiled/type_aliases.py
genematx: I've moved it to its own `tiled/ndslice.py`. I've also replaced the existing `slice_` in dependencies.py with `NDSlice` objects (while keeping the tests) as well.

tiled/adapters/hdf5.py
genematx: I'm not sure... I think this touches on the case when we may have multiple assets with possibly varying filetypes in the same DataSource. Correct me if I'm wrong, but this parameter would normally be set externally by Bluesky in the Resource document, right?
That being said, one of them should change (either here in the tests), of course. Thank you for catching it!

danielballan: My understanding is, `data_source.parameters` comes directly from the Stream Resource document. But the `asset_data_source_relation.parameter` is the internal business of the `HDF5Adapter`, at least to some degree.

PR #898, opened 2025-02-24
started enlarging test cases for sql adapter

skarakuzu:
### Checklist
- [ ] Add a Changelog entry


danielballan (approved):
We should certainly look into the oddity with SQLite (and consider suspending our support for it). But that can be a separate PR; this is all a great improvement.

Issue #824, opened 2024-12-06
Revived proposal for representing BlueskyRuns in Tiled


danielballan:
Brief summary of discussions with @padraic-shafer, @genematx, and @tacaswell, drawing on earlier discussions with @dylanmcreynolds and @whs92.

---

Goals:

- Provide a stable URL to a given column of data, regardless of whether it is in-line in the Event documents or external.
- Data in a given stream should be presented to the user in a flat namespace, not placing the URL path the details about internal versus external storage.
- Server should reveal to clients how the data is grouped into tables and arrays underneath, facilitating efficient chunked access (i.e. not separate requests for each column).
- It should be possible to add "views", such as simplified/flatter views, of the data in a BlueskyRun. (NeXus does something similar.) A canonical example is re-mapping a mapping scan (1D in the canonical Bluesky representation) into a more scientifically relevant N-dimensional view (reshaping).
- Lay track for potentially giving a way to access auxiliary information (e.g. proxied from Archiver Appliance).

Revive https://github.com/bluesky/tiled/pull/668 which proposed two ideas.
- A new structure family named ~union~ "consolidated" which let us represent a Bluesky Stream, backed my a combination of tables and arrays, nicely
- A concept of "views", alternate (simplified...) layouts of the canonical data

Both suggestions were developed and viewed positively at the time, but set aside back in March to make progress on other things. Now, we propose to implement both.

In URLs:

```
# Bluesky stuff...
.../<uuid>/streams/<stream name>  # "consolidated" structure family
.../<uuid>/streams/<stream name>?part=<part>  # table of columns stored together, or array
.../<uuid>/streams/<stream name>/<key>  # array
.../<uuid>/config/<stream name>/<obj name>  # table of configuration readings

# New concept: simplified/rearranged "views":
.../<uuid>/views/<view name>/...  # server-side registerable views

# New concept: auxiliary info (e.g. from EPICS Archiver Appliance)
../<uuid>/auxiliary/<aux name>/...  # information from outside Tiled

```

In Python API:

```py
run.streams["primary"].read()  # xarray.Dataset
run.streams["primary"]["image"]  # array
run.config["primary"]["fccd"].read()  # table of configuration readings

run.views["simple"].read()
# and equivalently...
run.read(view="simple")  # might even be the default for run.read(), which currently raises...
run.aux["archiver"]["PV..."].read()  # array
```

dylanmcreynolds:
[...]

PR #901, opened 2025-02-28
Client-side flattening of a container

genematx:
A third attempt at implementing a flat-namespaced container, this time (mostly) on the client-side. While the data is stored in a new `Composite` container, the flat namespace is implemented by the python client itself: the client keeps a mapping of table column names to their full paths. The distinct structure family serves a role akin to a spec here, but its functionality is thought to be expanded in the future.

Related: #668
Issue: #824

### Checklist
- [x] Add a Changelog entry


tiled/client/container.py
```diff
@@ -1051,6 +1050,57 @@ def write_dataframe(
         return client
 
 
+class Composite(Container):
+
+    @property
```
genematx, line 1055: possibly cached?

tiled/client/container.py
```diff
@@ -689,9 +688,8 @@ def new(
     # to attempt to avoid bumping into size limits.
     _SUGGESTED_MAX_UPLOAD_SIZE = 100_000_000  # 100 MB
 
-    def create_container(self, key=None, *, metadata=None, dims=None, specs=None):
-        """
-        EXPERIMENTAL: Create a new, empty container.
+    def create_container(self, key=None, *, metadata=None, specs=None, flat=False):
```
danielballan, line 691: I lean toward, "This should be a new method, `create_composite`." We can always add this `flat=True` alias later. What do you think?

tiled/catalog/adapter.py
```diff
@@ -634,7 +634,10 @@ async def create_node(
             await db.refresh(node)
             for data_source in data_sources:
                 if data_source.management != Management.external:
-                    if structure_family == StructureFamily.container:
+                    if structure_family in [
```
danielballan, line 637: Here and elsewhere below, it may be slightly better to make this a `set`. Negligible performance improvement, but a bit cleaner IMO.

tiled/structures/core.py
```diff
@@ -16,6 +16,7 @@ class StructureFamily(str, enum.Enum):
     array = "array"
     awkward = "awkward"
     container = "container"
+    composite = "composite"
```
danielballan, line 19: Bump this up one row to maintain alphabetical sort order.

tiled/client/container.py
```diff
@@ -1051,6 +1049,106 @@ def write_dataframe(
         return client
 
 
+class Composite(Container):
```
danielballan, line 1052: This should override `create_container` (and also `create_composite` if we add it) to raise an exception. The server would reject the request (unsupported structure family) but it's probably worth catching it client side and making it clear in the docstring that this never works.

danielballan (commented):
Awesome, nice to see this converging.

I'd like to test-drive it a bit, which I haven't yet done, but I gave a close read.

tiled/client/container.py
genematx: I wasn't sure. Thanks! Will fix it.

tiled/structures/core.py
genematx: Done

tiled/client/container.py
genematx: Done

tiled/client/container.py
genematx: Done

tiled/catalog/adapter.py
genematx: Done!

danielballan:
This will also need a database migration to extend the `StructureFamilies` enum in PostgreSQL. This command will create a new stub file with a random name.

```
python -m tiled.catalog.main revision -m "Add composite structure family"
```

Add content with reference to this example: https://github.com/bluesky/tiled/blob/main/tiled/catalog/migrations/versions/0b033e7fbe30_add_awkward_to_structurefamily_enum.py

Then update the list of revisions in `tiled.catalog.core.py`.

danielballan:
Here is @genematx's test script from our interactive session today, for Future Us:

```py
import pandas as pd
import numpy as np
df = pd.DataFrame({"colA": np.random.randn(10), "colB": np.random.randint(0, 10, 10), "colC": np.random.choice(["a", "b", "c", "d", "e"], 10)})

Y = c.create_composite("test", metadata={"attrs": {"b": 2}})

Y.write_array(np.random.randn(10, ), key="arr1", metadata={"attrs": {"c": 3}})
Y.write_array(np.random.randn(10, ), key="arr2", metadata={"attrs": {"d": 4}})
Y.write_array(np.random.randn(20, ), key="arr3", metadata={"attrs": {"e": 5}})
Y.write_dataframe(df, key="tab1", metadata={"attrs": {"f": 6}})
```

---

I like how this turned out:

```py
In [35]: c['test']
Out[35]: <Composite {'arr1', 'arr2', 'arr3', 'colA', 'colB', 'colC'}>

In [36]: c['test']['tab1']
KeyError                                  Traceback (most recent call last)
Cell In[36], line 1
----> 1 c['test']['tab1']

File ~/Repos/bnl/tiled/tiled/client/container.py:1149, in Composite.__getitem__(self, key, _ignore_inlined_contents)
   1147     key = self._flat_keys_mapping[key]
   1148 else:
-> 1149     raise KeyError(
   1150         f"Key '{key}' not found. If it refers to a table, use .parts['{key}'] instead."
   1151     )
   1153 return super().__getitem__(key, _ignore_inlined_contents)

KeyError: "Key 'tab1' not found. If it refers to a table, use .parts['tab1'] instead."

In [37]: c['test'].parts['tab1']
Out[37]: <DataFrameClient ['colA', 'colB', 'colC']>
```

Perhaps we should make that error message even more generous by checking whether `'tab1'` is in fact in `parts`.

---

Soon we will need `c['test'].read()` (which I guess returns an `xarray.Dataset`) but since Composite is usable without this, I would be in favor making that a separate follow-up PR.

---

I haven't done line review yet, but conceptually I think this has landed.

danielballan:
I can think of other counterarguments, too. Like, maybe it's not in `parts` _yet_ but it will be a moment later. And the point you raise about caching would make that even more complicated.

Let's leave it as is, at least for now.

tiled/client/container.py
```diff
@@ -1051,6 +1082,125 @@ def write_dataframe(
...
+    def create_container(self, key=None, *, metadata=None, specs=None):
+        """Composite nodes can not include nested containers by design."""
+        raise NotImplementedError("Cannot create a container within a composite node.")
+
+    def create_composite(self, key=None, *, metadata=None, specs=None):
+        """Ccomposite nodes can not include nested composites by design."""
+        raise NotImplementedError("Cannot create a composite within a composite node.")
+
+
+class CompositeContents:
```
danielballan, line 1164: Should this be named `CompositeParts` for consistency?

```py
>>> x.parts
<CompositeParts ...>
```

The word `Contents` is used, as in `self._contents` to mean something slightly distinct, I think.

I wouldn't insist on it, though; if you prefer it as is, I'm good. :+1:

danielballan (commented):
I took one more detailed read through.

tiled/adapters/sql.py
```diff
@@ -195,46 +195,55 @@ def structure(self) -> TableStructure:
...
         Parameters
         ----------
         key : a string to indicate which column to be retrieved
+
         Returns
         -------
         The column for the associated key.
         """
+
+        # Construct the metadata for the array: assume that relevant metadata is in the `key` field
```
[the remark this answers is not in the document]
genematx: Agree. Leaving the column metadata empty and potentially offloading this to specs and structure-clients if needed in the future.

tiled/adapters/table.py
```diff
@@ -145,7 +145,19 @@ def __getitem__(self, key: str) -> ArrayAdapter:
             max_size = max((len(i) for i in array.ravel()))
             array = array.astype(dtype=numpy.dtype(f"<U{max_size}"))
 
-        return ArrayAdapter.from_array(array)
+        # Construct the metadata for the array: assume that relevant metadata is in the `key` field
```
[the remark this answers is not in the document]
genematx: Fixed

tiled/catalog/adapter.py
```diff
@@ -1503,4 +1573,5 @@ def specs_array_to_json(specs):
     StructureFamily.container: CatalogContainerAdapter,
     StructureFamily.sparse: CatalogSparseAdapter,
     StructureFamily.table: CatalogTableAdapter,
+    StructureFamily.composite: CatalogCompositeAdapter,
```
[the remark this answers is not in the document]
genematx: Done.

tiled/client/container.py
genematx: Totally, `CompositeParts` is much better. Changed!
Also, made `self.contents` and `self.get_contents` public attribute/method.

tiled/catalog/migrations/versions/45a702586b2a_add_composite_structure_family.py
```diff
@@ -0,0 +1,31 @@
...
+        with op.get_context().autocommit_block():
+            op.execute(
+                sa.text(
+                    "ALTER TYPE structurefamily ADD VALUE IF NOT EXISTS 'composite' AFTER 'awkward'"
+                )
+            )
+
+
+def downgrade():
+    raise NotImplementedError
```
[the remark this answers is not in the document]
genematx: Oh, nice! Added.
Please, doublecheck, @danielballan

PR #906, opened 2025-03-04
Simple tiled server

### LINKED ISSUE #826: Utility for launching tiled server in background for tutorials
For development or production use of tiled, it is best to run a tiled server in its own process, using the CLI or a container.

But for tutorials and quick demos, it is convenient to launch a tiled server in the same process, such as from a Jupyter notebook. Otherwise, the user has to switch contexts, open a terminal, paste in some code...it breaks the flow. Requiring a separately-run server process also does not play well with Jupyter notebook publication workflows.

The server could be run as a subprocess or a thread. My gut feeling is that thread is the better option, because we may find reasons to want to interact with the server and this will be much simpler if it lives in-process.

On a thread, it could communicate via ASGI (passing HTTP messages in Python) or via TCP. Here, I think TCP is better because the client workflow will be more realistic, more similar to connecting to a normal Tiled server. That is, `from_uri(...)` rather than `from_context(Context.from_app(app))`.

My main concern is that this will be overused, applied in contexts beyond a tutorial. To that end, I think it should use temporary files only, and it should be minimally configurable---perhaps not at all configurable. Perhaps it should also have `Temp` in the name, like `TempTiledServer()`

@genematx has a prototype over in https://github.com/bluesky/bluesky-cookbook/pull/24/files#diff-de29f45755c18a7cb50720ad79693f53741163f9236c5cf6eea33cdfb5dd0b64

danielballan:
Closes #826 

image

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/server/simple.py
+++ b/tiled/server/simple.py
@@ -1,4 +1,5 @@
 import contextlib
+import copy
 import pathlib
 import secrets
 import tempfile
@@ -41,8 +42,9 @@ class SimpleTiledServer:
         api_key: Optional[str] = None,
     ):
         # Delay import to avoid circular import.
-        from tiled.catalog import from_uri as catalog_from_uri
-        from tiled.server.app import build_app
+        from ..catalog import from_uri as catalog_from_uri
+        from .app import build_app
+        from .logging_config import LOGGING_CONFIG
 
         if dir_path is None:
             dir_path = pathlib.Path(tempfile.TemporaryDirectory(delete=False).name)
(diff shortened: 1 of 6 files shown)


danielballan:
Needs tests. I propose `SimpleTiledServer` instead of `TempTiledServer` because this can be used to serve persistent data, i.e. `SimpleTiledServer(dir_path='./my_data')`. I think qualifying it with _something_ (`Simple...`) is important but I don't want to give the sense that this is how we actually run tiled in prod.

tiled/server/simple.py
```diff
@@ -0,0 +1,102 @@
...
+        from ..catalog import from_uri as catalog_from_uri
+        from .app import build_app
+        from .logging_config import LOGGING_CONFIG
+
+        if dir_path is None:
+            dir_path = pathlib.Path(tempfile.TemporaryDirectory(delete=False).name)
+        else:
+            dir_path = pathlib.Path(dir_path)
+        dir_path.mkdir(parents=True, exist_ok=True)
+        api_key = api_key or secrets.token_hex(32)
```
genematx, line 54: I wonder if we could use a shorter API key for the "simple" server?

genematx (approved):

genematx (approved):

tiled/server/simple.py
danielballan: I had the same thought.

The only downside I can think of is if some developer sees it and gets a bad impression of our security-handling, because that key is shorter than what cryptography best practice would recommend. But using shorter random keys in dev-like scenarios is not uncommon.

So let's do it, commit incoming....

danielballan:
Added docstrings, tests, and some polish.

On `test_no_collisoins` I'm getting an error I cannot explain. As it only affects this test that runs two servers at once, I suspect some kind of unintended coupling between the two servers.

```
E               tiled.client.utils.ClientError: 401: Invalid API key http://127.0.0.1:33597/api/v1/metadata/?include_data_sources=false
```

genematx (approved):

PR #908, opened 2025-03-06
Refactor Settings creation

DiamondJoseph:
### Checklist
- [ ] Add a Changelog entry


tiled/server/settings.py
```diff
@@ -1,82 +1,50 @@
...
+from pydantic import Field
+from pydantic.dataclasses import dataclass
+from pydantic_settings import BaseSettings, SettingsConfigDict
 
-DatabaseSettings = collections.namedtuple(
-    "DatabaseSettings", "uri pool_size pool_pre_ping max_overflow"
-)
+
+# hashable cache key for use in tiled.authn_database.connection_pool
+@dataclass(unsafe_hash=True)
```
danielballan, line 13: I suspect we will not need this, once we merge the refactor that places the connection pool in `app.state` instead of a global dict keyed on this object.

danielballan (approved):
This is great. Ready to merge, IMO. Not sure which merge order will make things easiest for us, so I'll leave it as "approved" for now.

tiled/server/settings.py
```diff
@@ -1,82 +1,55 @@
...
-        )  # 7 days
-    )
-    session_max_age: Optional[timedelta] = timedelta(
-        seconds=int(os.getenv("TILED_SESSION_MAX_AGE", 365 * 24 * 60 * 60))  # 365 days
-    )
+    single_user_api_key: str = secrets.token_hex(32)
+    single_user_api_key_generated: bool = "TILED_SINGLE_USER_API_KEY" not in os.environ
+    # Must be a string representation of a json list, e.g. ["one", "two"]
+    # The first key will be used for encryption. Each key will be tried in turn for decryption.
+    secret_keys: List[str] = [secrets.token_hex(32)]
```
DiamondJoseph, line 31: Another breaking change- this would currently be TILED_SECRET_KEYS instead of TILED_SERVER_SECRET_KEYS

danielballan:
Part of me likes the old bash-style `;`-separated secret keys, rather than JSON, but I accept that in-band signaling mechanisms are bugging and having strictly-defined format for encoding a list of things is good. :+1:

PR #914, opened 2025-03-06
Remove redundant dep declaration.

danielballan:
A dep was listed twice in the same section. You can see the correct (pinned) one on the line immediately above the line deleted here.

~~Trivial fix, not worth a CHANGELOG entry.~~

I added a CHANGELOG entry.


Kezzsim:
Simple and clean

PR #916, opened 2025-03-07
Test commit with CLI commands moved onto Settings object

DiamondJoseph:
Dummy commit to test out possible new pattern of how to create commands.
This form ensures that Settings object has been fully resolved before any commands are run, meaning that the Settings object can be treated as effectively immutable.

This pattern can be recursed down, with e.g.
```
class Settings(BaseSettings):
    command: CliSubCommand[Union[Client, Server]]

    def cli_cmd(self):
        CliApp.run_subcommand(self)

class Client(BaseModel):
    subcommand: CliSubCommand[Union[InMemory, FromUri]]

    def cli_cmd(self):
        CliApp.run_subcommand(self)

class InMemory(BaseModel):

    def cli_cmd(self):
        ...

```

### Checklist
- [ ] Add a Changelog entry


DiamondJoseph:
Deprecated and replaced by #930, a much more complete example.

PR #917, opened 2025-03-07
🔑 Add authenticated property to context

### LINKED ISSUE #818: Add an `authenticated` property
Let's add a property to the `Context` that lets us easily check whether the connection is authenticated (as opposed to anonymous). Basically, expose this expression as a property:

https://github.com/bluesky/tiled/blob/cb31d56fce7fcd93e3bd3caba2445c37c69eefd8/tiled/client/context.py#L177

to enable:

```py
>>> client.context.authenticated  # True or False
```

Kezzsim:
Resolves #818 
### Checklist
- [x] Have the `authenticated` property return `True` or `False` depending on the state of the client
- [x] Have the `authenticated` property show up in the tab context menu of ipython or in the output of  `dir(c.context)`

### Optional
- [x] Write a test for `authenticated` property
- [x] Type assertion

### DIFF
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -666,12 +666,17 @@ class Context:
         """
         Boolean indicated whether session is authenticated (true) or anonymous (false)
 
-        >>> c.authenticated
+        Examples
+        --------
+
+        An anonymous session at first, after login, is authenticated.
+
+        >>> client.context.authenticated
         False
-        >>> c.login()
+        >>> client.login()
         Username: USERNAME
         Password: <input is hidden>
-        >>> c.authenticated
+        >>> client.context.authenticated
         True
 
         """
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -273,10 +273,8 @@ class Context:
         auth_info = []
         if (self.api_key is None) and (self.http_client.auth is None):
             auth_info.append("(unauthenticated)")
-            self.authenticated = False
         else:
             auth_info.append("authenticated")
-            self.authenticated = True
             if self.server_info.authentication.links:
                 whoami = self.whoami()
                 auth_info.append("as")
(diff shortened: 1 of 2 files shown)


danielballan:
If we set this attribute inside `__repr__` we'll have an odd behavior wherein it is only set (and updated) when `__repr__` is called. Take a look at this:

```py

In [2]: client = from_profile('demo')

In [3]: 'authenticated' in dir(client.context)
Out[3]: False

In [4]: client.context  # Displaying the object calls `__repr__()` internally.
Out[4]: <Context (unauthenticated)>

In [5]: 'authenticated' in dir(client.context)
Out[5]: True
```

In #818 I suggested "exposing it as a property", i.e.

```
class Context:

    ...


    @property
    def authenticated(self):
        return ...
```

Thus, `authenticated` will always be defined, and the current value will be computed when the user checks it.

danielballan:
In this test, we check whether the repr shows the Context being authenticated or not. This would be a good place to test the new `authenticated` property has the expected value.

https://github.com/bluesky/tiled/blob/48bb130bea0509dd51809e4be1436628c2a9645c/tiled/_tests/test_authentication.py#L66-L94

And here for the API key case:

https://github.com/bluesky/tiled/blob/48bb130bea0509dd51809e4be1436628c2a9645c/tiled/_tests/test_authentication.py#L342-L358

tiled/client/context.py
```diff
@@ -661,6 +661,11 @@ def configure_auth(self, tokens, remember_me=True):
         auth.sync_set_token("refresh_token", tokens["refresh_token"])
         self.http_client.auth = auth
 
+    @property
+    def authenticated(self):
```
danielballan, line 665: I'm trying to form the habit of including type hints on all new Tiled code, as we gradually type the codebase.

```suggestion
    def authenticated(self) -> bool:
```

Also, would you add a docstring to this property? Just a one-liner is sufficient.

nmaytan (approved):
Tested and looks good!

danielballan (approved):

PR #926, opened 2025-03-13
Further changes to prepare for splitting access token parsing

DiamondJoseph:
This is another breakpoint of relatively readable code changes, prior to making non-backwards compatible changes to config file parsing. 

### Checklist
- [x] Add a Changelog entry
- [ ] Add the ticket number which this PR closes to the comment section

PR #928, opened 2025-03-17
Merge all configuration

DiamondJoseph:
Draft PR to temporarily park this PR while I work on other things.
Currently based on top of #926 with commits starting at 57fe63e

### Checklist
- [ ] Add a Changelog entry


tiled/_tests/test_compression.py
```diff
@@ -18,7 +20,7 @@ def app():
             )
         },
     )
-    return build_app(tree, authentication={"single_user_api_key": "secret"})
+    return build_app(tree, server_settings=Settings(single_user_api_key="secret"))
```
DiamondJoseph, line 23: Pass tree into Settings

tiled/server/settings.py
```diff
@@ -46,13 +108,41 @@ class Settings(BaseSettings):
...
+
+    uvicorn: Uvicorn = Field(Uvicorn())
 
     model_config = SettingsConfigDict(
         env_prefix="TILED_",
         nested_model_default_partial_update=True,
         env_nested_delimiter="_",
     )
 
+    def check_scalable(self):
```
DiamondJoseph, line 125: todo: Make this a validator for the class- currently only distinguishes the case where secret_keys has been explicitly set as an empty list or database uri has been explicitly set to blank, we need to check that those keys have been explicitly set at all.

danielballan:
We anticipate that `access_control` become a server-wide singleton, like `authentication`, instead of a setting applied separately to each `tree`. https://github.com/bluesky/tiled/issues/951

PR #935, opened 2025-03-27
started working on fixing type resolution in sql adapter

skarakuzu:
### Checklist
- [ ] Add a Changelog entry


danielballan:
Superseded by a new approach in #941, where unsigned ints are stored as signed ints in the next-wider size. For example, uint8 is stored as int16, etc. We tested that data round-trips with no overflow issues.

PR #937, opened 2025-03-27
Place lower bound on fastapi version.

danielballan:
The method `check_api_key` was added to `fastapi.security.api_key.APIKeyHeader` in https://github.com/fastapi/fastapi/pull/3142/, released in FastAPI version [0.115.8](https://fastapi.tiangolo.com/release-notes/#01158). We now rely on that method.

It is good to have a lower bound on `fastapi` in any case; we recently saw an issue in the wild where a user at NIST ended up with an old version somehow.

### Checklist
- [x] Add a Changelog entry


danielballan:
For the record, to reproduce the issue that prompted this:

```
```

```py
from_uri("http://localhost:8000", api_key="secret")
```

Server logs:

```
  File "/home/dallan/Repos/bnl/tiled/tiled/server/authentication.py", line 110, in __call__
    return self.check_api_key(None, self.auto_error)
           ^^^^^^^^^^^^^^^^^^
AttributeError: 'StrictAPIKeyHeader' object has no attribute 'check_api_key'
```

DiamondJoseph (approved):

PR #942, opened 2025-04-02
Update Awkward Adapter

genematx:
Minor changes to the AwkwardAdapter and AwkwardClient (including `__repr__`) to better support arrays of records.

### Checklist
- [ ] Add a Changelog entry


tiled/structures/awkward.py
```diff
@@ -56,7 +56,7 @@ def project_form(form, form_keys_touched):
                 form_key=form.form_key,
             )
 
-    elif isinstance(form, (awkward.forms.NumpyForm, awkward.forms.EmptyForm)):
+    elif isinstance(form, awkward.forms.NumpyForm):
```
danielballan, line 59: Needs a test

tiled/client/awkward.py
```diff
@@ -10,6 +10,27 @@ class AwkwardClient(BaseClient):
...
+                if len(out) + len(sample_repr) > 60:  # character count
+                    break
+                out += ", " + sample_repr
+                counter += 1
+            # Are there more fields that what we displayed above?
+            if len(form.fields) > counter:
+                out += f", ...}} ~{len(form.fields)} fields, "
+            else:
+                out += "}, "
+            out += f" {self.structure().length} items>"
```
danielballan, line 31: Probably worth while to special-case here to avoid the ungrammatical `1 items`

```py
In [6]: ac
Out[6]: <AwkwardClient {x, y},  1 items>
```

tiled/client/awkward.py
```diff
@@ -10,6 +10,27 @@ class AwkwardClient(BaseClient):
...
+            # And then show as many more as we can fit on one line.
+            counter = 1
+            for sample_repr in form.fields[1:]:
+                if len(out) + len(sample_repr) > 60:  # character count
+                    break
+                out += ", " + sample_repr
+                counter += 1
+            # Are there more fields that what we displayed above?
+            if len(form.fields) > counter:
+                out += f", ...}} ~{len(form.fields)} fields, "
```
danielballan, line 28: We end up with two spaces here where I think one is intended.

```py
In [12]: ac
Out[12]: <AwkwardClient {x0, x1, x2, x3, x4, x5, x6, x7, x8, x9, x10, ...} ~20 fields,  1 items>
```

tiled/client/awkward.py
```diff
@@ -10,6 +10,27 @@ class AwkwardClient(BaseClient):
...
+            else:
+                out += "}, "
```
danielballan, line 30: Again, two spaces where one is intended. I might suggest dropping the `,` here too. In the case where we print `N fields` it is necessary, but in the case where the fields all fit, it looks superfluous to me.

```py
In [8]: ac
Out[8]: <AwkwardClient {x, y, z},  2 items>
```

danielballan (changes requested):

tiled/client/awkward.py
genematx: good catch! fixed.

danielballan (approved):

danielballan:
Looks good :+1: 

```python
In [6]: ac
Out[6]: <AwkwardClient {x0, x1, x2, x3, x4, x5, x6, x7, x8, x9, x10, ...} ~20 fields, 1 item>

In [8]: ac
Out[8]: <AwkwardClient {x, y, z} 2 items>
```

Issue #943, opened 2025-04-04
issue with latest HDF5Adapter update


toqduj:
I love the fast development on Tiled, it's turning into quite the Swiss army knife.. 

However, there's a bug that appeared in the latest merge: when using the latest version with the refactored HDF5Adapter from @genematx, I get the error quoted below on my file (worked with the previous version). This is also what I mentioned in the PR here: https://github.com/bluesky/tiled/pull/897#issuecomment-2733538182 , but probably got snowed under. 

For testing: I've put the file over here on this WeTransfer share for the next 7 days: https://we.tl/t-9Y5WrPuw4k

I am not sure but I hope it is a small thing that can be resolved easily.. 

```python
[cedf54a875709e61] 127.0.0.1:58938 (public) - "GET /api/v1/metadata/testdata HTTP/1.1" 200 OK
[b8d6102cf0d7be9a] 127.0.0.1:58938 (public) - "GET /api/v1/search/testdata?page%5Boffset%5D=0&fields=&sort= HTTP/1.1" 200 OK
[4e61e96cdbac9f45] 127.0.0.1:58943 (public) - "GET /api/v1/metadata/testdata/MOUSE_20250324_1_160_stacked HTTP/1.1" 200 OK
[2168517947f2a6f0] 127.0.0.1:58943 (public) - "GET /api/v1/search/testdata/MOUSE_20250324_1_160_stacked?page%5Boffset%5D=0&fields=&sort= HTTP/1.1" 200 OK
/Users/bpauw/Code/tiled/tiled/adapters/hdf5.py:176: UserWarning: The dataset /entry1/collection_identifier is of object type, using a Python-only feature of h5py that is not supported by HDF5 in general. Read more about that feature at https://docs.h5py.org/en/stable/special.html. Consider using a fixed-length field instead. Tiled will serve an empty placeholder, unless the object is of size 1, where it will attempt to repackage the data into a numpy array.
  warnings.warn(
/Users/bpauw/Code/tiled/tiled/adapters/hdf5.py:176: UserWarning: The dataset /entry1/experiment/additional_parameters is of object type, using a Python-only feature of h5py that is not supported by HDF5 in general. Read more about that feature at https://docs.h5py.org/en/stable/special.html. Consider using a fixed-length field instead. Tiled will serve an empty placeholder, unless the object is of size 1, where it will attempt to repackage the data into a numpy array.
  warnings.warn(
[...]

PR #944, opened 2025-04-04
FIX: accept scalar data in HDF5

genematx:
Bug fix to treat the cases of scalar data in HDF5 datasets. Previously, passing them to `dask.array.concatenate` would have raised a ValueError.

Added tests for empty and scalar data in HDF5.

Issue: #943 

### Checklist
- [ ] Add a Changelog entry


danielballan (approved):

danielballan:
Since this fixes an unreleased regression, a CHANGELOG entry would just be noise.

PR #945, opened 2025-04-04
Add tiled --version

danielballan:
I probably should have delegated this "good first issue", but it was a good Friday 4:00pm activity. :-D 

### Checklist
- [x] Add a Changelog entry

Closes https://github.com/bluesky/tiled/issues/939

### DIFF
--- a/tiled/commandline/main.py
+++ b/tiled/commandline/main.py
@@ -149,6 +149,30 @@ def tree(
 cli_app.command("register")(register)
 
 
+@cli_app.callback(invoke_without_command=True)
+def callback(
+    ctx: typer.Context,
+    version: bool = typer.Option(
+        False, "--version", "-V", help="Show version and exit."
+    ),
+):
+    if version:
+        from .. import __version__
+
+        typer.echo(f"{__version__}")
+        raise typer.Exit()
+
+    # Only show this if no subcommand was invoked and --version wasn't used
+    if ctx.invoked_subcommand is None and not version:
+        typer.echo(
+            typer.style(
+                "Missing command. Use --help for available commands.",
+                fg=typer.colors.RED,
+            )
+        )
+        raise typer.Exit(1)
+
+
 main = cli_app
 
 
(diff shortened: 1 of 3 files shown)


tiled/commandline/main.py
```diff
@@ -149,6 +149,30 @@ def tree(
...
+    if ctx.invoked_subcommand is None and not version:
+        typer.echo(
+            typer.style(
+                "Missing command. Use --help for available commands.",
+                fg=typer.colors.RED,
+            )
+        )
+        raise typer.Exit(1)
+
+
```
DiamondJoseph, line 175: I'd split these two behaviours up, especially as only testing one path: it looks like [you can automatically print the help docs if no command is passed by configuration](https://typer.tiangolo.com/tutorial/commands/#show-the-help-message-if-no-command-is-given), then this becomes

```suggestion
@cli_app.callback(invoke_without_command=True)
def version_callback(
    ctx: typer.Context,
    version: bool = typer.Option(
        False, "--version", "-V", help="Show version and exit."
    ),
):
    if version:
        from .. import __version__

        typer.echo(f"{__version__}")
```


I've tested this locally and with `cli_app = typer.Typer(no_args_is_help=True)` above:

```sh
$ tiled --version
0.1.0b21.dev46+g4e104dee
$ tiled 
                                                                                                                                                        
 Usage: tiled [OPTIONS] COMMAND [ARGS]...                                                                                                               
                                                                                                                                                        
╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --version             -V        Show version and exit. 
...
```

danielballan:
Oh, good, that's much better. Thanks.

DiamondJoseph (approved):
Happy with the version command as is, this doesn't print help on no command but understand if you think that's too verbose.

padraic-shafer (approved):
Looks good.  This will be a helpful addition.

PR #946, opened 2025-04-07
Versioned specs

genematx:
### Checklist
- [x] Add a Changelog entry


danielballan:
This is no longer needed for `bluesky-tiled-plugins`, so we're going to put this aside. We are open to adding something like this if there is a use case for it, but we won't add it proactively.

PR #949, opened 2025-04-08
Xarray composite

### LINKED ISSUE #948: Add an xarray client for `composite`
It can be based on the existing xarray client, It should target the `parts` and not bother with the "wide table optimization".

danielballan:
The client object for representing `composite` structures was originally placed in `tiled.client.container` because it shares some code with the `container` object via inheritance. Now it is moved into a new module `tiled.client.composite`.

In addition, a `read()` method is implemented. It fetches data from the underlying structures (tables, arrays, etc.) via separate requests. If specific variables are passed to `read()` is selectively reads arrays or columns from tables. An xarray.Dataset is returned.

### Checklist
- [ ] Add a Changelog entry

Closes #948

### DIFF
--- a/tiled/_tests/test_composite.py
+++ b/tiled/_tests/test_composite.py
@@ -6,6 +6,7 @@ import pandas
 import pytest
 import sparse
 import tifffile as tf
+import xarray
 
 from ..catalog import in_memory
 from ..client import Context, from_context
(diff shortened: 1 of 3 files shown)


padraic-shafer (approved):
Seems to be a reasonable start.

PR #953, opened 2025-04-11
Add HTTP retries using stamina

danielballan:
See [Motivation in the `stamina` documentation](https://stamina.hynek.me/en/stable/motivation.html) for background.

Notably:
- Stamina has sensible [defaults](https://stamina.hynek.me/en/stable/motivation.html#stamina) for back-off, jitter, timeout, and max retries. We use them.
- We only retry in situations where there is no response—such as a connection issue or a timeout—or where the response as a `5XX` status code, indicating a server-side issue that may be transient. We do not retry, for example, when we receive a `403` or `404` response.
- Stamina enables globally disabling retries; example code snippets are added to the Tiled docs.
- Retries are disabled in tests.

## Demo

If the client tries to connect but the server has not quite started yet, it retries.

```python

In [2]: c = from_profile('local', api_key='secret')
Scheduled retry in 0.88 seconds due to ConnectError('[Errno 111] Connection refused') (attempt 1)
Scheduled retry in 0.49 seconds due to ConnectError('[Errno 111] Connection refused') (attempt 2)
Scheduled retry in 0.51 seconds due to ConnectError('[Errno 111] Connection refused') (attempt 3)
Scheduled retry in 1.6 seconds due to ConnectError('[Errno 111] Connection refused') (attempt 4)
Scheduled retry in 1.7 seconds due to ConnectError('[Errno 111] Connection refused') (attempt 5)
Scheduled retry in 3.2 seconds due to ConnectError('[Errno 111] Connection refused') (attempt 6)
```

A client error is immediately raised, without retries.

```python
In [3]: c.export("test.csv")  # cannot export container as CSV!
HTTPStatusError                           Traceback (most recent call last)
...
ClientError: 406: None of the media types requested by the client are supported. Supported: application/x-hdf5, application/json. Requested: text/csv. http://localhost:8000/api/v1/container/full/?format=csv
```

I stuck `raise Exception` in one endpoint to test `500` errors:

```python
In [6]: x = c.write_array([1, 2, 3], key="x")

In [7]: x.export('test.csv')
Scheduled retry in 0.62 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'\nFor more information, server admin can search server logs for correlation ID b88de6d017f0efa2.") (attempt 1)
Scheduled retry in 0.38 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'\nFor more information, server admin can search server logs for correlation ID b58e5929c6e2ad79.") (attempt 2)
Scheduled retry in 1.3 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'\nFor more information, server admin can search server logs for correlation ID 474901d8ea7d165d.") (attempt 3)
Scheduled retry in 1.5 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'\nFor more information, server admin can search server logs for correlation ID 297e6ed4535a1c88.") (attempt 4)
Scheduled retry in 2.4 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'\nFor more information, server admin can search server logs for correlation ID eed347baaf0f836a.") (attempt 5)
Scheduled retry in 3.7 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'\nFor more information, server admin can search server logs for correlation ID 5b9506b89262b166.") (attempt 6)
Scheduled retry in 5.0 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'\nFor more information, server admin can search server logs for correlation ID 28e1670338e32af7.") (attempt 7)
Scheduled retry in 5.0 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'\nFor more information, server admin can search server logs for correlation ID 2a3238abb1fc3660.") (attempt 8)
Scheduled retry in 5.0 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'\nFor more information, server admin can search server logs for correlation ID ef5e80c90bcf5705.") (attempt 9)
HTTPStatusError                           Traceback (most recent call last)
...
HTTPStatusError: Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/array/full//x?format=csv'
For more information, server admin can search server logs for correlation ID 9abe1804ee0ec1be.
```

### Checklist
- [x] Add a Changelog entry


tiled/client/utils.py
```diff
@@ -94,6 +95,24 @@ def __init__(self, message, request, response):
         super().__init__(message=message, request=request, response=response)
 
 
+def should_retry(exc: Exception) -> bool:
+    # If the error is an HTTP status error, only retry on 5xx errors.
+    if isinstance(exc, httpx.HTTPStatusError):
+        return exc.response.status_code >= 500
+
+    # Otherwise retry on all httpx errors.
+    return isinstance(exc, httpx.HTTPError)
```
maffettone, line 104: Tired eyes read this as `exec` for execute. 
```suggestion
def should_retry(exception: Exception) -> bool:
    # If the error is an HTTP status error, only retry on 5xx errors.
    if isinstance(exception, httpx.HTTPStatusError):
        return exception.response.status_code >= 500

    # Otherwise retry on all httpx errors.
    return isinstance(exception, httpx.HTTPError)
```

tiled/client/utils.py
```diff
@@ -143,25 +162,37 @@ def export_util(file, format, get, link, params):
...
                 suffix[1:] for suffix in Path(file).suffixes
             )  # e.g. "csv"
-        content = handle_error(
-            get(
-                link,
-                params={**parse_qs(urlparse(link).query), "format": format, **params},
-            )
-        ).read()
+        for attempt in retry_context():
+            with attempt:
```
maffettone, line 201: I have only seen the landing page for stamina. Would it be better to bake retry as a function decorator to `handle_error`? Removes lots of nesting complexity every time you have a request.

maffettone (changes requested):
I think this would be substantially simpler if you decorated the `handle_error` function with 
```
@stamina.retry(on=should_retry)
def handle_error(....
```

These are thematically linked. Is there anywhere in the repository that you are calling handle error without the retry now?

tiled/client/utils.py
danielballan: Great question. I explored some options because, as you'd imagine, I was reluctant to do a bunch of tedious edits throughout `tiled.client`.

It comes down to this observation from the stamina docs:

> Sometimes you only want to retry a part of a function. Since iterators can’t catch exceptions and context managers can’t execute the same block multiple times, we need to combine them to achieve that.

danielballan: I did intentionally wrap `stamina.retry_context` in our own `retry_context` so that we can tweak the settings in one place. (But for now the defaults seem fine.)

danielballan: Decorating `handle_error` would only work in `handle_error` ran the actually HTTP request internally, like this:

```py
handle_error(httpx.get, "http://...")
```

Because it receives the request:

```py
handle_error(httpx.get("http://..."))
```

it's too late to retry.

danielballan: The decorator can only catch and retry exceptions arising from the call stack of the decorated function.

```py
In [1]: import httpx

In [2]: import stamina

In [3]: @stamina.retry(on=httpx.HTTPError)
   ...: def handle_error(response):
   ...:     ...
   ...:     return response.raise_for_error()
   ...:

In [4]: handle_error(httpx.get("http://localhost:9999"))  # ConnectionError
...
ConnectError: [Errno 111] Connection refused
```

maffettone: Ahhh I see. I'm also noticing there are some examples in `array` that don't use the retry directly on the `handle_error` call, so we can't just nest a `handle_error: with retries: handle_http_error`. 
Thanks for also answering some of my questions before I could even hit enter.

maffettone (approved):

danielballan:
Expanding on thoughts from a conversation about with @maffettone. In `main`, Tiled wraps requests like this:

```py
def handle_error(response):
    try:
        return response.raise_for_status()
    except ...:
        # hand-craft a nice exception message with a correlation ID, etc.

handle_error(httpx.get(...))
```

we are retrying two categories of error.

1. Exceptions that are raised by calling `raise_for_status()` on the response, e.g. 500 errors.
2. Exceptions that are raised immediate when the request is made, e.g. connection errors.

The first kind could be handled by decorating `handle_error` like:

```py
@stamina.retry(on=should_retry):
def handle_error(response):
    # as above...
    ...
```

But retrying the second kind requires refactoring every place that a request is made. I considered this usage:

```py
retry = stamina.RetryingCaller().on(should_retry)

handle_error(retry(httpx.get, ...))
```

This works for (2) but not for (1). A HTTP response with an error code passes straight through `retry` because it does not raise an exception until `raise_for_status` is called.

If we combine these two approaches: a decorator on `handle_error` to catch (1) and `retry` function to catch (2), that is a working solution. But I feel it is potentially confusing because the two categories of error will be handled separately and might end up with accidentally-different configuration.

The approach in this PR, with an context manager and a loop, is verbose but explicit and consistent, and I think it is the winner.

danielballan:
@cjboyle, Niko mentioned you might be game to look at this PR. I'd be interested in your thoughts.

danielballan:
Rebased on `main` to resolve conflicts.

cjboyle:
I'm not too familiar with `stamina` specifically (and haven't had a use for it in other projects).

In Tiled, the only time I've hit HTTP connection timeouts was during debugging sessions with server-side breakpoints, in which case the solution is likely to extend/disable the timeout rather than retry... but I think that's an edge case.

I can't think of a case where it would be a factor at the moment (and the defaults are sane), but seeing that there are no global variables for `wait_initial`, total `attempts`, etc., could we add these as environment variables to pass to `retry_context(...)`?

danielballan:
Good idea. I can imagine that we might want to extend the total attempts and timeout if, for example, an overnight scan is running and we would want to give it every chance to recover from some outage.

danielballan:
Good idea. I have opened a discussion on stamina. https://github.com/hynek/stamina/discussions/103

danielballan:
I made `TILED_RETRY_ATTEMPTS` and `TILED_RETRY_TIMEOUT` configurable env vars. I think the others, like `wait_initial`, should not be configurable unless we hear of a use case for it. I'd like to keep things simple, and I'd like to guide clients toward behavior that puts reasonable load on the server.

tiled/client/utils.py
```diff
@@ -104,9 +105,21 @@ def should_retry(exception: Exception) -> bool:
     return isinstance(exception, httpx.HTTPError)
 
 
+# Expose the timeout and max attempts as configurable via env vars. The rest of
+# the parameters (wait_initial, wait_jitter, etc.) are intentionally not
+# included here, for simplicity and to make it more difficult to configure
+# clients to load the server too aggressively.
+TILED_RETRY_ATTEMPTS = int(os.getenv("TILED_RETRY_ATTEMPTS", "10"))
+TILED_RETRY_TIMEOUT = float(os.getenv("TILED_RETRY_ATTEMPTS", "45.0"))
```
cjboyle, line 113: Typo on this line. :slightly_smiling_face:

danielballan: THanks!

danielballan:
I tested the following (with no server running, so a `Connection error`) and saw expected behavior:

```
TILED_RETRY_ATTEMPTS=2 python -c 'from tiled.client import from_uri; from_uri("http://localhost:8000")'
 TILED_RETRY_ATTEMPTS=20 TILED_RETRY_TIMEOUT=10000 python -c 'from tiled.client import from_uri; from_uri("http://localhost:8000")'
TILED_RETRY_TIMEOUT=5 python -c 'from tiled.client import from_uri; from_uri("http://localhost:8000")'
```

cjboyle (approved):

Issue #905, opened 2025-03-04
Support writable blob storage


danielballan:
## Overview

Tiled current supports to forms of writable storage:
- SQL database for appendable tabular datasets
- filesystem for everything else (arrays as Zarr, sparse arrays as parquet files, awkward array as buffers on disk)

This issue is to add a third form of writable storage, with an `s3://...` URI, that would you preferred over `filesystem` if present. To start, it could be used by the Zarr Adapter, which I believe supports blob storage natively (maybe via `fsspec`). That is, each chunk is stored as a blob rather than as a file on disk.

## Why Do We Want This?

- We are already beginning to bump into scaling limitations storing data as files on disk, with too many files in one directory. Blob storage lets us scale "wider".
- We envision detectors moving from file-based storage to blob storage in the medium- and long-term, and Tiled should be ready to support that.

## Dev Environment

This is the probably the right way to get a local blob store to develop against. I see a container in there: https://github.com/minio/minio

Before involving Tiled, test that you can write Zarr arrays in there. These docs pages look relevant:
- https://zarr.readthedocs.io/en/v3.0.0/user-guide/storage.html
- https://zarr.readthedocs.io/en/v3.0.0/api/zarr/storage/index.html#zarr.storage.FsspecStore

## Implementation

Extend this `Storage` class, trivially, add a `blob` key:

https://github.com/bluesky/tiled/blob/14c0220cbabc96863f84948bb348482bf7f05022/tiled/structures/data_source.py#L49-L57

I _think_ no changes are needed to the CLI or the config parsing; any `s3://...` URI provided via either entry point should flow through as is.

I _think_ no changes are needed to the [`tiled.catalog.adapter`](https://github.com/bluesky/tiled/blob/main/tiled/catalog/adapter.py) module because `writable_storage` should just flow through.

This will need to be updated to consult whether `writable_storage` has `blob` value set and use that, or fall back to the current behavior using filesystem storage.

From there, we could do the same for some other adapters that support writing.

Kezzsim:
[...]

PR #955, opened 2025-04-11
🧊 Support for Blob Storage

Kezzsim:
Created to resoslve #905 
**Blob** (_Binary Large Objects_) is a recent cloud native data format with many advancements over traditional file systems which allows it to handle larger volumes of data in containers with less stringent size limitations.

[Zarr has added support for blob storage](https://zarr.readthedocs.io/en/stable/user-guide/storage.html) through [Fsspec](https://filesystem-spec.readthedocs.io/en/latest/), and conversely through [s3fs](https://github.com/fsspec/s3fs), a library which more succinctly implements cloud based connections.

This PR creates a `blob` storage dataclass for tiled, which alongside `filesystem` and `sql` will provide a location where data can be written to and read from. Unlike the presently available storage options, Blobs require `uri`, `key` and `secret` in order to connect to a datasource, omitting the last two parameters if said datasource is public.


tiled/structures/data_source.py
```diff
@@ -44,10 +44,21 @@ def from_json(cls, d):
         return cls(assets=assets, **d)
 
 
+@dataclasses.dataclass
+class Blob:
```
danielballan, line 48: This described a `Bucket`, not a `Blob`, right?

tiled/adapters/zarr.py
```diff
@@ -46,16 +47,27 @@ def init_storage(
...
-        store = zarr.storage.DirectoryStore(str(directory))
+        if storage.bucket:
+            data_uri = storage.bucket.uri
+            s3 = s3fs.S3FileSystem(
+                client_kwargs={"endpoint_url": data_uri},
+                key=storage.bucket.key,
+                secret=storage.bucket.secret,
+                use_ssl=False,
+            )
+            store = s3fs.S3Map(root='/', s3=s3)
```
danielballan, line 63: The `"".join(...)` expression below, on Line 66, should be applied to `root`.

tiled/structures/data_source.py
Kezzsim: Yeah I wasn't sure if we wanted to just swap it for the class name, but `Bucket` makes more sense conscidering what we're defining.
Resolved in https://github.com/bluesky/tiled/pull/955/commits/cd8f1101db57eaec9a8ee2ecb3bc6052643d8012

example_configs/s3_style_storage.yml
```diff
@@ -0,0 +1,13 @@
+authentication:
```
danielballan, line 1: Some of the existing [`example_configs`](https://github.com/bluesky/tiled/tree/main/example_configs) are in sub-directories with other supporting files. In this case, a subdirectory that included both this config and `compose.yml` which launches minio would be handy.

Kezzsim: Resolved in 909ed03, also added a `README.md` with instructions on how to run.
The minio container is now configured to create the bucket automatically.

tiled/adapters/zarr.py
```diff
@@ -49,16 +50,28 @@ def init_storage(
...
-            f"/{quote_plus(segment)}" for segment in path_parts
-        )
         # Zarr requires evenly-sized chunks within each dimension.
         # Use the first chunk along each dimension.
         zarr_chunks = tuple(dim[0] for dim in data_source.structure.chunks)
         shape = tuple(dim[0] * len(dim) for dim in data_source.structure.chunks)
-        directory = path_from_uri(data_uri)
-        directory.mkdir(parents=True, exist_ok=True)
-        store = zarr.storage.DirectoryStore(str(directory))
+        if storage.bucket:
```
danielballan, line 57: In the old model, the adapter was handed multiple storage options and had to pick the one it wanted. Now, the _caller_ in `tiled.catalog.adapter` picks _one_ storage option and passes just that one in.

https://github.com/bluesky/tiled/blob/f3331ef90de056e671ba5acb0a80350c473a696d/tiled/catalog/adapter.py#L672-L695

So, the task here is to check `isinstance(storage, BucketStorage)` versus `FileStorage`. Additionally, the `supported_storage` attribute on this class should be extended to include `BucketStorage`. This tells the caller to offer `BucketStorage` if that is the highest-priority item in `writable_storage`.

tiled/adapters/zarr.py
```diff
@@ -365,9 +378,12 @@ def from_catalog(
         /,
         **kwargs: Optional[Any],
     ) -> Union[ZarrGroupAdapter, ArrayAdapter]:
-        zarr_obj = zarr.open(
-            path_from_uri(data_source.assets[0].data_uri)
-        )  # Group or Array
+        parsed = urlparse(data_source.assets[0].data_uri)
```
danielballan, line 381: Take a look at what `SQLAdapter` does here. If this is bucket storage, we need to get the credentials for this bucket, similarly to how `SQLAdapter` does. The credentials are _not_ in the database, but they are in the config, and we need to match the URL of the bucket store with one in the config.

tiled/utils.py
```diff
@@ -733,7 +735,8 @@ def path_from_uri(uri) -> Path:
         path = Path(parsed.path[1:])
     else:
         raise ValueError(
-            "Supported schemes are 'file', 'sqlite', and 'duckdb'. "
```
danielballan, line 736: No change to this function. Now, unlike when we started this PR, `path_from_uri` is only ever called on local filesystem paths.

tiled/adapters/zarr.py
danielballan: https://github.com/bluesky/tiled/blob/f3331ef90de056e671ba5acb0a80350c473a696d/tiled/adapters/sql.py#L71-L74

tiled/storage.py
```diff
@@ -92,6 +109,8 @@ def parse_storage(item: Union[Path, str]) -> Storage:
         result = FileStorage(item)
     elif scheme == "postgresql":
         result = SQLStorage(item)
+    elif scheme == "bucket":
```
danielballan, line 112: There won't be a bucket scheme. Instead, this function must be extended to accept `dict` (a YAML object from config). While we're here, we might accept a dict for  PostgreSQL as well as for Blob/Bucket.

We currently accept SQL creds like this only:
```yaml
- postgresql://username:password@host:port/database
```

but we could additionally accept a more structured input:

```yaml
- uri: postgresql://host:port/database
  username: username
  password: password
```

And buckets would of course be similar.

Issue #951, opened 2025-04-10
Rework AccessControlPolicy


danielballan:
_Documenting design conversations with @nmaytan and others._

The current `AccessControlPolicy` enables server admins to define access policies with a Python class that filters results based on `metadata`. We have found that relying on `metadata` to access rules gets messy. We propose to add a new column to the `nodes` table for this purpose, alongside but separate from `metadata`.

We want to retain the current flexibility for different facilities / server admins to implement policies tailored to their requirements. In support of this, the content of this column will be an opaque JSON blob, from Tiled's point of view. It is a dedicated space for the AccessControlPolicy to leave a note to itself, which can be then be incorporated efficiently in queries that filter the nodes that a user can access.

## Example

1. An authenticated user creates a dataset. The AccessControlPolicy will be consulted and provide JSON to be stored with this node, based on the identity of the authenticated user: `{"user": "dallan@bnl.gov"}`.
2. When the users access datasets, any nodes marked up their `user` are included in the results that they can access.
4. The user may, at time of creation or later, apply tags to the dataset. The AccessControlPolicy maintains a notion of which users may apply which tags. When a tag is applied, the JSON stored with the node is updated, such as `{"tags": ["beamline_xyz", "proposal_12345"]}`. Now the identity of the creator is no longer relevant; the node is accessible to any users or groups associated with this tag.

An AccessPolicy will be applied to the server as a whole, not per-tree.

danielballan:
In the example above, the AccessControlPolicy resolves each tag to an access control list, including users and groups and associated scopes for each.

If the AccessControlPolicy were configured with a YAML file, it might look like this:

```yaml
# Define roles, lists of scopes
roles:
  - role: facility_user
    scopes: ["read:data", ...]
  - role: beamline_scientist
    scopes: ["read:data", ...]

# For each tag, list users and/or groups
tags:
  - tag: proposal_12345
    users:
      - user: dallan
        role: ...
      - user: nmaytan
        scopes: ["read:metadata"]  # allowed to hard-code one-off scopes?
  - tag: beamline_xyz
    groups:
      - group: instadmin-xyz
        role: ...
- tag: data_admins
  users:
    - user: msnyder
      role: ...

# Who can add and remove a given tag to a node they have access to?
tag_owners:
  proposal_12345:
    groups:
       - foo
```


```py
class AccessControlPolicy:
    def __init___(self, config: dict):
       self.config = config
       self.acls = compile(config)
 
[...]

PR #957, opened 2025-04-16
Expose auth-state information in Session model schema

### LINKED ISSUE #956: Getting loaded UserSessionState after login
Currently the Session model does not expose the `state` column from the database (populated by the LDAP and/or other custom authenticators). Would it make sense to add this? 

A use case for us would be to store authentication info at login time, then fetch the info from the `Principal` object in an access policy to filter bluesky runs. E.g. with beamline information stored from a `run_start` event, determine whether the current user is a member of the beamline staff group from LDAP.

https://github.com/bluesky/tiled/blob/8bd96d60f2e29b20d5dadb3ad4e49125995fc89e/tiled/server/schemas.py#L338-L358

cjboyle:
### Checklist
- [x] Add a Changelog entry

Closes #956

This updates the session model to expose state information stored at login-time, such that it may be used in the access control stack. Added integration tests, more so to demonstrate varying levels of access based on the user attributes.

### DIFF
--- a/tiled/server/schemas.py
+++ b/tiled/server/schemas.py
@@ -350,11 +350,15 @@ class Session(pydantic.BaseModel):
     uuid: uuid.UUID
     expiration_time: datetime
     revoked: bool
+    state: Optional[Dict[Any, Any]]
 
     @classmethod
     def from_orm(cls, orm: tiled.authn_database.orm.Session) -> Session:
         return cls(
-            uuid=orm.uuid, expiration_time=orm.expiration_time, revoked=orm.revoked
+            uuid=orm.uuid,
+            expiration_time=orm.expiration_time,
+            revoked=orm.revoked,
+            state=orm.state,
         )
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 Write the date in place of the "Unreleased" in the case a new version is released. -->
 # Changelog
 
+## Unreleased
+
+### Changed
+
+- Exposed `Session.state` information from database to enhance custom access
+  control developments.
+
 ## 0.1.0-b22 (2025-04-21)
 
 ### Added
--- a/tiled/_tests/test_access_control.py
+++ b/tiled/_tests/test_access_control.py
@@ -548,10 +548,20 @@ def custom_attributes_context():
         yield context
 
 
-def test_custom_attributes_admin_full_access(
-    enter_username_password, custom_attributes_context
+@pytest.mark.parametrize(
+    ("username", "password", "nodes"),
+    [
+        ("admin", "admin", ["A", "B", "C", "D"]),
+        ("alice", "secret1", ["A"]),
+        ("bob", "secret2", ["A", "B"]),
+        ("cara", "secret3", ["A", "C", "D"]),
+    ],
+)
+def test_custom_attributes_with_data_access(
+    enter_username_password, custom_attributes_context, username, password, nodes
 ):
-    with enter_username_password("admin", "admin"):
+    """Test that the user has access to the data based on their auth attributes."""
+    with enter_username_password(username, password):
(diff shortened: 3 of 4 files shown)


danielballan:
Thanks @cjboyle! @dylanmcreynolds Can I rope you in to look over this one?

danielballan:
Also, heads up, @nmaytan, I think this should work fine with #951 but we should keep it in mind and make sure that the design does not evolve in any way that would break this.

tiled/server/schemas.py
```diff
@@ -350,11 +350,15 @@ class Session(pydantic.BaseModel):
...
+    state: Optional[Dict[Any, Any]]
 
     @classmethod
     def from_orm(cls, orm: tiled.authn_database.orm.Session) -> Session:
         return cls(
-            uuid=orm.uuid, expiration_time=orm.expiration_time, revoked=orm.revoked
+            uuid=orm.uuid,
+            expiration_time=orm.expiration_time,
+            revoked=orm.revoked,
+            state=orm.state,
```
dylanmcreynolds, line 361: This requires the additoin of a migrations script, no?

dylanmcreynolds (changes requested):
This PR is exactly something I need. I'm excited. I don't really have any notes other than change to ORM for Session should probably have a migration script

tiled/server/schemas.py
danielballan: No, because the `state` was already in the database:

https://github.com/bluesky/tiled/blob/8faba4c5f2814b286d6a9ec9bdde0599d7692c7f/tiled/authn_database/orm.py#L216

This PR just exposes it.

danielballan: For completeness, the migration that we added when this field was added to the ORM: https://github.com/bluesky/tiled/blob/main/tiled/authn_database/migrations/versions/c7bd2573716d_add_session_state_column.py

danielballan: It's confusing because we have both SQLAlehemy ORMs, which define the table schemas, and pydantic BaseModels, which model the same entity but for the purposes of working with FastAPI.

There are frameworks that do both at once, in one object, but in my opinion they get messy quickly. Both SQLAlchemy and pydantic are complex, and the union is even more complex.

dylanmcreynolds: That explains it!

dylanmcreynolds (approved):

danielballan:
Rebased to fix merge conflicts in `CHANGELOG`

danielballan (approved):
In a follow-up, we should consider if and how this can work with requests that are authenticated using user-generated API keys.

danielballan:
I'm coming back here to give a courtesy heads-up to @cjboyle that we're reshuffling related APIs in https://github.com/bluesky/tiled/pull/963. We're rushing that in so we can have it in time for our upcoming operating cycle (Monday).

The essential change in your PR here—exposing session state to enable it to be used in authorization logic—is retained. That is of interest to us as well. But now, in #963, instead of passing `path_parts` and incidental details about the `node` (a.k.a. `entry`) to the Access Policy, we add a dedicated column to the nodes table for this purpose, `access_blob`. This is JSON markup that can specifically be used to filter results based on access. Based our improved understanding of our own requirements, of what related frameworks do, and on helpful conversations with  @edbarnard at Molecular Foundry and @dylanmcreynolds at ALS, we think this is a step in the right direction.

The change might disrupt your activities in the short term, so this is friendly advice to pin tiled to v0.1.0b26 and an invitation to connect with @nmaytan and me to make sure tiled's reworked take on access control meets your requirements as well as ours. There is room for changes: this is just the first take on a new direction.

cjboyle:
No complaints here! I've been loosely following the developments on the other PR and think it is exactly what we need to meet one of our use cases. I'll likely test things out as soon as it is available.

PR #958, opened 2025-04-18
ENH: Enhance Streaming Exporters

genematx:
Minor changes to streaming serializers/exporters:

1. Added support for async exporters;
2. Replaced `df.iterrows` with `df.to_dict` in `tiled/serialization/table.py:json_sequence` to preserve dtypes of individual columns.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/server/core.py
+++ b/tiled/server/core.py
@@ -1,5 +1,6 @@
 import collections.abc
 import dataclasses
+import inspect
 import itertools
 import math
 import operator
@@ -388,7 +389,7 @@ async def construct_data_response(
         raise UnsupportedMediaTypes(
             f"This type is supported in general but there was an error packing this specific data: {err.args[0]}",
         )
-    if isinstance(content, types.GeneratorType):
+    if isinstance(content, types.GeneratorType) or inspect.isasyncgen(content):
         response_class = StreamingResponse
     else:
         response_class = Response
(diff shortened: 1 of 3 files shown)


tiled/server/core.py
```diff
@@ -388,7 +389,7 @@ async def construct_data_response(
         raise UnsupportedMediaTypes(
             f"This type is supported in general but there was an error packing this specific data: {err.args[0]}",
         )
-    if isinstance(content, types.GeneratorType):
+    if isinstance(content, types.GeneratorType) or inspect.isasyncgen(content):
```
danielballan, line 392: This mixes two paradigms. Does it need to? Could we do an `isinstance` check on https://docs.python.org/3/library/collections.abc.html#collections.abc.AsyncGenerator or, alternatively, check the (non-async) generator using https://docs.python.org/3/library/inspect.html#inspect.isgenerator?

danielballan (approved):
Nice! One minor code-style suggestion, otherwise good to go.

tiled/server/core.py
genematx: yes, sure! I also noticed this; probably should have changed it rightaway.

genematx: fixed now with `inspect` for both cases.

PR #960, opened 2025-04-23
Implement query using SQL `LIKE` operator.

danielballan:
Needs unit test

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -1280,13 +1280,9 @@ def binary_op(query, tree, operation):
 
 
 def like(query, tree):
-    dialect_name = tree.engine.url.get_dialect().name
     keys = query.key.split(".")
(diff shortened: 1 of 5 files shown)


tiled/catalog/adapter.py
```diff
@@ -1408,7 +1420,7 @@ def structure_family(query, tree):
 CatalogNodeAdapter.register_query(StructureFamilyQuery, structure_family)
 CatalogNodeAdapter.register_query(SpecsQuery, specs)
 CatalogNodeAdapter.register_query(FullText, full_text)
-# TODO: Regex
```
danielballan, line 1411: I am dropping this TODO because SQLite does not support regex (out of the box) and `LIKE` covers the most common use case.

Of course it could be revisited, but I don't think this code comment is helping anything now.

PR #963, opened 2025-04-25
New authz

nmaytan:
This is still WIP

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -145,6 +145,7 @@ class RootNode:
         self.ancestors = []
         self.key = None
         self.data_sources = None
+        self.access_blob = {"tags": ["_PUBLIC_NODE"]}
 
 
 class Context:
@@ -321,6 +322,10 @@ class CatalogNodeAdapter:
         self.access_policy = access_policy
         self.startup_tasks = [self.startup]
         self.shutdown_tasks = [self.shutdown]
+
+    @property
+    def access_blob(self):
+        return self.node.access_blob
 
     def metadata(self):
         return self.node.metadata_
@@ -613,16 +618,19 @@ class CatalogNodeAdapter:
         self,
         structure_family,
         metadata,
+        access_blob=None,
         key=None,
         specs=None,
         data_sources=None,
     ):
+        access_blob = access_blob or {}
         key = key or self.context.key_maker()
         data_sources = data_sources or []
 
         node = orm.Node(
             key=key,
             ancestors=self.segments,
+            access_blob=access_blob,
             metadata_=metadata,
             structure_family=structure_family,
             specs=[s.model_dump() for s in specs or []],
@@ -1042,6 +1050,7 @@ class CatalogCompositeAdapter(CatalogContainerAdapter):
         self,
         structure_family,
         metadata,
+        access_blob=None,
         key=None,
         specs=None,
         data_sources=None,
@@ -1081,7 +1090,12 @@ class CatalogCompositeAdapter(CatalogContainerAdapter):
                 specs = [Spec(name="flattened")] + (specs or [])
 
         return await super().create_node(
-            structure_family, metadata, key=key, specs=specs, data_sources=data_sources
+            structure_family,
+            metadata,
+            access_blob=access_blob,
+            key=key,
+            specs=specs,
+            data_sources=data_sources,
         )
 
 
--- a/tiled/server/authentication.py
+++ b/tiled/server/authentication.py
@@ -388,7 +388,12 @@ async def get_current_principal(
                 schemas.Identity(id=identity["id"], provider=identity["idp"])
                 for identity in decoded_access_token["ids"]
             ],
+            roles=[
+                schemas.Role(name=role["name"], scopes=role["scopes"])
+                for role in decoded_access_token["rls"]
+            ],
         )
+        print(principal.roles)
     else:
         # No form of authentication is present.
         principal = SpecialUsers.public
@@ -491,6 +496,9 @@ async def create_tokens_from_session(settings, db, session, provider):
         "sub": principal.uuid.hex,
         "sub_typ": principal.type,  # Why is this str and not Enum?
         "scp": list(set().union(*[role.scopes for role in principal.roles])),
+        "rls": list(
+            {"name": role.name, "scopes": role.scopes} for role in principal.roles
+        ),
         "state": session.state,
         "ids": [
             {"id": identity.id, "idp": identity.provider}
--- a/tiled/server/schemas.py
+++ b/tiled/server/schemas.py
@@ -411,6 +411,7 @@ class PostMetadataRequest(pydantic.BaseModel):
     metadata: Dict = {}
     data_sources: List[DataSource] = []
     specs: Specs = []
+    access_tags: Optional[List[str]] = None
 
     # Wait for fix https://github.com/pydantic/pydantic/issues/3957
     # to do this with `unique_items` parameters to `pydantic.constr`.
--- a/tiled/catalog/orm.py
+++ b/tiled/catalog/orm.py
@@ -68,6 +68,7 @@ class Node(Timestamped, Base):
     key = Column(Unicode(1023), nullable=False)
     ancestors = Column(JSONVariant, nullable=True)
     structure_family = Column(Enum(StructureFamily), nullable=False)
+    access_blob = Column("access_blob", JSONVariant, nullable=False)
     metadata_ = Column("metadata", JSONVariant, nullable=False)
     specs = Column(JSONVariant, nullable=False)
 
@@ -96,6 +97,7 @@ class Node(Timestamped, Base):
             "time_created",
             "id",
             "metadata",
+            "access_blob",
             postgresql_using="gin",
         ),
         # This is used by ORDER BY with the default sorting.
--- a/tiled/client/container.py
+++ b/tiled/client/container.py
@@ -639,6 +639,7 @@ class Container(BaseClient, collections.abc.Mapping, IndexersMixin):
         key=None,
         metadata=None,
         specs=None,
+        access_tags=None,
     ):
         """
         Create a new item within this Node.
@@ -653,6 +654,7 @@ class Container(BaseClient, collections.abc.Mapping, IndexersMixin):
         """
         self._cached_len = None
         metadata = metadata or {}
+        access_tags = access_tags or []
         specs = specs or []
         normalized_specs = []
         for spec in specs:
@@ -666,6 +668,7 @@ class Container(BaseClient, collections.abc.Mapping, IndexersMixin):
                 "structure_family": StructureFamily(structure_family),
                 "specs": normalized_specs,
                 "data_sources": [asdict(data_source) for data_source in data_sources],
+                "access_tags": access_tags,
             }
         }
         body = dict(item["attributes"])
@@ -730,7 +733,14 @@ class Container(BaseClient, collections.abc.Mapping, IndexersMixin):
     # to attempt to avoid bumping into size limits.
     _SUGGESTED_MAX_UPLOAD_SIZE = 100_000_000  # 100 MB
 
-    def create_composite(self, key=None, *, metadata=None, specs=None):
+    def create_composite(
+        self,
+        key=None,
+        *,
+        metadata=None,
+        specs=None,
+        access_tags=None,
+    ):
         """Create a new, empty composite container.
 
         Parameters
@@ -743,6 +753,8 @@ class Container(BaseClient, collections.abc.Mapping, IndexersMixin):
         specs : List[Spec], optional
             List of names that are used to label that the data and/or metadata
             conform to some named standard specification.
+        access_tags: List[str], optional
+            A list of tag names which are used to confer access to the new node
 
         """
         return self.new(
@@ -751,9 +763,17 @@ class Container(BaseClient, collections.abc.Mapping, IndexersMixin):
             key=key,
             metadata=metadata,
             specs=specs,
+            access_tags=access_tags,
         )
 
-    def create_container(self, key=None, *, metadata=None, specs=None):
+    def create_container(
+        self,
+        key=None,
+        *,
+        metadata=None,
+        specs=None,
+        access_tags=None,
+    ):
         """Create a new, empty container.
 
         Parameters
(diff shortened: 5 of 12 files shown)


tiled/client/base.py
```diff
@@ -461,6 +475,10 @@ def update_metadata(self, metadata=None, specs=None):
         specs : List[str], optional
             List of names that are used to label that the data and/or metadata
             conform to some named standard specification.
+        access_blob: dict, optional
+            Server-specific authZ info in blob form, used to confer access to the node.
+            May be nested. Must contain only basic types
```
danielballan, line 480: This, copied from `metadata` above, may not fully apply. (I don't think this content can be nested.) For now, maybe the first sentence is sufficient.

There are many instances of this, in this file and in `container.py`. I will note only this one.

tiled/client/base.py
```diff
@@ -496,19 +514,27 @@ def update_metadata(self, metadata=None, specs=None):
         >>> md['unwanted_key'] = DELETE_KEY
         >>> node.update_metadata(metadata=md)  # Update the copy on the server
         """
-        if isinstance(metadata, list) and len(metadata) == 2:
```
danielballan, line 499: I don't recall this; it might be slipped past previous review. I guess this is meant to support:

```py
stuff = x.metadata_copy()
# Alter stuff...and then...
x.update_metadata(stuff)
```

But if that is desired the user should do `x.update_metadata(*stuff)`. I think this whole conditional branch should go.

If this breaks tests or creates hassle, feel free to punt it to a separate GH Issue.

tiled/server/authentication.py
```diff
@@ -491,6 +495,9 @@ async def create_tokens_from_session(settings, db, session, provider):
         "sub": principal.uuid.hex,
         "sub_typ": principal.type,  # Why is this str and not Enum?
         "scp": list(set().union(*[role.scopes for role in principal.roles])),
+        "rls": list(
+            {"name": role.name, "scopes": role.scopes} for role in principal.roles
```
danielballan, line 499: I am skeptical of encoding the scopes for each role in the `Authorization` header of every request. We have the union of the scopes above, in `scp`.

What makes this necessary? Can we find another way?

authz_workspace/access_blob_queries.py
```diff
@@ -0,0 +1,105 @@
...
+
+from tiled.catalog import orm
+from tiled.catalog.adapter import CatalogNodeAdapter
+from tiled.queries import JSONSerializable
+from tiled.query_registration import register
+
+
+@register(name="access_blob_filter")
+@dataclass
+class AccessBlobFilter:
```
danielballan, line 17: I think two things follows from our decision to specify a schema for `access_blob`, instead of leaving it as arbitrary JSON for access policies to define.

1. This should hard-code the keys `user` and `tags`, instead of making them configurable. It simplifies the interface and removes unhelpful gernality.
2. This query should _probably_ be promoted into core, in `tiled.catalog.adapter` so that other access policy implementation can import it.

tiled/client/base.py
nmaytan: I borrowed the comment/wording as I was adhering to the "`access_blob` is unspecified" idea, much like metadata. As you brought up in other comments, I know we discussed being OK with imposing some structure, but I still followed the principle of keeping it free-form for as long as it was possible (which it turned out was pretty much entirely)

I am happy to go and comb these out, if you still think we want to back away for that idea. I am of two minds about it still- the flexibility is perhaps a nice-to-have, but I also find the idea of a powerful reference implementation very enticing.

tiled/client/base.py
nmaytan: Right, that is indeed the idea as I understand it. I will go back and look at it, I tried to stick to existing convention throughout. Agreed that unpacking seems like the right way to ask a user to do this.

tiled/server/authentication.py
nmaytan: The key problem is that a user having **N** scopes (even if **N** is `len(ALL_SCOPES)`), and a user being designated the admin role, are not necessarily the same thing. One can easily imagine, in my view, a principal which is granted all scopes and yet is not formally an admin.

This admin role check is used to bypass tag-based access restrictions, of course. if we are comfortable saying "_only_ an admin ever has _this_ scope" then perhaps this can be avoided, but I don't like that as much as explicitly checking the role.

As an alternative, I could boil this down to a boolean (is/is not admin) that gets encoded - I thought perhaps it better to have the role information accessible in general but the admin role is the key thing we want for authz purposes.

nmaytan: Oh, I also realize I misread your question a little bit (sorry for that)- the `scopes` themselves could go. It's the roles that are important. I kept the scopes so that the form would be consistent (roles structure in the key == roles structure in the orm). We don't explicitly need the scopes for what we're doing.

authz_workspace/access_blob_queries.py
nmaytan: Following up on the other comment thread, I had stuck to the "unspecified `access_blob`" approach. I am not strongly opposed to enforcing a schema, though. If you'll confirm, I'll make this more rigid.

I originally had put this in `tiled.catalog.adapter` and `tiled.queries`, as above I'm open to moving it back inside.

tiled/server/authentication.py
danielballan: Thanks for refreshing my memory of our motivation behind this.

> if we are comfortable saying "only an admin ever has this scope" then perhaps this can be avoided

I think the key point is, a user can bypass authorization checks if they already have the power to impersonate other users. That power is conferred specifically by `admin:apikeys` scope. I agree with the sense of your statement above—“smells like an admin” feels wrong—but I think basing the bypass on this capability feels right. I could imagine in the future adding a dedicated scope (`authz:root`) that lets us separate this from impersonation, which might be slightly better.

I have in mind an implicit future requirement: support for API gateways in place of our internal authentication system. When we integrate with a gateway, it will provide Tiled with JWTs and (via configuration) will have to learn about Tiled scopes. But there is no reason it should need to adopt Tiled’s internal authentication system’s notion of Roles.

authz_workspace/access_blob_queries.py
danielballan: Will think on this for a short while.

tiled/client/base.py
nmaytan: This is removed now, should not impact any tests. Works perfectly fine without it .

tiled/server/authentication.py
nmaytan: I started looking at this, but quickly realized that the `Principal` schema doesn't have a `scopes` field in which to pack this intersection of scopes.

nmaytan: This has been addressed by threading scopes retrieved from authentication through get_entry and the access_policy. Our new access policy now checks for `admin:apikey` to confer admin access.

authz_workspace/access_blob_queries.py
nmaytan: These keys are now `user_id` and `tags` and are hardcoded into the filter implementaiton, which has been moved back into Tiled (catalog)

tiled/client/base.py
nmaytan: This has been focused in on `access_tags` in the client API.

authz_workspace/tag_based_pass_access_policy.py
```diff
@@ -0,0 +1,492 @@
...
+    cursor.execute(
+        """
+CREATE TABLE IF NOT EXISTS store (
+    obj_name TEXT PRIMARY KEY,
+    obj_blob BLOB
+)
+"""
+    )
+    loaded_tags_blob = sqlite3.Binary(
+        pickle.dumps(loaded_tags, protocol=pickle.HIGHEST_PROTOCOL)
```
danielballan, line 184: I'll just comment for posterity that the approach of storing the tags as a pickle encoded blob inside a SQLite file is a temporary step. It leverages SQLite's file-locking to achieve atomic syncs to disk. In the future, we expect to externalize the tag state from process memory to a database. When we take that step, we'll encode the tags in the database idiomatically.

authz_workspace/tag_based_pass_access_policy.py
```diff
@@ -0,0 +1,492 @@
...
+    )
+    cursor.execute(
+        "REPLACE INTO store (obj_name, obj_blob) VALUES (?, ?)",
+        ("tags_state", loaded_tags_blob),
+    )
+    conn.commit()
+
+
+class TagBasedPASSAccessPolicy(TagBasedAccessPolicy):
+    def __init__(self, url, tags_db, sync_proposals={}, *args, **kwargs):
```
danielballan, line 194: Avoid mutable defaults. (This can be `None`, and the in the method body, `sync_proposals = sync_proposals or {}`.)

tiled/access_policies.py
```diff
@@ -105,3 +127,502 @@ async def filters(self, node, principal, scopes, path_parts):
...
+
+def interning_constructor(loader, node):
+    value = loader.construct_scalar(node)
+    return intern(value)
+
+
+InterningLoader.add_constructor("tag:yaml.org,2002:str", interning_constructor)
+
+
+class LoadedTags(NamedTuple):
```
danielballan, line 155: I haven't seen this pattern before; I think it might be crossing two different idioms.

```py
In [8]: class Thing(collections.namedtuple("Thing", ["a", "b", "c"])):
   ...:     ...
   ...:
```

```py
In [11]: @dataclasses.dataclass
    ...: class Thing:
    ...:     a: dict
    ...:     b: dict
    ...:     c: dict
    ...:
```

Dataclasses are newer to the language, and typed, and probably would be best here.

tiled/access_policies.py
```diff
@@ -105,3 +127,502 @@ async def filters(self, node, principal, scopes, path_parts):
...
+        scopes=None,
+    ):
+        self.provider = provider
+        self.tag_config_path = Path(tag_config)
+        self.scopes = scopes if (scopes is not None) else ALL_SCOPES
+        self.read_scopes = [intern("read:metadata"), intern("read:data")]
+        self.reverse_lookup_scopes = [intern("read:metadata"), intern("read:data")]
+        self.unremovable_scopes = [intern("read:metadata"), intern("write:metadata")]
+        self.admin_scopes = [intern("admin:apikeys")]
+        self.public_tag = intern("public").casefold()
```
danielballan, line 178: Should this intern the result of casefolding?
```suggestion
        self.public_tag = intern("public".casefold())
```

tiled/catalog/adapter.py
```diff
@@ -1375,6 +1388,37 @@ def specs(query, tree):
     return tree.new_variation(conditions=tree.conditions + conditions)
 
 
+def access_blob_filter(query, tree):
+    dialect_name = tree.engine.url.get_dialect().name
+    access_blob = orm.Node.access_blob
+    if len(query.user_id) == 0 and len(query.tags) == 0:
```
danielballan, line 1394: This caught me by surprise because I would expect `"user"` to be `None` (or in JSON parlance `null`) rather than `""`. This would catch either case:

```suggestion
    if (not query.user_id) and len(query.tags) == 0:
```

or even

```suggestion
    if not (query.user_id or query.tags)
```

as both empty strings and empty lists are false-y. YMMV on how that reads.

tiled/queries.py
```diff
@@ -479,6 +479,48 @@ def SpecQuery(spec):
...
+    >>> c.search(AccessBlobFilter("bill", ["tag_for_bill", "useful_data"]))
+    """
+
+    user_id: str
+    tags: List[str]
+
+    def encode(self):
+        return {
+            "user_id": self.user_id,
+            "tags": self.tags,
```
danielballan, line 513: We typically JSON encode/decode compound values like this in the other queries. At this precise moment I cannot recall what exact problem this solves—maybe just paranoia. But following existing patterns may be best.

tiled/queries.py
```diff
@@ -479,6 +479,48 @@ def SpecQuery(spec):
...
+
+    Examples
+    --------
+
+    Search for user "bill", as well as tags in ["tag_for_bill", "useful_data"]
+
+    >>> c.search(AccessBlobFilter("bill", ["tag_for_bill", "useful_data"]))
+    """
+
+    user_id: str
```
danielballan, line 507: Consider `Optional[str]` where `None` means "not set". A `__post_init__` could be added to validate that if `tags` is nonempty the `user_id` must be `None`:

```py
def __post_init__(self):
    if (self.user_id is not None) and self.tags:
        raise ValueError(...)
```

tiled/server/schemas.py
```diff
@@ -547,6 +560,8 @@ class PatchMetadataResponse(pydantic.BaseModel, Generic[ResourceLinksT]):
     # May be None if not altered
     metadata: Optional[Dict]
     data_sources: Optional[List[DataSource]]
+    # May be None if not altered
```
danielballan, line 563: Nit: The comment on Line 560 applies to all three; this is superfluous.

danielballan (commented):
Great stuff. I did a line review.

tiled/access_policies.py
nmaytan: Good point, this is technically true. I put the `casefold` there only to show intention for how this key is used, but yes this makes sense to do.

tiled/access_policies.py
nmaytan: I didn't have a strong reason for this, and did consider a `dataclass`, NamedTuple just seemed somewhat more lightweight for this use case. No problem swapping.

tiled/queries.py
nmaytan: Both `user_id` and `tags` are always used by the query (necessarily) - we want to know: what are you the "creator" of, and what do tags let you see? Neither should ever be `None` - thought the tags list could be empty for a user with no tags.

tiled/queries.py
nmaytan: Fair enough - I didn't see a reason to do this for `List[str]`. Most cases where we encode are `JSONSerializable` (~arbitrary) - but not all, and some of these type are still encoded. I'll do the same.

tiled/catalog/adapter.py
nmaytan: Agreed, this was me just picking up the syntax from the other filters - I think I misinterpreted that a `len` check had been done on a string elsewhere.

tiled/queries.py
danielballan: I see now that I conflated the query with the content of the access blob. This is correct.

danielballan (approved):

PR #965, opened 2025-04-30
Refactor Storage.

danielballan:
- Provide a formal protocol for Adapters to declare what writable storage types they support.
- Let writable storage be specified in priority order, where an adapter users the first supported storage listed.
- Handle credentials properly for SQL-backed storage.

This does not include support for blob (bucket) storage, but it should make implementing it straightforward.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -159,7 +159,6 @@ class Context:
 
         self.writable_storage = []
         self.readable_storage = set()
-        self.readable_filesystem_storage = []
 
         # Back-compat: `writable_storage` used to be a dict: we want its values.
         if isinstance(writable_storage, dict):
@@ -180,10 +179,6 @@ class Context:
         # credentials (if applicable).
         for item in self.readable_storage:
             register_storage(item)
-        # Stash a copy of filesystem-based readable storage.
-        self.readable_filesystem_storage = set(
-            item for item in self.readable_storage if isinstance(item, FileStorage)
-        )
 
         self.key_maker = key_maker
         adapters_by_mimetype = adapters_by_mimetype or {}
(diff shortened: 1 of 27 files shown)


danielballan:
Before, credentials in the postgresql URI would be persisted in the database. It's against security best practice to leak credentials into a database, and of course it will cause problems if (when) the credential is rotated.

From the `assets` table in the catalog database:

```
3   postgresql://postgres:secret@localhost:5432       0                                            2025-04-30 15:55:28  2025-04-30 15:55:28
```

In this PR, credentials not not included. The (credential-free) URI is matched with the credentials provided in server config to read or write data later.

```
4   postgresql://localhost:5432                       0                                            2025-04-30 16:02:54  2025-04-30 16:02:54
```

tiled/catalog/adapter.py
```diff
@@ -158,23 +157,32 @@ def __init__(
     ):
         self.engine = engine
 
+        self.writable_storage = []
+        self.readable_storage = set()
+        self.readable_filesystem_storage = []
```
genematx, line 162:
```suggestion
        self.readable_filesystem_storage = set()
```

tiled/catalog/adapter.py
```diff
@@ -483,11 +491,10 @@ async def get_adapter(self):
                 # Protect against misbehaving clients reading from unintended parts of the filesystem.
                 asset_path = path_from_uri(asset.data_uri)
                 for readable_storage in self.context.readable_filesystem_storage:
```
genematx, line 493: I'm still wondering if we need the extra attribute ("readable_filesystem_storage")... Is there an intention to use it externally somewhere? Maybe we can just use the condition from line 185 in this loop to make the code a bit shorter?

genematx (approved):

tiled/storage.py
```diff
@@ -0,0 +1,115 @@
...
+]
+
+
+@dataclasses.dataclass(frozen=True)
+class Storage:
+    "Base class for representing storage location"
+    uri: str
+
+    def __post_init__(self):
+        ensure_uri(self.uri)
```
genematx, line 25:
```suggestion
        self.uri = ensure_uri(self.uri)
```

PR #967, opened 2025-04-30
add launch.json

gwbischof:
Adding a launch.json this allows you to launch tiled in a devcontainer with debugging, like breakpoints.

This adds a play-button to launch tiled in the dev-container.
Enables breakpoints, and debugging tools like step-over, etc.
It is also possible to launch multiple terminals in the dev-container so that you can curl (for example) your tiled from a separate terminal.

image


.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
...
+    "configurations": [
+      {
+        "name": "Debug Tiled Server",
+        "type": "python",
+        "request": "launch",
+        "module": "tiled",
+        "args": [
+          "serve",
+          "config",
+          "--host", "0.0.0.0",
```
danielballan, line 12: It's generally not best practice for a development container with a placeholder API key (`"secret"`) to accept external connections. Is the default (`127.0.0.1`) not sufficient?

danielballan (commented):
Neat!

.vscode/launch.json
gwbischof: Oh yep good catch, 127.0.0.1 should work.

gwbischof: Ohh 127.0.0.1 works :D This launch command would (in many cases) be running inside of the dev-container, so I didn't think 127.0.0.1 would allow the mapping of the container port to the host port (8000:8000)

gwbischof, line 12:
```suggestion
          "--host", "127.0.0.1",
```

gwbischof: Updated

gwbischof: If we don't want to map the container port to the host port we could change the docker-compose to use `expose` instead of `ports`

gwbischof: I think this should be good. 
Should I remove this change to the compose?
```
    ports:
      - "127.0.0.1:3000:3000"
 ```

File: docker-compose.yml [hunk 2]
```diff
@@ -4,8 +4,10 @@ services:
     image: ghcr.io/bluesky/tiled:0.1.0-b24
     environment:
       - TILED_SINGLE_USER_API_KEY=${TILED_SINGLE_USER_API_KEY}
+    
+    # Note that the port mapping does not expose tiled externally.
     ports:
-      - 8000:8000
+      - "127.0.0.1:8000:8000"
```
gwbischof, line 10:
```suggestion
      - "8000:8000"
```

docker-compose.yml
```diff
@@ -4,8 +4,10 @@ services:
     image: ghcr.io/bluesky/tiled:0.1.0-b24
     environment:
       - TILED_SINGLE_USER_API_KEY=${TILED_SINGLE_USER_API_KEY}
+    
+    # Note that the port mapping does not expose tiled externally.
     ports:
-      - 8000:8000
+      - "8000:8000"
```
gwbischof, line 10:
```suggestion
      - 8000:8000
```

docker-compose.yml
```diff
@@ -4,6 +4,8 @@ services:
     image: ghcr.io/bluesky/tiled:0.1.0-b24
     environment:
       - TILED_SINGLE_USER_API_KEY=${TILED_SINGLE_USER_API_KEY}
+    
+    # Note that the port mapping does not expose tiled externally.
```
gwbischof, line 8:
```suggestion
```

docker-compose.yml
```diff
@@ -34,8 +35,10 @@ services:
     image: docker.io/grafana/grafana:8.2.6
     depends_on:
       - prometheus
+
+    # Note that the port mapping does not expose grafana externally.
     ports:
-      - 3000:3000
+      - "127.0.0.1:3000:3000"
```
gwbischof, line 41:
```suggestion
      - 3000:3000
```

docker-compose.yml
```diff
@@ -34,6 +35,8 @@ services:
     image: docker.io/grafana/grafana:8.2.6
     depends_on:
       - prometheus
+
+    # Note that the port mapping does not expose grafana externally.
```
gwbischof, line 39:
```suggestion
```

docker-compose.yml
```diff
@@ -4,6 +4,7 @@ services:
     image: ghcr.io/bluesky/tiled:0.1.0-b24
     environment:
       - TILED_SINGLE_USER_API_KEY=${TILED_SINGLE_USER_API_KEY}
+
```
gwbischof, line 7:
```suggestion
```

danielballan (approved):

PR #969, opened 2025-05-05
Update data_source from the server response

genematx:
This is to make sure that the `data_source` in the clients is updated by the received response from the server (similar to `metadata` and `structure`).

### Checklist
- [x] Add a Changelog entry


danielballan (approved):
Needs CHANGELOG to node removal of parameter in HTTP API

PR #970, opened 2025-05-06
Mount node

danielballan:
See CHANGELOG for context.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/_tests/test_mount_node.py
+++ b/tiled/_tests/test_mount_node.py
@@ -34,7 +34,6 @@ def test_mount_node(tmpdir):
                 "tree": "catalog",
                 "args": {
                     "uri": catalog_uri,
-                    "init_if_not_exists": True,
                     "writable_storage": [tmpdir / "data"],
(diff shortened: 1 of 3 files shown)


tiled/_tests/test_mount_node.py
```diff
@@ -0,0 +1,98 @@
...
+    # Now it is as if we have two separate catalogs, but they happen
+    # to be housed in one database.
+    multi_tree_config = {
+        "trees": [
+            {
+                "path": "/a",
+                "tree": "catalog",
+                "args": {
+                    "uri": catalog_uri,
+                    "writable_storage": [tmpdir / "data"],
```
genematx, line 37: This is probably unrelated to the changes in this PR specifically, but what would happen if, for example, array with the same name would be created in each of the catalogs (an presumably written to the same storage)? I wonder if it would be better to encourage the admin to create separate storages per catalog? (or even enforce it?)

genematx (approved):

Issue #973, opened 2025-05-07
SQL adapter cannot write columns with '-'


canismarko:
If I try to use the a device in a plan that contains anything besides [A-Za-z_] I get the following exception in the Tiled server upon writing. This is common since ophyd-async uses "-" as a delimiter for device names.

```python
adbc_driver_manager.ProgrammingError: INVALID_ARGUMENT: Failed to prepare query: ERROR:  syntax error at or near "-"
    | LINE 7:         sim_motor_2-user_setpoint DOUBLE PRECISION NULL,
    |                            ^
    | 
    | Query was:
    |     CREATE TABLE IF NOT EXISTS table_1cced0874bd3f64bfdfdd11f75aaba56 (
    |         _dataset_id INTEGER NULL,
    |         _partition_id SMALLINT NULL,
    |         seq_num BIGINT NULL,
    |         time DOUBLE PRECISION NULL,
    |         sim_motor_2-user_setpoint DOUBLE PRECISION NULL,
    |         ts_sim_motor_2-user_setpoint DOUBLE PRECISION NULL)
    |     . SQLSTATE: 42601
    +------------------------------------
```

This is not a problem if I use a device without a '-' in the name.

Here is the full stack trace: https://gist.github.com/canismarko/2c66a7ddb9a43a176c9d7d6605ac2eae#file-sqlerror-py

# Steps to Reproduce

This MWE triggers the exception because `sim_motor_2.user_setpoint` has the name "sim_motor_2-user_setpoint" and this becomes the name of the data key.

```python
from bluesky import RunEngine, plans as bp
from bluesky.callbacks.tiled_writer import TiledWriter
from ophyd_async.core import init_devices
from ophyd_async.epics.motor import Motor
from tiled.client import from_uri


with init_devices():
    sim_motor_2 = Motor("25idc:simMotor:m2", name="sim_motor_2")

RE = RunEngine({})

api_key = "*****"
client = from_uri("http://<host>:<port>", api_key=api_key)
tw = TiledWriter(client['testing'])
RE.subscribe(tw)

plan = bp.count([sim_motor_2.user_setpoint], md=dict(purpose="Test postgres SQL database."))

RE(plan)
```

# Possible Solution

If I change [this line](https://github.com/bluesky/tiled/blob/c807c871aad802570c1c03abb4960bd9dbc5ce04/tiled/adapters/sql.py#L819C1-L819C83) so that it quotes `{name}`, everything works fine:

```python
[...]

PR #974, opened 2025-05-07
Allow for special characters in sql column names

genematx:
This allows (some) special characters to be used in column names created by the `SQLAdapter`. The modified implementation ensures that the SQL identifiers in compiled parameterized requests with enclosed in double quotation marks. Examples of allowed special characters: `- _ ? :` and spaces. Furthermore, some of the special characters are explicitly black-listed: `" ' ` ; =` etc.

The implemented rules are less strict than the typical industry standards (which would normally allow only only letters, numbers, and underscore) to ensure the compatibility with existing `data_key` identifiers, while also taking precautions against potential sql ingestions.

Issue: #973 

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/_tests/adapters/test_sql.py
+++ b/tiled/_tests/adapters/test_sql.py
@@ -6,7 +6,7 @@ import adbc_driver_sqlite
 import pyarrow as pa
 import pytest
 
-from tiled.adapters.sql import SQLAdapter, check_table_name
+from tiled.adapters.sql import TABLE_NAME_PATTERN, SQLAdapter, is_safe_identifier
 from tiled.storage import parse_storage, register_storage
 from tiled.structures.core import StructureFamily
 from tiled.structures.data_source import DataSource, Management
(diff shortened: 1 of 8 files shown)


tiled/_tests/adapters/test_sql.py
```diff
@@ -429,6 +429,18 @@ def test_check_table_name_long_name(
             "_here_is_my_table",
             pytest.raises(ValueError, match="Illegal table name!"),
         ),
+        (
```
genematx, line 432: not sure if these should be valid or not.

tiled/adapters/sql.py
```diff
@@ -305,8 +305,13 @@ def read(self, fields: Optional[List[str]] = None) -> pandas.DataFrame:
         Returns the concatenated pyarrow table as pandas dataframe.
         """
 
+        cols = (
```
genematx, line 308: I've added the list of columns into SELECT

tiled/_tests/adapters/test_sql.py
danielballan: I like that we are strict about this. https://github.com/bluesky/tiled/blob/382ce4f9ca6b201af168731b17aac6b321bf1e4c/tiled/adapters/sql.py#L827

tiled/adapters/sql.py
danielballan: Cool. I think in that case we _don't_ want to select `_partition_id` or `_dataset_id`. We only use them for filtering on, and we drop them below.

Likewise, the code below that drops extra columns is superfluous now and can be removed.

danielballan: It does not:

```
ValueError: Target schema's field names are not matching the table's field names: ['a'], ['a', 'b']
```

genematx: Fixed, thanks!

tiled/adapters/sql.py
```diff
@@ -342,7 +347,7 @@ def read_partition(
         Returns the concatenated pyarrow table as pandas dataframe.
         """
         query = (
-            f"SELECT * FROM {self.table_name} "
+            f'SELECT * FROM "{self.table_name}" '
```
danielballan, line 350: Should the performance optimization applied in `read` also be applied here in `read_partition`?

tiled/adapters/sql.py
```diff
@@ -316,14 +317,18 @@ def read(self, fields: Optional[List[str]] = None) -> pandas.DataFrame:
 
         # The database may have stored this in a coarser type, such as
         # storing uint8 data as int16. Cast it to the original type.
-        data = data.drop_columns(["_partition_id", "_dataset_id"])
-        schema = self.structure().arrow_schema_decoded
-        data.cast(schema)
+        # Pick only the present columns from the schema.
+        to_drop = set(data.column_names).intersection(["_partition_id", "_dataset_id"])
+        data = data.drop_columns(to_drop)
```
danielballan, line 322: I don't follow the logic of this change. Wouldn't the set `to_drop` always be `["_partition_id", "_dataset_id"]`?

tiled/adapters/sql.py
```diff
@@ -305,8 +305,9 @@ def read(self, fields: Optional[List[str]] = None) -> pandas.DataFrame:
         Returns the concatenated pyarrow table as pandas dataframe.
         """
 
+        cols = ", ".join([f'"{c}"' for c in fields]) if fields else "*"
         query = (
-            f"SELECT * FROM {self.table_name} "
+            f'SELECT {cols} FROM "{self.table_name}" '
```
danielballan, line 310: :+1: :+1: :+1:

Do we need to take any additional precautions against SQL injection here? I've grown comfortable relying on frameworks for this and putting user-defined variables directly into SQL strings is scary.

Maybe just check that each `c` in `fields` cannot have any quotes in it? Better if there were some standard function in sqlalchemy or elsewhere that we could use.

danielballan (changes requested):
Great stuff! I have some questions, but I like the direction.

tiled/adapters/sql.py
genematx: This is needed in case we are fetching all columns; if only some fields are selected, trying to drop non-existent columns raises an error.
I've refactored this part, so it's no longer needed.

tiled/adapters/sql.py
genematx: oh yes, it definitely should! fixed.

PR #978, opened 2025-05-20
Moving tag compilation into separate service

nmaytan:
_Edit by @danielballan_: We're splitting up this God object into:

- `TagBasedAccessPolicy` which interacts with the Tiled server, implementing the API it expects
- `AccessTagsParser` which wraps a read-only client to a SQLite database with "compiled" information about users and tags
- Outside of Tiled, facility-specific code that produces that compiled representation. Some of this (the parts that aren't facility specific) can likely be upstreamed in a future PR.
---

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/access_policies.py
+++ b/tiled/access_policies.py
@@ -121,6 +121,59 @@ class SimpleAccessPolicy:
         return queries
 
 
+class TagsParser:
+    def __init__(self, uri):
+        self.db = sqlite3.connect(f"{uri}?ro", uri=True, check_same_thread=False)
+
+    def is_tag_defined(self, name):
+        cursor = self.db.cursor()
+        cursor.execute("SELECT 1 FROM tags WHERE name = ?;", (name,))
+        row = cursor.fetchone()
+        found_tagname = bool(row)
+        return found_tagname
+
+    def get_public_tags(self):
+        cursor = self.db.cursor()
+        cursor.execute("SELECT name FROM public_tags;")
+        public_tags = {name for (name,) in cursor.fetchall()}
+        return public_tags
+
+    def get_scopes_from_tag(self, tagname, username):
+        cursor = self.db.cursor()
+        cursor.execute(
+            "SELECT scope_name FROM user_tag_scopes WHERE tag_name = ? AND user_name = ?;",
+            (tagname, username),
+        )
+        user_tag_scopes = {scope for (scope,) in cursor.fetchall()}
+        return user_tag_scopes
+
+    def is_tag_owner(self, tagname, username):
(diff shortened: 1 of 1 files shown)


tiled/access_policies.py
```diff
@@ -121,6 +121,59 @@ async def filters(self, node, principal, authn_scopes, scopes):
         return queries
 
 
+class TagsParser:
+    def __init__(self, uri):
```
danielballan, line 125: There are two reasons to take the live `db` object here instead of the `uri`.

1. A general preference toward "`__init__` should take _exaclty_ what it needs"
2. Concretely, a memory-backed SQLite database—handy in tests because it's fast and isolated—can _only_ be passed around this way, if we want to test `TagParser` with some database-population code, etc.

We often use this pattern, so that we can construct things from YAML config.

```py
class TagParser
    @classmethod
    def from_uri(cls, uri):
        db = ...
        return cls(db)
```

tiled/access_policies.py
```diff
@@ -129,268 +121,89 @@ async def filters(self, node, principal, authn_scopes, scopes):
...
 
 
 class TagBasedAccessPolicy:
     def __init__(
         self,
         *,
         provider,
-        tag_config,
-        group_parser,
+        tags_db,
```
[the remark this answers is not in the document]
danielballan: I answered my own questions here: it's expected to be a dict with the key `uri`, leaving room to later add other keys, perhaps similar to what we've done with the AuthN `database` config.

https://github.com/NSLS2/tiled-site-config/pull/289/files#diff-f6b9451502751960a2399feced35f3a19395cd661d1e89d1bc3f111afe035451R5-R6

That's fair enough. We're slightly inconsistent on this. The catalogs just take a `uri`; the AuthN database takes a dict. I'm not sure which is right.

tiled/access_policies.py
```diff
@@ -129,268 +121,89 @@ async def filters(self, node, principal, authn_scopes, scopes):
...
-InterningLoader.add_constructor("tag:yaml.org,2002:str", interning_constructor)
-
-
-@dataclass(frozen=True)
-class LoadedTags:
-    tags: dict = field(default_factory=dict)
-    public: set = field(default_factory=set)
-    scopes: dict = field(default_factory=dict)
-    owners: dict = field(default_factory=dict)
+class TagsParser:
```
danielballan, line 124:
```suggestion
class AccessTagsParser:
```

danielballan:
The unit test coverage is unsurprisingly poor. We will address this after the maintenance period.

PR #980, opened 2025-05-22
FIX: apply user access_tag to tables when unspecified

genematx:
Fixing an issue where a table can not be created if an access pass is not set.

### Checklist
- [x] Add a Changelog entry


nmaytan (approved):
Thanks for catching this- I think I overlooked it after allowing an empty tags list

danielballan (approved):
Needs a CHANGELOG entry, then good to go

PR #982, opened 2025-05-30
Control `page_size` more carefully

### LINKED ISSUE #645: Add a page_size method to container.values(), keys(), items()
## Summary

We should add a page_size keyword to the `values()`, `keys()`, and `items()` methods of Tiled containers to allow for advanced control of the amount of data downloaded at one time.

## Background

While testing python code to access Tiled repositories, it was noticed that 
`catalog.values()[index]`
is much slower than 
`catalog[uid]`
despite being the same basic sort of operation. Indeed, the Tiled documentation https://blueskyproject.io/tiled/reference/python-client.html even claims to support "efficient random access".

Digging into the Tiled requests, it seems that all calls to `catalog.values()` request the default page_size of 100. This means that 100 values will be fetched, even if we know we are only going to access one.

## Proposal

I propose a two-fold change to the data access. First, `catalog._keys_slice` and `catalog._items_slice` (which actually fetch the data) would be updated so that they request a page_size of `max(start - stop, DEFAULT_PAGE_SIZE)`. This would instantly make all calls to `values()`, `items()`, and `keys()` more efficient.

Second, for advanced use cases, it would be easy to add a `page_size` keyword to the `values()`, `items()`, and `keys()` functions that could override the default page size. This would be especially useful when populating a table in a GUI with search results from a catalog, where the correct page_size should really be equal to the number of rows that the GUI requests to update.

Comments welcome.

danielballan:
This resolves the issue where `c.values().last()`—and likewise the back-compat wrapper `db[-1]`—is slow at CHX. Since the fix is entirely client-side, it is easy to test this against NSLS-II production (`tiled.nsls2.bnl.gov`).

```py
In [4]: %time chx.values().last()
CPU times: user 10.2 ms, sys: 1.92 ms, total: 12.1 ms
Wall time: 16.6 s
Out[4]: <BlueskyRun v2.0 {'primary'} scan_id=167143 uid='ded4a92e' 2025-05-13 14:05>
```

On this PR branch:

```py
In [5]: %time chx.values().last()
CPU times: user 11.5 ms, sys: 1.06 ms, total: 12.5 ms
Wall time: 851 ms
Out[5]: <BlueskyRun v2.0 {'primary'} scan_id=167143 uid='ded4a92e' 2025-05-13 14:05>
```

I noticed that HEX was also looking slow. This is `main` again:

```py
In [8]: %time hex.values().last()
CPU times: user 5.6 ms, sys: 956 μs, total: 6.56 ms
Wall time: 5.58 s
Out[8]: <BlueskyRun v3.0 streams: {'tomo', 'baseline'} scan_id=85 uid='d343682b' 2025-05-27 15:22>
```

and the PR branch:

```py
In [4]: %time hex.values().last()
CPU times: user 290 ms, sys: 75.5 ms, total: 365 ms
Wall time: 936 ms
Out[4]: <BlueskyRun v3.0 streams: {'baseline', 'tomo'} scan_id=85 uid='d343682b' 2025-05-27 15:22>
```

Closes #645 

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,16 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ## Unreleased
 
+### Added
+
+- It is now possible to explicitly control the page size used for fetching
+  batches of metadata, e.g. `client.values().page_size(N)`.
+
 ### Fixed
 
 - An auth bug that prevented a user to create a table with empty access_tags.
+- When accessing a small number of results, the page size is set appropriately
+  to avoid needlessly downloading additional results.
 
 ## v0.1.0-b28 (2025-05-21)
 
--- a/docs/source/reference/python-client.md
+++ b/docs/source/reference/python-client.md
@@ -63,6 +63,7 @@ batched across requests, which affects performance.
    :toctree: generated
 
    tiled.iterviews.ValuesView.page_size
+```
 
 Likewise for `.keys()` and `.items()`.
 
--- a/docs/source/reference/python-client.md
+++ b/docs/source/reference/python-client.md
@@ -55,6 +55,15 @@ and several convenience methods:
    tiled.iterviews.ValuesView.tail
 ```
 
+as well as a method for controlling the "page size" in which results will be
+batched across requests, which affects performance.
+
+```{eval-rst}
+.. autosummary::
+   :toctree: generated
+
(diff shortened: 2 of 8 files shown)


Kezzsim (approved):
Performs as expected when used as a client on my laptop, approving for the time being

PR #989, opened 2025-06-05
💫 Upgrade Zarr to Version 3

Kezzsim:
### Checklist
- [x] Pass all tests touching `arrays`
- [x] Semantically update method names to newer versions

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -248,7 +248,7 @@ server = [
     "tifffile",
     "uvicorn[standard]",
     "xarray",
-    "zarr <3",
+    "zarr >=3",
     "zstandard",
 ]
 # These are needed by the client and server to transmit/receive sparse arrays.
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -248,7 +248,7 @@ server = [
     "tifffile",
     "uvicorn[standard]",
     "xarray",
-    "zarr <3",
+    "zarr >=3",
```
danielballan, line 251: There are many dependency group: all must be updated.

danielballan:
Capturing discussion:

- Install Zarr 3 conditional of Python version >= 3.11
- Never install Zarr 2; we are Zarr 3 only.
- In `tiled.adapters.zarr`, add a utility function `check_python_version` that raises `NotImplementedError` is Python < 3.11. Call this at the top of all constructors (e.g. `__init__`, `init_storage`).
- In relevant tests, use the `needs_zarr3` decorator to skip tests that would fail.

As a result, Tiled servers and clients will still operate on Python 3.9 and 3.10, but specifically tiled server _that use the Zarr adapter_ will need a version of Python that supports Zarr 3.

danielballan:
The reduction in coverage is due to the branching based on Python version: a given Python version only touches one branch or the other.

PR #991, opened 2025-06-10
Support optional params in mimetypes with serializers

### LINKED ISSUE #919: Pass mimetype to serializers, and support mimetype parameters in registry
Tiled "serializers" are functions which take a Python data structure as input and produce bytes as output, for transport over the network. The serializers currently look like this:

https://github.com/bluesky/tiled/blob/48bb130bea0509dd51809e4be1436628c2a9645c/tiled/serialization/array.py#L38-L43

These get registered to a mimetype, potentially multiple of them.

https://github.com/bluesky/tiled/blob/48bb130bea0509dd51809e4be1436628c2a9645c/tiled/serialization/array.py#L46-L48

I propose changing the signature to add ~`mimetype`~ `media_type` as the first argument.

```py
def f(media_type, array, metadata):
    ...
```

This is a common pattern in callback registries. (We adopted the same pattern in caproto.) It enables the same callback to be subscribed to multiple channels and easily distinguish which one of them is calling. I see two use cases:

1. Serializers that can handle many types but need to do different things (like PIL). We currently handle this with a convenience function.
2. Support for mimetype parameters, such as `text/csv;header=present` and `text/csv;header=absent`. It would be convenient for `serializer_csv` to be called in both cases, and to be able to adjust its output correctly.

Also, in support of (2), I think this the registry should be tweaked to support optional mimetype parameters, such as `text/csv;header=present`. When the registry is passed a mimetype with parameters, it should perform dispatch based on the type without arguments (`text/csv`) but it should pass the full mimetype with arguments to the serializer, which can then parse the mimetype and do what it likes with the arguments.

jmaruland:
### Checklist
- [x] Add a Changelog entry

This PR closes #919

With this PR, we are adding the option of passing additional/optional parameters for mimetypes that can be used to modify the export of data

### DIFF
--- a/tiled/serialization/table.py
+++ b/tiled/serialization/table.py
@@ -12,7 +12,7 @@ from ..utils import APACHE_ARROW_FILE_MIME_TYPE, XLSX_MIME_TYPE, modules_availab
 @default_serialization_registry.register(
     StructureFamily.table, APACHE_ARROW_FILE_MIME_TYPE
 )
-def serialize_arrow(df, metadata, preserve_index=True):
+def serialize_arrow(mimetype, df, metadata, preserve_index=True):
     import pyarrow
 
     if isinstance(df, pyarrow.Table):
@@ -30,7 +30,7 @@ def serialize_arrow(df, metadata, preserve_index=True):
 @default_deserialization_registry.register(
     StructureFamily.table, APACHE_ARROW_FILE_MIME_TYPE
 )
-def deserialize_arrow(buffer):
+def deserialize_arrow(mimetype, buffer):
     import pyarrow
 
     return pyarrow.ipc.open_file(buffer).read_pandas()
@@ -39,7 +39,7 @@ def deserialize_arrow(buffer):
 # There seems to be no official Parquet MIME type.
 # https://issues.apache.org/jira/browse/PARQUET-1889
 @default_serialization_registry.register(StructureFamily.table, "application/x-parquet")
-def serialize_parquet(df, metadata, preserve_index=True):
+def serialize_parquet(mimetype, df, metadata, preserve_index=True):
     import pyarrow.parquet
 
     table = pyarrow.Table.from_pandas(df, preserve_index=preserve_index)
@@ -49,14 +49,14 @@ def serialize_parquet(df, metadata, preserve_index=True):
     return memoryview(sink.getvalue())
 
 
-def serialize_csv(df, metadata, preserve_index=False):
+def serialize_csv(mimetype, df, metadata, preserve_index=False):
     file = io.StringIO()
     df.to_csv(file, index=preserve_index)
     return file.getvalue().encode()
 
 
 @default_deserialization_registry.register(StructureFamily.table, "text/csv")
-def deserialize_csv(buffer):
+def deserialize_csv(mimetype, buffer):
     import pandas
 
     return pandas.read_csv(io.BytesIO(buffer), header=None)
@@ -77,7 +77,7 @@ default_serialization_registry.register(
 
 
 @default_serialization_registry.register(StructureFamily.table, "text/html")
-def serialize_html(df, metadata, preserve_index=False):
+def serialize_html(mimetype, df, metadata, preserve_index=False):
     file = io.StringIO()
     df.to_html(file, index=preserve_index)
     return file.getvalue().encode()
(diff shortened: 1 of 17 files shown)


tiled/serialization/table.py
```diff
@@ -49,14 +49,19 @@ def serialize_parquet(df, metadata, preserve_index=True):
     return memoryview(sink.getvalue())
 
 
-def serialize_csv(df, metadata, preserve_index=False):
+def serialize_csv(mimetype, df, metadata, preserve_index=False):
     file = io.StringIO()
+    if ";" in mimetype:
```
danielballan, line 54: As agreed, _this_ PR will only add one example of handling mimetype parameters. But we expect future PRs to handle more examples of this, so it may be worth generalizing to a separate function that we can reuse. That will also enable us to test it separately for correctness.

````py
def parse_mimetype(mimetype: str): -> tuple[str, dict]
    "Given a mimetype like 'x/y;a=b;c=d' return ('x/y', {'a': 'b', 'c': 'd'}."
    ...
```

tiled/server/core.py
```diff
@@ -375,10 +380,10 @@ async def construct_data_response(
     try:
         if filter_for_access is not None:
             content = await ensure_awaitable(
-                serializer, payload, metadata, filter_for_access
+                serializer, format, payload, metadata, filter_for_access
```
danielballan, line 383: The fact that this passes tests reveals a gap in our test suite. Reading the code above, this needs to be `media_type`, the chosen match between the (list of) types provided by the user via a `format` query parameter or an `Accept` header and the set of types supported by the server.

tiled/server/core.py
```diff
@@ -317,18 +317,23 @@ async def construct_data_response(
     if format is not None:
         media_types_or_aliases = format.split(",")
         # Resolve aliases, like "csv" -> "text/csv".
-        media_types = [
-            serialization_registry.resolve_alias(t) for t in media_types_or_aliases
-        ]
+        media_types = []
+        for t in media_types_or_aliases:
+            # Optional parameters must be ignored to resolve the alias first
```
danielballan, line 322: Reading this more fully, this is stripping off the parameter in the wrong place. We want to remove it _just_ for the comparison with set of formats supported by the server, but not permanently, because we do want to pass the parameter through to the serializer. (This _happens_ to work in the tested case, but not in general. See my comment below about how instead of passing `format` to `ensure_awaitable(serializer, ...)` it should be `media_type`.)

The right place to obtain the bare `media_type` is somewhere within the loop

```py
for media_type in media_types:
    ...
```

This is where we compare the `media_type` requested by the client to the set supported by the server.

```py
if media_type in media_types_for_spec:
```

For that to work, we want the `media_type` on the left to be just the base, without parameters, something like:

```py
if base_media_type in media_types_for_spec
    ...
```

Note that the `parse_mimetype` utility suggested in a previous comment could be used to obtain `base_media_type`.

danielballan (changes requested):

tiled/server/core.py
jmaruland: With the implementation of `parse_mimetype`, we can separate `base_media_type` from the optional parameters now

tiled/serialization/table.py
```diff
@@ -49,14 +54,19 @@ def serialize_parquet(df, metadata, preserve_index=True):
...
 
-def serialize_csv(df, metadata, preserve_index=False):
+def serialize_csv(mimetype, df, metadata, preserve_index=False):
     file = io.StringIO()
+    opt_params = parse_mimetype(mimetype)[1]
+    if "header" in opt_params:
+        if opt_params["header"] == "absent":
+            df.to_csv(file, index=preserve_index, header=False)
+            return file.getvalue().encode()
     df.to_csv(file, index=preserve_index)
```
danielballan, line 64:
```suggestion
include_header = opt_params.get("header", "present") == "absent"
df.to_csv(file, header=include_header, index=preserve_index)
```

danielballan (approved):

danielballan:
Just a quick interactive test, which works as expected:

```py
In [4]: c.write_dataframe({'a': [1,2,3]}, key='x')
Out[4]: <DataFrameClient ['a']>

In [6]: c['x'].export('test1.csv', format='text/csv')

In [7]: !cat test1.csv
a
1
2
3

In [8]: c['x'].export('test1.csv', format='text/csv;header=absent')

In [9]: !cat test1.csv
1
2
3
```

Issue #345, opened 2022-10-18
Front end dependencies can move into devDependencies


dylanmcreynolds:
There are a number of dependencies in package.json that could be moved into a devDependencies section for smaller builds. I don't know the whole list but what jumps out is:

```
    "@testing-library/jest-dom": "^5.16.2",
    "@testing-library/react": "^12.1.2",
    "@testing-library/user-event": "^13.5.0",

    "@types/jest": "^27.4.0",
    "@types/js-yaml": "^4.0.5",
    "@types/material-ui": "^0.21.12",
    "@types/node": "^16.11.25",

    "js-yaml": "^4.1.0",
```

dylanmcreynolds:
Do we really need `@types/node`?

PR #992, opened 2025-06-17
updated dependencies

AditiChikkali:
This PR updates frontend dependencies and improves code compatibility and tests

Changes made
- Updated frontend dependencies to address security vulnerabilities and improved deprecated patterns
- Revised tsconfig.json and fixed TypeScript errors for compatibility with updated packages and modern JavaScript features
- Refactored MUI DataGrid usage to align with the latest API
- Removed unnecessary and default test files
- Moved frontend dependencies into devDependencies as suggested in [#345](url)

### DIFF
--- a/web-frontend/package-lock.json
+++ b/web-frontend/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "tiled",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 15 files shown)


danielballan:
Wonderful! This is great to see—thank you @AditiChikkali.

One thing jumps out at a glance: `node_modules` was committed, results in 147,305 lines added and 484 files changes. We'll need to "rewrite history" in these commits before merging so that they don't enter the git history of the `main` branch.

I think I know how this came about. I had been using `web-frontend` as my working directory when doing development on the React front-end, and so `node_modules` have gone into `web-frontend/node_modules`. The `.gitignore` ensures that they are not committed:

https://github.com/bluesky/tiled/blob/ef5a5f1c3f3428b40fc68443d1ede4cd2b9e47b0/.gitignore#L14

I think you worked from the repository root, and thus ended up with `node_modules` there, and accidentally committed it.

---

In the back of my mind, I was thinking that eventually web-frontend _might_ be split off into its own repository someday, and to that end it will be helpful to keep it self-contained in a subdirectory. There are pros and cons to splitting it off, and I'd like to keep it where it is for now, while laying track for the possibility of splitting it off later.

PR #994, opened 2025-06-24
updated frontend dependencies

AditiChikkali:
This PR updates frontend dependencies and improves code compatibility and tests

Changes made

- Updated frontend dependencies to address security vulnerabilities and improved deprecated patterns
- Revised tsconfig.json and fixed TypeScript errors for compatibility with updated packages and modern JavaScript features
- Refactored MUI DataGrid usage to align with the latest API
- Removed unnecessary and default test files
- Moved frontend dependencies into devDependencies as suggested in [#345](https://github.com/bluesky/tiled/pull/url)
- Formatted project files using Prettier for consistent code style.

### DIFF
--- a/web-frontend/.eslintrc.json
+++ b/web-frontend/.eslintrc.json
@@ -1,20 +1,15 @@
 {
-    "root": true,
-    "extends": [
-        "eslint:recommended",
-        "plugin:@typescript-eslint/recommended"
-    ],
-    "parser": "@typescript-eslint/parser",
-    "plugins": [
-        "@typescript-eslint"
-    ],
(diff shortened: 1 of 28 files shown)


danielballan:
It's notable that this also removes `package-lock.json` from source control (which is where that `-23,394` lines diff comes from). I'm on board with dropping it, especially with unit tests on the way.

dylanmcreynolds:
> It's notable that this also removes `package-lock.json` from source control (which is where that `-23,394` lines diff comes from).

Looks like `package-lock.json` is still in source control, no? This is a good thing.

Issue #446, opened 2023-06-13
Support deleting items from a catalog


danielballan:
Deletion in supported in the _client_ (and used by the externally-developed writable adapters) but not yet supported by the writable adapter in core.

Careful attention must be paid to SQLAlchemy "cascades": deleting a node should also clean up its data sources, assets, revisions.

PR #997, opened 2025-07-01
Fix methods for node deletion

genematx:
This pull request improves delete functionality in the client API. The most notable updates include the adoption of the Closure Table concept for node hierarchy, and replacing `delete_tree` with `delete_contents` (in `Container` and `Composite` nodes); all nodes has a (self-)`delete` method inherited from the `BaseClient`. Furthermore, added support for deletion of internally-managed SQL assests.

### Enhancements to Testing:
* **New Tests for Deletion Scenarios**: Added tests for deleting external nodes, SQL assets, and assets registered multiple times. These tests ensure correctness and robustness of the new `delete_contents` method. 
* **Refactored Composite Tests**: Refactored and expanded tests in `test_composite.py` to include fixtures for external data sources and new scenarios for deletion.

These changes are related to #446 and #835.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/_tests/test_writing.py
+++ b/tiled/_tests/test_writing.py
@@ -24,8 +24,9 @@ from starlette.status import (
 )
 
 from ..catalog import in_memory
-from ..catalog.adapter import CatalogContainerAdapter, WouldDeleteData
+from ..catalog.adapter import CatalogContainerAdapter
 from ..client import Context, from_context, record_history
+from ..client.utils import ClientError
 from ..mimetypes import PARQUET_MIMETYPE
 from ..queries import Key
 from ..server.app import build_app
@@ -528,7 +529,7 @@ async def test_delete(tree):
                 key="delete_me",
             )
 
-        with pytest.raises(WouldDeleteData):
+        with pytest.raises(ClientError):
             client.delete_contents("delete_me")  # Cannot easily delete internal data
         client.delete_contents("delete_me", external_only=False)
 


tiled/client/container.py
```diff
@@ -382,11 +382,44 @@ def __getitem__(self, keys, _ignore_inlined_contents=False):
...
+        """
+
         self._cached_len = None
+        if keys is None:
+            keys = self.keys()
+        keys = [keys] if isinstance(keys, str) else keys
         for attempt in retry_context():
             with attempt:
-                handle_error(self.context.http_client.delete(f"{self.uri}/{key}"))
+                for key in set(keys):
```
danielballan, line 411: Should this loop go _outside_ the retry loop?

tiled/client/container.py
```diff
@@ -382,11 +382,44 @@ def __getitem__(self, keys, _ignore_inlined_contents=False):
...
+        Parameters
+        ----------
+        keys : str or list of str
+            The key(s) to delete. If a list, all keys in the list will be deleted.
+            If None (default), delete all contents.
+        recursive : bool, optional
+            If True, descend into sub-containers and delete their contents too.
+            Defaults to False.
+        external_only : bool, optional
+            If True, only delete externally-managed data keys. Defaults to True.
```
danielballan, line 402:
```suggestion
            If True, only delete externally-managed data. Defaults to True.
```

Avoiding "data keys" as an overloaded term (bluesky). And the sentence works fine with just "data".

tiled/client/composite.py
```diff
@@ -97,6 +98,38 @@ def create_composite(self, key=None, *, metadata=None, specs=None):
...
+        Only constituent arrays or entire tables (with keys listed in `parts`)
+        can be deleted.
+
+        Parameters
+        ----------
+        keys : str or list of str
+            The key(s) to delete. If a list, all keys in the list will be deleted.
+            If None (default), delete all contents.
+        external_only : bool, optional
+            If True, only delete externally-managed data keys. Defaults to True.
```
danielballan, line 117:
```suggestion
            If True, only delete externally-managed data. Defaults to True.
```

Avoiding "data keys" as an overloaded term (bluesky). And the sentence works fine with just "data".

tiled/client/base.py
```diff
@@ -866,11 +866,27 @@ def metadata_revisions(self):
...
+    def delete(self, recursive: bool = False, external_only: bool = True) -> None:
+        """Delete the node and it's contents, if any.
+
+        Parameters
+        ----------
+        recursive : bool, optional
+            If True, descend into sub-nodes and delete their contents too.
+            Defaults to False.
+        external_only : bool, optional
+            If True, only delete externally-managed data keys. Defaults to True.
```
danielballan, line 878:
```suggestion
            If True, only delete externally-managed data. Defaults to True.
```

Avoiding "data keys" as an overloaded term (bluesky). And the sentence works fine with just "data".

tiled/catalog/adapter.py
```diff
@@ -1153,16 +1163,35 @@ async def append_partition(self, *args, **kwargs):
...
-def delete_asset(data_uri, is_directory):
+def delete_asset(data_uri, is_directory, parameters=None):
     url = urlparse(data_uri)
     if url.scheme == "file":
         path = path_from_uri(data_uri)
         if is_directory:
             shutil.rmtree(path)
         else:
             Path(path).unlink()
+    elif url.scheme in {"duckdb", "sqlite"}:
```
danielballan, line 1174: What about postgresql?

genematx: Thank you! The test didn't catch it. Added `postgresql` here and also expanded the test in `tiled/_tests/test_catalog.py::test_delete_sql_assets`.

danielballan:
I see one usability issue we should fix while we are here. Given some data:

```py
c = from_uri("http://localhost:8000", api_key="secret")
x = c.create_container("x")
x.write_array([1,2,3])
```

if I try to delete it without `external_only=False`:

```py
c.delete_contents(recursive=True)
```

the server logs a helpful error message

```
tiled.catalog.adapter.WouldDeleteData: Some items in this tree are internally managed. Deleting the records will also delete the underlying data files. If you want to delete them, pass external_only=False.
```

but the client just gets a `500 Internal Server Error'` and enters a retry loop. Let's catch this exception, similarly to what we do below, and return `409` (seems best) or maybe `400`, propagating that error message.

https://github.com/bluesky/tiled/blob/5c5f0c028c8806d39880b4303fd58fe2df4bf801/tiled/server/app.py#L316-L331

PR #998, opened 2025-07-02
Ensure Correct SQL Schema when Adding Data

genematx:
This PR adds an explicit data type (schema) conversion in SQL adapter when appending new rows to an existing table. This is necessary to correctly parse the type of entries where default data type determined by ADBC in the appended rows may differ from and can not be implicitly cast to the data type in the target table. For example, writing a column with empty lists, e.g. `[[], [], []]`:

```
column_name: list<item: null>
  child 0, item: null
```

to a table column defined as lists of strings

```
column_name: list<item: string>
  child 0, item: string
```

results in a conversion error in PostgreSQL adapter.

Note: to be merged after #1040.

### Checklist
- [x] Add a Changelog entry


danielballan (commented):
Good fix, but this should get a unit test (or an update to an existing one) before we merge.

PR #999, opened 2025-07-03
Websocket endpoint

vshekar:
### Checklist
- [ ] Add a Changelog entry


danielballan:
Agreed. We [added a column to the nodes table](https://github.com/bluesky/tiled/pull/999/files#diff-1c920f04742c4a5fd50ba32d99bc9d2749646c424ce462690b85db6b442feef8R74) which I think is the right out-of-band encoded to address these questions.

That will also facilitate (in a future PR...) search queries like, "List the nodes that are streaming."

danielballan:
TO DO:

- [ ] Tests
- [ ] At least a _start_ at docs
- [ ] Rename interfaces to avoid hard-coding `redis` in the public API (paving the way for NATS etc.)

danielballan:
A heartbeat for @kivel: We have removed "redis" from public APIs. The next step toward generalization would be some polymorphic adapter ("adapter" in the general sense, not a "Tiled Adapter") that will enable a common interface to Redis-backed and NATS-backed (etc.) implementations. That might happen in a later PR, but the track is laid.

danielballan:
Skipping these tests on Windows, for now, would be acceptable IMO.

tiled/_tests/test_websockets.py
```diff
@@ -33,6 +35,7 @@ def tiled_websocket_context(tmpdir):
         yield context
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="Requires Redis service")
```
danielballan, line 38: This can be done at the module level, I believe. Less work for Future Us when we add more tests or if we add Windows support.

gwbischof: Sounds good, updated

tiled/client/stream.py
```diff
@@ -0,0 +1,186 @@
...
+from starlette.testclient import TestClient
+from websockets.sync.client import connect
+
+from tiled.client.context import Context
+
+Callback = Callable[["Subscription", dict], None]
+"A Callback will be called with the Subscription calling it and a dict with the update."
+
+
+class WebsocketWrapper:
```
danielballan, line 18: If this is implemented as _two_ wrapper classes, with identical public APIs, the check on `is_test_client` can be performed once (when you decide which wrapper class to instantiate, instead of internally on every call to `recv`.

tiled/server/app.py
```diff
@@ -234,7 +235,7 @@ async def lifespan(app: FastAPI):
         yield
         await shutdown_event()
 
-    app = FastAPI(lifespan=lifespan, dependencies=[Depends(move_api_key)])
+    app = FastAPI(lifespan=lifespan)  # , dependencies=[Depends(move_api_key)])
```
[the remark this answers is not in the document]
danielballan: I pushed a fix for this. Tested it interactively with my web browser. Works as before.

image

tiled/config_schemas/service_configuration.yml
```diff
@@ -73,6 +73,18 @@ properties:
               args:
                 uri: "catalog.db"
             ```
+  redis:
```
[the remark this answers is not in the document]
gwbischof: I fixed up this file. But it looks like this file is just for validating config files not something that gets loaded into settings?

PR #1004, opened 2025-07-08
Publish container image when merged to main

dylanmcreynolds:
### Checklist
- [ ] Add a Changelog entry

It is often handy to test containers from main. I really think Tiled used to do this.

### DIFF
--- a/.github/workflows/publish-image.yml
+++ b/.github/workflows/publish-image.yml
@@ -47,7 +47,7 @@ jobs:
 
       - name: Build and push Docker image
         uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-image.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
       - name: Build and push Docker image
         uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671
-        if: github.ref_type == 'tag'
+        if: github.ref_type == 'tag' || github.ref == 'refs/heads/main'
```
DiamondJoseph, line 50: You probably also want to modify the `meta` action above to give the container some sensible tags and labels?

PR #1005, opened 2025-07-09
Add type hints to server utils

abbiemery:
Add types to most of the functions in server utils for improved readability.

### Checklist
- [ ] Add a Changelog entry


danielballan:
Agreed. Thanks!

PR #1008, opened 2025-07-10
Move shareable authz utilities into Tiled

### LINKED ISSUE #951: Rework AccessControlPolicy

nmaytan:
* Code pertaining to access control has been moved into an `access_control` subdirectory
* The access tags compiler and db schema are made available in Tiled
* Unit test coverage has been added for the new access policy and access control features
* API keys can now be limited to access for only specific tags
* Some improvements/bug fixes

This PR should mean we can close #951 

### Checklist
- [x] Add a Changelog entry


docs/source/explanations/access-control.md
```diff
@@ -35,95 +35,43 @@ Rephrasing these two items now using the jargon of entities in Tiled:
...
+Roles are pre-defined lists of scopes, and are also defined in this file. This mapping
+confers these scopes to these users for data which is tagged with the corresponding tagname.
 
-```
-ALICE_PASSWORD=secret1 BOB_PASSWORD=secret2 CARA_PASSWORD=secret3 tiled serve config example_configs/config.yml
-```
+Tags can also inherit the ACLs of other tags, using the `auto_tags` field. There is also a
+`public` tag which is a special tag used to mark data as public (all users can read).
+
+Lastly, only "owners" of a tag can apply that tag to a node. Tag owners are defined in
```
genematx, line 67: I'm thinking, is it possible to declare tag owners in the same section where we declare the main tag definitions? Something like:
```
tags:
  alice_tag:
    users:
      - name: "alice"
        role: "facility_admin"
        owner: true
      - name: "chris"
        scopes: ["read:data", "read:metadata"]
        owner: false   # default, doesn't need to be set
  chris_tag:
    users:
      - name: "alice"    # here alice is not the owner...
        role: "facility_admin"
      - name: "chris"
        role: "facility_admin"
         owner: true     # ...but chris is
  biologists_tag:
    users:
      - name: "alice"
        role: "facility_admin"
    groups:
      - name: "biologists"
        scopes: ["read:data", "read:metadata"]
         owner: true      # anyone from biologists can apply the tag
```
i think this would simplify configs and make them more readable. I keep finding myself to having scroll back and forth in the file to check who's the owner of which tag.
Are there any conceptual limitations to this design (e.g. an owner may not be a user of a tag)?

nmaytan: Good question - I agree there is a small cost in ergonomics for having them defined in separate tables. This consideration has come up in design discussions, and the reason we decided to separate these is that it's plausible that these tables may each be subject to different permissions. i.e. different roles in the organization may have rights to modify who can apply a tag but not the permissions that tag confers, and vice versa. 

This is something we can discuss further, but probably deserves a separate conversation/PR.

The structure of the table just reflects that tags are the focus item, but this direction (vs the inverse) also helps with deduplication & makes it easier to understand in totality who owns the tag.

example_configs/toy_authentication.yml
```diff
@@ -7,27 +7,29 @@ authentication:
...
     scopes:
     - "read:metadata"
     - "read:data"
-    public:
-    - D
+    - "write:metadata"
+    - "write:data"
+    - "create"
+    tags_db:
+      uri: "file:example_configs/access_tags/compiled_tags.sqlite"
```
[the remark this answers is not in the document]
nmaytan: I might agree - maybe it would make sense to pipe this through [ensure_specified_sql_driver](https://github.com/bluesky/tiled/blob/7158cd60aa2e508c1a3fcfd9db1230a15062f44b/tiled/utils.py#L819) and raise a specific exception if we get any driver besides `sqlite+aiosqlite:///`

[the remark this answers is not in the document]
nmaytan: Give 8cdefb0 a glance, this accepts:
* `sqlite:///`
* `sqlite+aiosqlite:///`
* `file:`
* simple path to db

tiled/client/context.py
```diff
@@ -875,7 +886,9 @@ def show_principal(self, uuid):
                     )
                 ).json()
 
-    def create_api_key(self, uuid, scopes=None, expires_in=None, note=None):
+    def create_api_key(
+        self, uuid, scopes=None, access_tags=None, expires_in=None, note=None
```
danielballan, line 890: For back-compat, better to add the new kwarg at the end.

danielballan (approved):

PR #1010, opened 2025-07-11
Fix reading out arrays from database storage

### LINKED ISSUE #1037: Arrays stored in Postgres are read out as their string representation
*copied from the Mattermost for future reference*:

> Hello,
> 
> I'm not sure if this is a bug, or whether there is a special way of reading waveform/numpy arrays stored in Postgres. We are testing things with the Ophyd `SynPeriodicSignal`, passed through the TiledWriter. The database schema shows the correct column data type (`double precision[]`), and the catalog metadata seems to be correct as well: 
> ```python
> {'data_keys': {
>     ...
>     'sim_spec_line': {
>         'dtype': 'array',
>         'dtype_numpy': '<f8',
>         'object_name': 'sim_spec_line',
>         'precision': 3,
>         'shape': [4096],
>         'source': 'SIM:sim_spec_line'
>      }
>  },
> }
> ```
> When reading the data through the ArrayClient, it seems to convert the NDArray to an ellipsed string at some point, with data type "<U102".
> ```python
> In [61]: arr = client["a0d153cf-67f5-42be-b322-15ab10b81a3c"]["streams/primary/sim_spec_line"]
> 
> In [62]: arr
> Out[62]: <ArrayClient shape=(1070,) chunks=((1070,),) dtype=<U102>
> 
> In [63]: arr.read()
> Out[63]: 
> array(['[-0.00656741 -0.04152672 -0.06356263 ...  0.01019265 -0.03305269\n -0.00117206]',
>        '[ 0.05749623  0.0738975  -0.10622574 ...  0.00866997  0.02036999\n -0.02306699]',
>        '[-0.03044402 -0.00865704  0.02546475 ... -0.09221372  0.03200744\n -0.01383083]',
>        ...,
>        '[ 0.00058396 -0.05751252 -0.04448293 ... -0.04872262  0.03310412\n  0.00636689]',
>        '[-0.0394815   0.00756512 -0.0665533  ...  0.0065802  -0.00323965\n  0.02780476]',
>        '[ 0.01984417  0.00830684 -0.02297906 ...  0.02111964  0.01840572\n  0.02363917]'],
>       shape=(1070,), dtype='<U102')
> 
> In [64]: arr.structure()
> Out[64]: ArrayStructure(data_type=BuiltinDtype(endianness=<Endianness.little: 'little'>, kind=<Kind.unicode: 'U'>, itemsize=408, dt_units=None), chunks=((1070,),), shape=(1070,), dims=None, resizable=False)
> ```
> 
> (`U102` is the decimal unicode value for lowercase "f", if that gives any hints)

cjboyle:
This fixes reading out array columnar data stored in Postgres, e.g. from `waveform` PVs. Prior to this the data is an array of arrays, where the top-level datatype is `object`. As such, this now checks for underlying Numpy arrays and combines them into a ND array with the underlying datatype bubbled up.

### Checklist
- [x] Add a Changelog entry

resolves #1037


danielballan:
Thank you! Would be up for updating the unit test(s) to cover this new code path? See https://github.com/bluesky/tiled/blob/main/tiled/_tests/adapters/test_sql_types.py

cjboyle:
Though it did not seem to affect the original issue, I've extended the `ARROW_TO_DUCKDB_TYPES` and `ARROW_TO_PG_TYPES` lists to add array column types (along with relevant tests). SQLite *might* work, using the `BLOB` type, but extra type information (bit size) would obviously be lost.

As noted in the new tests for `ArrayAdapter`, my fix will work for strictly uniform-sized arrays. If jagged arrays are intended to be supported a different solution will be necessary.

tiled/_tests/adapters/test_sql_arrays.py
```diff
@@ -0,0 +1,224 @@
...
+from tiled._tests.adapters.test_sql import assert_same_rows
+from tiled.adapters.sql import SQLAdapter
+from tiled.storage import parse_storage, register_storage
+from tiled.structures.core import StructureFamily
+from tiled.structures.data_source import DataSource, Management
+from tiled.structures.table import TableStructure
+
+names = ["f0", "f1", "f2", "f3", "f4", "f5"]
+data0 = [
+    pa.array([np.ones(3, dtype=np.int8) * i for i in range(1, 6)]),
```
genematx, line 22: I wonder if doing something like
```
rng = numpy.random.default_rng(12345)
arr = rng.random(size=(3,), dtype="float64")
```
could be a better test here, especially since we're checking for equality eventually?

tiled/_tests/test_array.py
```diff
@@ -166,6 +181,25 @@ def test_array_interface(context):
...
+def test_uniform_nested_array_projected_to_ndarray(context):
+    client = from_context(context)["nested_arrays"]["uniform"]
+    assert client.dtype == numpy.int_
+    assert client.read().dtype == numpy.int_
+    assert numpy.array_equal(client.read(), nd_array)
+
+
+@pytest.mark.parametrize("kind", ["ragged", "objects"])
+def test_unparsable_nested_array_stringified(kind, context):
+    # This behavior is due to the fact that ragged Numpy arrays, and those with
```
genematx, line 193: Perhaps this could be a good future use case for the `AwkwardClient`.

cjboyle: Yeah... I've been somewhat cognizant that my fix is more of a bandaid to a specific use-case of arrays in SQL, so I think ideally the fix should actually be applied earlier in `SQLAdapter.__getitem__` before calling `ArrayAdapter.from_array(...)`. 

I hadn't really thought of it, but I'll test to see if I can make a better solution that considers `AwkwardAdapter`.

genematx: oh, it was just a thought... :) I don't think we need to implement awkward (or ragged) arrays for this right away, but could be a good candidate since we have those adapters.
Your bandaid is much better than what I did before with converting them to strings!

cjboyle:
I had a hiccup at the finish line updating things to consider Awkward, so I'll table that for a future PR.

PR #1011, opened 2025-07-14
Use correct api_key when printing server details

tpoliaw:
When no api key was given manually when running `tiled serve directory`, the one used by the server was different from the one printed to the user.

This is not the case for other ways of running the server as they do not need to retain the api key to run a client against the newly started server.

### DIFF
--- a/tiled/commandline/_serve.py
+++ b/tiled/commandline/_serve.py
@@ -5,6 +5,8 @@ from typing import List, Optional
 
 import typer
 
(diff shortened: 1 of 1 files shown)


tpoliaw:
I think it's only `serve directory` due to the 

> \# Set the API key manually here, rather than letting the server do it,
> \# so that we can pass it to the client.

In the other cases both the `print_server_info` method and the actual server use the generated api key from the `Settings` instance.

PR #1012, opened 2025-07-15
Fix zarr test

ZohebShaikh:
Currently these tests are not running on main due to unsupported use of yield, more info [here](https://github.com/bluesky/tiled/actions/runs/16326120465/job/46116588017#step:6:891)

### Checklist
- [ ] Add a Changelog entry


danielballan (approved):
Good catch!

PR #1013, opened 2025-07-15
Add filters for warnings in tests

ZohebShaikh:
### Checklist
- [ ] Add a Changelog entry


ZohebShaikh:
These warnings can be caught in the `pyproject.tomal` as well but I did not know the preference that NSLS-II has so I went with the safer choice of adding warning filters on the tests with warnings. Later when the zarr test are fixed we could add give error for warnings

PR #1016, opened 2025-07-15
Use clearer error message when server config is non-file

tpoliaw:
`assert False` is currently used as the comment assumes it is
unreachable but the 'impossible to reach' state can be reached if a
non-normal file is passed as config, for instance

    mkfifo /tmp/demo
    TILED_CONFIG=/tmp/demo uvicorn tiled.server.app:app

would result in 'AssertionError: It should be impossible to reach this
line.'


danielballan (approved):

danielballan (approved):
CI failures are unrelated. (Fix in progress...)

PR #1018, opened 2025-07-16
Ensure background tasks are run and cancelled correctly

tpoliaw:
* Due to a mismatch in keys (`background` vs `background_tasks`), any
  background tasks passed in arguments to build_app would not be run.

* Any background tasks from authenticators would be run but would only
  be cancelled on shutdown if a database was configured.

To help avoid future errors, the tasks map is now a TypedDict so that
type checkers can flag access to missing keys. Access to these keys can
now be direct rather than requiring `.get()` with a fallback.

### DIFF
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -11,7 +11,7 @@ from contextlib import asynccontextmanager
 from functools import cache, partial
 from pathlib import Path
 from textwrap import dedent
-from typing import Optional, Union
+from typing import Any, Callable, Coroutine, Optional, TypedDict, Union
 
 import anyio
 import packaging.version
@@ -76,6 +76,10 @@ logger.addHandler(handler)
 current_principal = contextvars.ContextVar("current_principal")
 
 
+AppTask = Callable[[], Coroutine[None, None, Any]]
+"""Async function to be run as part of the app's lifecycle"""
+
+
 def custom_openapi(app):
     """
     The app's openapi method will be monkey-patched with this.
@@ -112,7 +116,7 @@ def build_app(
     deserialization_registry: Optional[SerializationRegistry] = None,
     compression_registry: Optional[CompressionRegistry] = None,
     validation_registry: Optional[ValidationRegistry] = None,
-    tasks=None,
+    tasks: Optional[dict[str, list[AppTask]]] = None,
     scalable=False,
     access_policy=None,
 ):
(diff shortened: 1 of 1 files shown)


tiled/server/app.py
```diff
@@ -111,7 +111,7 @@ def build_app(
     deserialization_registry: Optional[SerializationRegistry] = None,
     compression_registry: Optional[CompressionRegistry] = None,
     validation_registry: Optional[ValidationRegistry] = None,
-    tasks=None,
+    tasks: Optional[dict[str, list[Any]]] = None,
```
DiamondJoseph, line 114:
```suggestion
    tasks: Optional[dict[str, list[Awaitable]]] = None,
```

Shouldn't this be required, as we await the items in the list?

tiled/server/app.py
```diff
@@ -909,3 +911,9 @@ def print_server_info(
 
 class UnscalableConfig(Exception):
     pass
+
+
+class TaskMap(TypedDict):
+    background: list[Any]
+    startup: list[Any]
+    shutdown: list[Any]
```
DiamondJoseph, line 919:
```suggestion
    background: list[Awaitable]
    startup: list[Awaitable]
    shutdown: list[Awaitable]
```

perhaps?

tiled/server/app.py
tpoliaw: I think they should be callables that return awaitables but I was being lazy. Have now updated (here and in the typeddict)

tiled/server/app.py
```diff
@@ -111,7 +115,7 @@ def build_app(
     deserialization_registry: Optional[SerializationRegistry] = None,
     compression_registry: Optional[CompressionRegistry] = None,
     validation_registry: Optional[ValidationRegistry] = None,
-    tasks=None,
+    tasks: Optional[dict[str, list[AppTask]]] = None,
```
DiamondJoseph, line 118:
```suggestion
    tasks: Optional[TaskMap] = None,
```

How far away are we from being able to

DiamondJoseph (approved):

danielballan:
There were merge conflicts, so I took the liberty of rebasing on `main` and force-pushing. We'll see if CI passes.

danielballan (approved):

danielballan:
Thanks for the cleanup, @tpoliaw!

PR #1020, opened 2025-07-16
Use dedent to allow indented multiline strings

tpoliaw:
Keeping the multiline text blocks indented to the level of the code
around them makes it easier to see the context and structure of the
code compared to them jumping back to the left margin.


danielballan (approved):
Nice pattern! Thank you.

PR #1021, opened 2025-07-17
🪣 Support for OBject Storage

### LINKED ISSUE #905: Support writable blob storage

Kezzsim:
Resolves #905

### DIFF
--- a/tiled/adapters/zarr.py
+++ b/tiled/adapters/zarr.py
@@ -4,7 +4,7 @@ import copy
 import os
 from importlib.metadata import version
 from typing import Any, Iterator, List, Optional, Set, Tuple, Union, cast
-from urllib.parse import quote_plus, urlparse
+from urllib.parse import quote_plus, urljoin, urlparse
 
 import zarr
 from numpy._typing import NDArray
@@ -18,7 +18,13 @@ from ..adapters.utils import IndexersMixin
 from ..catalog.orm import Node
 from ..iterviews import ItemsView, KeysView, ValuesView
 from ..ndslice import NDSlice
-from ..storage import FileStorage, ObjectStorage, Storage, get_storage
+from ..storage import (
+    SUPPORTED_OBJECT_URI_SCHEMES,
+    FileStorage,
+    ObjectStorage,
+    Storage,
+    get_storage,
+)
 from ..structures.array import ArrayStructure
 from ..structures.core import Spec, StructureFamily
 from ..structures.data_source import Asset, DataSource
@@ -31,7 +37,6 @@ if ZARR_LIB_V2:
     from zarr.storage import DirectoryStore as LocalStore
     from zarr.storage import init_array as create_array
 else:
-    from obstore.store import AzureStore, GCSStore, S3Store
     from zarr import create_array
     from zarr.storage import LocalStore, ObjectStore
 
@@ -68,28 +73,27 @@ class ZarrArrayAdapter(Adapter[ArrayStructure]):
         # Use the first chunk along each dimension.
         zarr_chunks = tuple(dim[0] for dim in data_source.structure.chunks)
         shape = tuple(dim[0] * len(dim) for dim in data_source.structure.chunks)
+        data_type = data_source.structure.data_type.to_numpy_dtype()
+        data_uri = urljoin(
+            storage.uri + "/", "/".join(quote_plus(segment) for segment in path_parts)
+        )
 
-        data_path = "/".join(quote_plus(segment) for segment in path_parts)
-        if isinstance(storage, ObjectStorage):
-            zarr_store = ObjectStore(store=storage.get_object_store(prefix=data_path))
+        if ZARR_LIB_V2:
+            zarr_store = LocalStore(str(path_from_uri(data_uri)))
+        else:
+            zarr_store = ObjectStore(store=storage.get_obstore_location(data_uri))
 
-        if isinstance(storage, FileStorage):
-            directory = path_from_uri(storage.uri + data_path)
-            directory.mkdir(parents=True, exist_ok=True)
-            zarr_store = LocalStore(str(directory))
+        create_array(zarr_store, shape=shape, chunks=zarr_chunks, dtype=data_type)
+
+        # Update data source to include the new asset
         data_source.assets.append(
             Asset(
-                data_uri=storage.uri + data_path,
+                data_uri=data_uri,
                 is_directory=True,
                 parameter="data_uri",
             )
         )
-        create_array(
-            zarr_store,
-            shape=shape,
-            chunks=zarr_chunks,
-            dtype=data_source.structure.data_type.to_numpy_dtype(),
-        )
+
         return data_source
 
     @property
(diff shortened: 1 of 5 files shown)


tiled/adapters/zarr.py
```diff
@@ -46,29 +49,46 @@ def init_storage(
...
+                storage.uri
+                + "/"
+                + storage.config["bucket"]
+                + "".join(f"/{quote_plus(segment)}" for segment in path_parts)
+            )
+            mapping = {"s3": S3Store, "azure": AzureStore, "google": GCSStore}
+            urlProp = {"s3": "endpoint", "azure": "endpoint", "google": "url"}
+            object_store_class = mapping[storage.provider]
+            object_store = object_store_class(
+                **{urlProp[storage.provider]: storage.uri},
```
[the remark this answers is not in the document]
genematx: also, from what I understand, `urlProp` is used somewhere down the line, not by the obstore class itself, right? E.g., `obstore.store.S3Store` used for minio/s3 has the following signature (without `endpoint`):
```
__init__(
    bucket: [str](https://docs.python.org/3/library/stdtypes.html#str) | None = None,
    *,
    prefix: [str](https://docs.python.org/3/library/stdtypes.html#str) | None = None,
    config: [S3Config](https://developmentseed.org/obstore/latest/api/store/aws/#obstore.store.S3Config) | None = None,
    client_options: [ClientConfig](https://developmentseed.org/obstore/latest/api/store/config/#obstore.store.ClientConfig) | None = None,
    retry_config: [RetryConfig](https://developmentseed.org/obstore/latest/api/store/config/#obstore.store.RetryConfig) | None = None,
    credential_provider: [S3CredentialProvider](https://developmentseed.org/obstore/latest/api/store/aws/#obstore.store.S3CredentialProvider) | None = None,
    **kwargs: Unpack[[S3Config](https://developmentseed.org/obstore/latest/api/store/aws/#obstore.store.S3Config)],
) -> None
```

tiled/storage.py
```diff
@@ -244,6 +247,22 @@ def register_storage(storage: Storage) -> None:
...
-    "Look up Storage by URI."
-    return _STORAGE[uri]
+def get_storage(uri: str) -> Storage | Tuple[Storage, str]:
+    scheme = urlparse(uri).scheme
+    if scheme == "file":
+        return FileStorage(uri)
+    elif scheme in {"sqlite", "duckdb"}:
+        return EmbeddedSQLStorage(uri)
+    elif scheme == "http":
+        # Split on the first single '/' that is not part of '://'
```
danielballan: This works but it's a lot to digest:

- regex
- string emptiness check
- `lstrip`
- `replace`

Above, we have the parsed URL `urlparse(uri)`, which we use to get the `scheme`. If we keep a reference to that in some local variable, `parsed_uri`, we can easily grab the path:

```py
full_path = parsed_uri.path  # includes bucket and the rest
bucket_name, blob_path = full_path.split("/", 1)
```

Aside: If you haven't seen it before, the second argument to `split` works like this:

```py
>>> 'a/b/c'.split('/')
['a', 'b', 'c']
>>> 'a/b/c'.split('/', 1)  # split on the first instance of '/' only
['a', 'b/c']
```

Now, to get the "bucket only" version of the URL:

```py
from urllib.parse import urlunparse

# Get the URL without the blob_path, just the bucket.
bucket_uri = urlunparse(parsed_uri._replace(path='/' + bucket_name))
# Look up storage by bucket URI.
storage = _STORAGE[bucket_uri]
# Return a copy encapsulating the blob_path.
return storage.with_blob_path(blob_path)  # Note: This method would need to be added to ObjectStorage.
```

Example:

```py
>>> uri = 'https://example.com/bucket/a/b/c'
>>> parsed_uri = urlparse(uri)
>>> parsed_uri
ParseResult(scheme='https', netloc='example.com', path='/bucket/a/b/c', params='', query='', fragment='')

>>> full_path = parsed_uri.path
>>> full_path
'/bucket/a/b/c'

>>> full_path.lstrip('/')
'bucket/a/b/c'

>>> bucket_name, blob_path = full_path.lstrip('/').split('/', 1)
>>> bucket_name
'bucket'
>>> blob_path
'a/b/c'
```

Kezzsim:
Pertaining to the minio testing container:

- [X] Specify Object fixture in `test_writing.py`
- [ ] Set fixture to xfail when minio url isn't available
- [X] Add GitHub Actions to setup the minio container
- [ ] Add code to reset the state of the bucket after each test

tiled/adapters/zarr.py
```diff
@@ -60,29 +63,33 @@ def init_storage(
...
-        store = LocalStore(str(directory))
-        create_array(
-            store,
-            shape=shape,
-            chunks=zarr_chunks,
-            dtype=data_source.structure.data_type.to_numpy_dtype(),
-        )
+
+        data_path = "/".join(quote_plus(segment) for segment in path_parts)
+        if isinstance(storage, ObjectStorage):
```
genematx: should this raise if zarr is v2? or rather should the error be raised here, or somewhere else?

Kezzsim: I think (correct me if I'm wrong @danielballan  ) that we're ready to move past V2 and stop supporting it, at least when it comes to bucket storage.

tiled/catalog/adapter.py
genematx: Since the writing adapters are all server-side (and we control them), do we really need backcompat here? @danielballan

PR #1025, opened 2025-07-21
🗝️ Add "KeyIn" & "KeyNotIn" queries

Kezzsim:
Requested by @cjtitus in NSLS-II Ticket [SXSS-592](https://jira.nsls2.bnl.gov/browse/SXSS-592).
This attempts to register queries that can tell if a container's metadata has a **KEY** matching a particular string in it.

- [x] Register the queries
- [x] Finalize the business logic to perform the queries
- [ ]  Add specific tests for methods
- [ ] Add queries to [Tiled Docs](https://blueskyproject.io/tiled/reference/queries.html)

### DIFF
--- a/tiled/adapters/mapping.py
+++ b/tiled/adapters/mapping.py
@@ -790,10 +790,10 @@ def keynotin(query: Any, tree: MapAdapter) -> MapAdapter:
 
     """
     matches = {}
-    if len(query.value) == 0:
+    if len(query.key) == 0:
         return tree
     for key, value, term in iter_child_metadata(query.key, tree):
-        if term not in query.value:
+        if term not in query.key:
             matches[key] = value
     return tree.new_variation(mapping=matches)
 
--- a/tiled/adapters/mapping.py
+++ b/tiled/adapters/mapping.py
@@ -28,9 +28,11 @@ from ..queries import (
     Eq,
     FullText,
     In,
+    KeyIn,
     KeysFilter,
     NotEq,
     NotIn,
+    KeyNotIn,
     Regex,
     SpecsQuery,
     StructureFamilyQuery,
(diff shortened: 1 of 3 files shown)


danielballan:
The behavior looks aligned with the requirements now. Seems ready for adding docs and unit tests in `tiled._tests.test_queries`, particular on nested keys.

We should also give some consideration for the name. I think `KeyIn` and `KeyNotIn` is confusing here. Either of @cjtitus' suggestions seem reasonable:

> Possible name pairs: HasKey/NotHasKey ? Exists/NotExists? Something better?

I'll add `KeyExists` / `KeyMissing` for consideration. (But not sure if that's better than the above.)

danielballan:
Oh, great idea. Let’s go with that.

If we start with just `KeyExists` we can always elect to add `KeyAbsent` later as an alias—if we find that the optional boolean parameter is hard for users to discover.

tiled/catalog/adapter.py
```diff
@@ -1479,6 +1480,19 @@ def in_or_not_in(query, tree, method):
     return _IN_OR_NOT_IN_DIALECT_DISPATCH[dialect_name](query, tree, method)
 
 
+def has_key(query, tree):
+    # Functionally in SQLAlchemy 'is not None' does not work as expected
+    if tree.engine.url.get_dialect().name == "sqlite":
```
danielballan, line 1485: Due to changes in `main`, this now needs to be `tree.context.engine`. That should address the new test failures.

genematx:
I think the issue was with the import; there is
```
from typing import Callable, Dict, List, Optional, Union, cast
```
and also
```
from sqlalchemy.sql.expression import cast as sql_cast
```

danielballan:
Thanks, @genematx! Looking good, @Kezzsim.

Last thing: Above, I think @cjtitus made a persuasive case that KeyPresent/KeyAbsent are a nice pair of names, at the same time that you made the case for KeyExists with an optional boolean.

Thoughts on renaming KeyExists to KeyPresent? Then, if later on we find that we want to make the "reverse" more discoverable to users, we can always come back and add KeyAbsent. My feelings is that the boolean is likely good enough, but if we choose a name with a good antonym, we keep the option open for later.

tiled/adapters/mapping.py
```diff
@@ -754,7 +754,7 @@ def notin(query: Any, tree: MapAdapter) -> MapAdapter:
 MapAdapter.register_query(NotIn, notin)
 
 
-def keyexists(query: Any, tree: MapAdapter) -> MapAdapter:
+def keypresent(query: Any, tree: MapAdapter) -> MapAdapter:
```
danielballan, line 757: Nit-pick: Can it be `key_present`?

tiled/catalog/adapter.py
```diff
@@ -1462,7 +1462,7 @@ def structure_family(query, tree):
 CatalogNodeAdapter.register_query(Contains, contains)
 CatalogNodeAdapter.register_query(In, partial(in_or_not_in, method="in_"))
 CatalogNodeAdapter.register_query(NotIn, partial(in_or_not_in, method="not_in"))
-CatalogNodeAdapter.register_query(KeyExists, has_key)
+CatalogNodeAdapter.register_query(KeyPresent, has_key)
```
danielballan, line 1465: Can we rename this function to `key_present` just to make it easier to find? The implementation names generally match the query names.

danielballan:
Testing interactively, I'm not convinced the nested keys work as advertised.

```py
In [29]: x = c.write_dataframe({'a': [1,2,3]}, key='x', metadata={"sample": {"color": "blue"}})

In [30]: c
Out[30]: <Container {'x'}>

In [31]: c.search(KeyPresent("sample.color"))
Out[31]: <Container {}>
```

Since this is almost certainly important for our applications (e.g. querying on keys nested inside `start.`) I think we should hold for that.

Kezzsim:
As discussed with @danielballan there's now a way to insert more specific test data on the fly to tiled which eventually will make db tests more atomic. `prep_test` exists now:

```
def prep_test(client, mapping):
    for k, v in mapping.items():
        client.write_array(
            v.read(),
            key=k,
            metadata=dict(v.metadata()),
            specs=v.specs,
        )
```

It does not work with map adapter presently because that appears to be static, but this does let us test nested keys.

PR #1027, opened 2025-07-22
refactor: Replace use of dependency_overrides for accessing root_tree

tpoliaw:
FastAPI provides the dependency_overrides mechanism for testing and it
is not intended to be used in normal operation.

In this case, the root_tree is already set as state on the app instance
so it is accessible via fastapi's built in dependency injection using
the app field of the request.


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ## v0.1.0-b31 (Unreleased)
 
+### Changed
+
+- Refactored internal server function ``get_root_tree()`` to not use FastAPI
```
tpoliaw, line 10: It's still using the dependency injection, it's just not using the overrides.

DiamondJoseph: I just copied the doc string from the last version- looks like the CI just needs linting run again and for us to find a good wording here

danielballan (commented):
Thanks! This misuse of `dependency_overrides` was a misstep from very early in the development, one I regret we took so long to revisit. Thanks for tidying it up.

PR #1028, opened 2025-07-23
Remove incorrect root_tree type hints

tpoliaw:
It's unclear what type the root_tree parameter should expect but I'm
unsure how it could ever be an instance of pydantic BaseSettings so
removing the hints for now on the basis that missing hints are better
than incorrect ones.


DiamondJoseph:
I believe it should be a `ContainerAdapter` - once we get to a point where ContainerAdapter is Generic on the types of Adapters it can contain it should be specifically a `ContainerAdapter[AnyAdapter]`

danielballan (approved):

PR #1031, opened 2025-07-23
Separate CLI client/server deps

### LINKED ISSUE #1029: CLI usage forces server dependencies since #1011
The changes from #1011 introduces breaking changes to "client-only" deployments -- forcing server dependencies to be installed.

Ideally tiled CLI would remain usable for client side commands in these environments.

padraic-shafer:
The changes from #1011 introduce breaking changes to "client-only" deployments -- forcing server dependencies to be installed.

This PR updates the CLI `serve` command to import the server settings only when this command is called.  This enables client-only environments to use client CLI commands without installing `tiled[server]` dependencies.

Resolves #1029.

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/commandline/_serve.py
+++ b/tiled/commandline/_serve.py
@@ -5,8 +5,6 @@ from typing import List, Optional
 
 import typer
 
-from tiled.server.settings import get_settings
-
 serve_app = typer.Typer(no_args_is_help=True)
 
 SQLITE_CATALOG_FILENAME = "catalog.db"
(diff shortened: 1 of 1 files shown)


tacaswell (approved):
We will likly see a minor performance gain as well.

padraic-shafer:
Codecov wants a test.  I'll start on that.

genematx (approved):

padraic-shafer:
Because the test environment requires both server and client dependencies, I'm so far only coming up with hacky and invasive ways to test this conditional import.

Are reviewers OK with skipping a test for this change?

tacaswell:
Yes, I'm OK with ignoring code-cov here.  It looks like it was already not tested and asking for new tests for this change is overkill.

padraic-shafer:
I'll go ahead and merge this based on the approvals and discussion above re: code-cov.

PR #1032, opened 2025-07-23
Add missing asserts

ZohebShaikh:
```
def test_pass():
   True == False
```
Tests like this will pass if it does not have assert

### Checklist
- [x] Add a Changelog entry


DiamondJoseph (commented):
All of this is good except I want to understand the regex

Issue #1024, opened 2025-07-21
Error concatenating `()` shapes in HDF5ArrayAdapter


thomashopkins32:
# Description

I have array data written under the same HDF dataset but across multiple HDF files. Within a single file's dataset, the shape is `()` and it contains a single value.

For example, let's say that I have 5 HDF files that all have the dataset `entry/data/x` that has shape `()` and the value `1.0`. What I expect to get back from tiled is an array of 5 values: `[1.0, 1.0, 1.0, 1.0, 1.0]`. Instead, I get the following error:
```
  File "cditools/.pixi/envs/dev/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 2470, in run_sync_in_worker_thread
    return await future
          ^^^^^^^^^^^^^
  File "cditools/.pixi/envs/dev/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 967, in run
    result = context.run(func, *args)
  File "cditools/.pixi/envs/dev/lib/python3.13/site-packages/tiled/adapters/hdf5.py", line 381, in from_catalog
    return HDF5ArrayAdapter.from_catalog(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        data_source,  # type: ignore
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
        **kwargs,
        ^^^^^^^^^
    )
    ^
  File "cditools/.pixi/envs/dev/lib/python3.13/site-packages/tiled/adapters/hdf5.py", line 232, in from_catalog
    array = cls.lazy_load_hdf5_array(
        *file_paths, dataset=dataset, swmr=swmr, libver=libver
    )
  File "cditools/.pixi/envs/dev/lib/python3.13/site-packages/tiled/adapters/hdf5.py", line 208, in lazy_load_hdf5_array
    array = dask.array.concatenate(arrs, axis=0) if len(arrs) > 1 else arrs[0]
            ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "cditools/.pixi/envs/dev/lib/python3.13/site-packages/dask/array/core.py", line 4475, in concatenate
    meta = _concatenate(seq_metas, axis=axis)
ValueError: zero-dimensional arrays cannot be concatenated
```

# My Suggested Solution

[...]

PR #1040, opened 2025-07-25
FIX: Allow HDF5 Adapter to Combine Scalar Values into Arrays

genematx:
If multiple files containing scalar values are passed to HDF5Adapter, concatenating them fails. This PR changes concatenation to stacking for these cases and ensures that if scalars are present in HDF5 files they are presented ~~either as arrays of shape `()` or `(n,)`, if `n > 1` files are open~~ as arrays of shape `(n,)` if `n>=1` files are open.

Issue: #1024 

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/_tests/test_hdf5.py
+++ b/tiled/_tests/test_hdf5.py
@@ -120,6 +120,7 @@ def test_from_file(example_file, buffer):
         file["a"]["b"]["c"]["d"]
 
 
+@pytest.mark.filterwarnings("ignore: The dataset")
 @pytest.mark.parametrize("key", ["d", "e", "f", "g", "h", "i"])
 def test_from_file_with_empty_data(example_file_with_empty_data, buffer, key):
     """Serve a single HDF5 file at top level."""
@@ -134,6 +135,7 @@ def test_from_file_with_empty_data(example_file_with_empty_data, buffer, key):
         file["a"]["b"]["c"][key]
 
 
+@pytest.mark.filterwarnings("ignore: The dataset")
 @pytest.mark.parametrize("key", ["int", "float", "str", "bytes", "bool"])
 def test_from_file_with_scalars(example_file_with_scalars, buffer, key):
     """Serve a single HDF5 file at top level."""
@@ -148,6 +150,7 @@ def test_from_file_with_scalars(example_file_with_scalars, buffer, key):
         file["a"]["b"]["c"][key]
 
 
+@pytest.mark.filterwarnings("ignore: The dataset")
 def test_from_file_with_vlen_str_dataset(example_file_with_vlen_str_in_dataset, buffer):
     """Serve a single HDF5 file at top level."""
     h5py = pytest.importorskip("h5py")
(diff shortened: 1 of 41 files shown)


tiled/_tests/test_hdf5.py
```diff
@@ -136,18 +136,20 @@ def test_from_file_with_empty_data(example_file_with_empty_data, buffer, key):
 
 
 @pytest.mark.filterwarnings("ignore: The dataset")
-@pytest.mark.parametrize("key", ["int", "float", "str", "bytes", "bool"])
-def test_from_file_with_scalars(example_file_with_scalars, buffer, key):
-    """Serve a single HDF5 file at top level."""
+@pytest.mark.parametrize("key", ["bool", "int", "float", "str", "bytes"])
+@pytest.mark.parametrize("num", [1, 5])
+def test_from_file_with_scalars(example_file_with_scalars, buffer, key, num):
```
DiamondJoseph, line 141:
```suggestion
@pytest.mark.parametrize("key", ["bool", "int", "float", "str", "bytes"])
@pytest.mark.parametrize("num", [1, 5])
def test_from_file_with_scalars(example_file_with_scalars, buffer, key: str, num: int):
```

While Dan is away you're at the mercy of me insisting on type hints ;)

thomashopkins32 (commented):
I believe it is still impossible to specify an empty shape `()` from the Ophyd level and have this work. I tested with a descriptor with shape `()` and attempting to read it back gives me the following error:
```
  File "/home/thopkins1/Documents/public-repos/tiled/tiled/adapters/hdf5.py", line 386, in from_catalog
    return HDF5ArrayAdapter.from_catalog(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        data_source,  # type: ignore
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
        **kwargs,
        ^^^^^^^^^
    )
    ^
  File "/home/thopkins1/Documents/public-repos/tiled/tiled/adapters/hdf5.py", line 249, in from_catalog
    raise ValueError(
    ...<2 lines>...
    )
ValueError: Shape mismatch between array data and structure: () != (1,)
```

The Eiger detector's FileWriter interface stores metadata in HDF5 files with `()` shape. We want to combine these across different files which will result in an array of these values.

genematx:
> I believe it is still impossible to specify an empty shape `()` from the Ophyd level and have this work. I tested with a descriptor with shape `()` and attempting to read it back gives me the following error:
> The Eiger detector's FileWriter interface stores metadata in HDF5 files with `()` shape. We want to combine these across different files which will result in an array of these values.

I think the question is what to do with _single_ scalars in the dataset (that are not concatenated/stacked). For `n` scalars, the shape of the result is `(n,)`, but a single value has original shape `()`, which we pass unchanged (similarly like we do not append leftmost dimensions to single non-concatenated arrays in HDF5Adapter; this happens before in Ophyd/Bluesky/TiledWriter).

Most likely, single scalars should be converted to shape `(1,)` for consistency.

thomashopkins32 (approved):
Works for me in the single scalar of shape `()`, multiple files case.

Possibly related to this issue is trying to read multi-dimensional arrays across multiple files. E.g. I have 10 HDF5 files with a single image of shape `(2048, 2048)` in each. I expect tiled to read it back as `(10, 2048, 2048)` (similar to how single scalars now work) but instead it does `(20480, 2048)`.

Unfortunately for me, Eiger detectors write single images with no leading singleton dimension so we are stuck with `(2048, 2048)` per file.

PR #1042, opened 2025-07-28
ENH: DataSources Index

genematx:
This adds a B-tree index on the `node_id` column of the `data_sources` table. The index supports the lookup and join queries and speeds up the node creation and access almost two-fold.

index_compare

### Checklist
- [x] Add a Changelog entry


danielballan:
Since CI cleared the last commit of substance, I did not wait for it to run again on the change to the CHANGELOG.

PR #1043, opened 2025-07-28
chore: Type hint Utils module

DiamondJoseph:
Adds type hints to the DictView/ListView and CachingMap implementations, to make it clearer how to use them elsewhere in the codebase. Ideally, DictView would be removed in favour of typing `Mapping` and ListView in favour of passing `tuples` which accomplish the same desired behaviours. There may also be a builtin that could suffice for the Caching map implementations.

### Checklist
- [ ] Add a Changelog entry


tiled/client/base.py
```diff
@@ -261,12 +262,12 @@ def metadata_copy(self):
...
 
     @property
-    def specs(self):
+    def specs(self) -> ListView[Spec]:
         "List of specifications describing the structure of the metadata and/or data."
         return ListView([Spec(**spec) for spec in self._item["attributes"]["specs"]])
 
     @property
-    def access_blob(self):
+    def access_blob(self) -> DictView[Any, Any]:
```
danielballan, line 270: This can also be constrained to be [JSON](https://github.com/bluesky/tiled/blob/406ea84e33b177a070bb02361c6f2c90c0546bdc/tiled/type_aliases.py#L15).

tiled/client/context.py
```diff
@@ -473,7 +473,7 @@ def from_app(
         return context
 
     @property
-    def tokens(self):
+    def tokens(self) -> DictView[Any, Any]:
```
danielballan, line 476:
```suggestion
    def tokens(self) -> DictView[Literal["access_token", "refresh_token"], str]:
```

tiled/media_type_registration.py
```diff
@@ -51,7 +52,7 @@ def __init__(self):
         for ext, media_type in self.DEFAULT_ALIASES.items():
             self.register_alias(ext, media_type)
 
-    def media_types(self, structure_family):
+    def media_types(self, structure_family) -> DictView[Any, Any]:
```
danielballan, line 55:
```suggestion
    def media_types(self, structure_family) -> DictView[str, str]:
```

danielballan:
Switching from `DictView` to `Mapping` may make sense. But I think `ListView` is distinct from `tuple`, semantically. A `tuple` is not merely "`list` but make it immutable". It is semantically a _record_ with distinctly typed fields, whereas a `list` is a collection of like objects, typically looped over.

danielballan (approved):
One small suggestion to tighten the types up a bit further, in one instance.

Merge at all.

PR #1044, opened 2025-07-28
chore(types): Use common base type for all access policy types

tpoliaw:
Prep work for #928 

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/type_aliases.py
+++ b/tiled/type_aliases.py
@@ -30,7 +30,7 @@ JSON = Mapping[str, JSON_ITEM]
 Scopes = Set[str]
 Query = Any  # for now...
 Filters = List[Query]
-AccessBlob = Dict[str, Any]
+AccessBlob = Mapping[str, Any]
 AccessTags = Set[str]
 
 AppTask = Callable[[], Coroutine[None, None, Any]]


danielballan:
I'm good with this, but I'd like to sequence merging it after @nmaytan's #1008 because it will be easier to rebase _this_ on _that_ than vice versa. (I'm happy to tackle the rebase.)

tiled/access_control/protocols.py
```diff
@@ -0,0 +1,38 @@
...
+        self,
+        node: BaseAdapter,
+        principal: Principal,
+        authn_access_tags: Optional[Set[str]],
+        authn_scopes: Scopes,
+        scopes: Scopes,
+    ) -> Filters:
+        pass
+
+    async def modify_node(
```
nmaytan, line 31: There should be a corresponding `init_node` method as well, besides that this looks good.

danielballan: I pushed a commit taking a crack at this, and additionally adding types to the signatures of the policy implementations. CI is failing, and I think I need to wait for fresh eyes to spot this issue, but it's close.

PR #1046, opened 2025-07-29
Add Connection Pooling for Storage DBs

genematx:
Add pooling for ADBC connections to storage databases. This implementation follows the official [suggested solution](https://github.com/apache/arrow-adbc/blob/main/docs/source/python/recipe/postgresql_pool.py) from Arrow, which instantiates a sqlalchemy `QueuePool` with a custom `creator` object set to an instance of an ADBC connection.

Furthermore, it is ensured that `SQLAdapter` closes (returns to the pool) the connection once it is done writing/reading the data. Additional method is added to the `SQLStorage` class to dispose of the pool when the server is shutting down to release the DB resources.

This prevents the Tiled server from swarming the DB with extra connections and significantly reduces the CPU load.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/adapters/sql.py
+++ b/tiled/adapters/sql.py
@@ -1,20 +1,7 @@
 import copy
 import hashlib
-import os
 import re
-from pathlib import Path
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    Iterator,
-    List,
-    Literal,
-    Optional,
-    Tuple,
-    Union,
-    cast,
-)
+from typing import Any, Callable, Iterator, List, Literal, Optional, Tuple, Union, cast
 
 import numpy
 import pandas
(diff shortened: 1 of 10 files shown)


tiled/adapters/sql.py
```diff
@@ -325,9 +285,12 @@ def append_partition(
...
         }:
             table = table.rename_columns(upr_lwr_case_mapping)
-        with self.conn.cursor() as cursor:
+
+        conn = self.storage.connect()
+        with conn.cursor() as cursor:
             cursor.adbc_ingest(self.table_name, table, mode="append")
-        self.conn.commit()
+        conn.commit()
+        conn.close()
```
danielballan, line 293: Can a context manager or a `finally` block be used to ensure that this runs even if `commit()` raises? (See also just below, same issue.)

tiled/storage.py
```diff
@@ -34,40 +42,96 @@ def path(self):
...
+
+        if self.uri.startswith("duckdb:"):
+            import adbc_driver_duckdb.dbapi
+
+            return adbc_driver_duckdb.dbapi.connect(str(filepath))
+
+        elif self.uri.startswith("sqlite:"):
+            import adbc_driver_sqlite.dbapi
+
+            return adbc_driver_sqlite.dbapi.connect(str(filepath))
```
danielballan, line 114: Some kind of check here, such as `else: raise ValueError(...)` would be appropriate since the object and method are public and might receive a different scheme.

danielballan (commented):
The general cleanup / restructuring here is great. I left some small suggestions.

danielballan (approved):

PR #1049, opened 2025-07-29
Type hints for Database objects

DiamondJoseph:
Uses the DeclarativeBase class as a the base for ORM models, which is the recommended SQLAlchemy 2.0 way of getting type hinting. Additionally type hints methods that take an AsyncSession or AsyncEngine.

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/authn_database/base.py
+++ b/tiled/authn_database/base.py
@@ -1,5 +1,7 @@
-from sqlalchemy.orm import declarative_base
+from sqlalchemy.orm import DeclarativeBase
+
 
 # Everything imports this so we put it in its own module to
 # avoid circular imports.
-Base = declarative_base()
+class Base(DeclarativeBase):
+    pass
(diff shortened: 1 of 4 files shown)


tiled/catalog/core.py
```diff
@@ -20,7 +21,7 @@
 REQUIRED_REVISION = ALL_REVISIONS[0]
 
 
-async def initialize_database(engine):
+async def initialize_database(engine: AsyncEngine):
     # The definitions in .orm alter Base.metadata.
     from . import orm  # noqa: F401
```
tpoliaw, line 26: This was removed for the authn case - should it be removed here as well? Not sure what side-effects the comment is referring to

danielballan: Yes.

To fill in some context, the comment refers that the ORM definitions need to be imported (i.e. `.orm` needs to be executed by the interpreter) before this call is run:

```python
# Create all tables.
 await connection.run_sync(Base.metadata.create_all)
```

because the ORMs implicitly modify `Base.metadata` at definition time. Since we do not import anything else from `.orm` here, we need to keep this import. It looks like the one in `authn_database` became redundant some point and was vestigial.

PR #1050, opened 2025-07-30
WIP extracting the authenticator model changes from #928

tpoliaw:


tpoliaw:
@danielballan I think it's reviewable now I've patched the rebase I mangled.

PR #1054, opened 2025-07-31
WIP: Allow bearer access token access

ZohebShaikh:
### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -435,6 +435,7 @@ def build_app(
             "refresh_token_max_age",
             "session_max_age",
         ]:
+            # Need to add OIDC providers information somewhere here 
             if authentication.get(item) is not None:
                 setattr(settings, item, authentication[item])
         for item in [
--- a/tiled/server/authentication.py
+++ b/tiled/server/authentication.py
@@ -4,7 +4,7 @@ import uuid as uuid_module
 import warnings
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
-from typing import Any, Optional, Sequence, Union
+from typing import Any, List, Optional, Sequence, Union
 
 from fastapi import (
     APIRouter,
@@ -143,12 +143,16 @@ def create_refresh_token(session_id, secret_key, expires_delta):
     return encoded_jwt
 
 
-def decode_token(token, secret_keys):
+def decode_token(token:str, secret_keys:List[str],allow_bearer_token:bool):
     credentials_exception = HTTPException(
         status_code=HTTP_401_UNAUTHORIZED,
         detail="Could not validate credentials",
         headers={"WWW-Authenticate": "Bearer"},
     )
+    
+    if allow_bearer_token:
+        ...
+        # payload = jwt.decode(token,settings.)
     # The first key in settings.secret_keys is used for *encoding*.
     # All keys are tried for *decoding* until one works or they all
     # fail. They supports key rotation.
@@ -212,7 +216,7 @@ async def get_decoded_access_token(
     if not access_token:
         return None
     try:
-        payload = decode_token(access_token, settings.secret_keys)
+        payload = decode_token(access_token, settings.secret_keys,settings.allow_bearer_token)
     except ExpiredSignatureError:
         raise HTTPException(
             status_code=HTTP_401_UNAUTHORIZED,
@@ -993,7 +997,7 @@ def authentication_router() -> APIRouter:
     ):
         "Mark a Session as revoked so it cannot be refreshed again."
         request.state.endpoint = "auth"
-        payload = decode_token(refresh_token.refresh_token, settings.secret_keys)
+        payload = decode_token(refresh_token.refresh_token, settings.secret_keys,allow_bearer_token=False)
         session_id = payload["sid"]
         # Find this session in the database.
         session = await lookup_valid_session(db, session_id)
@@ -1032,7 +1036,7 @@ def authentication_router() -> APIRouter:
 
     async def slide_session(refresh_token, settings, db):
         try:
-            payload = decode_token(refresh_token, settings.secret_keys)
+            payload = decode_token(refresh_token, settings.secret_keys,settings.allow_bearer_token)
         except ExpiredSignatureError:
             raise HTTPException(
                 status_code=HTTP_401_UNAUTHORIZED,
--- a/tiled/server/settings.py
+++ b/tiled/server/settings.py
@@ -25,6 +25,7 @@ class Settings(BaseSettings):
 
     tree: Any = None
     allow_anonymous_access: bool = False
+    allow_bearer_token: bool = False
     allow_origins: List[str] = Field(default_factory=list)
     authenticator: Any = None
     # These 'single user' settings are only applicable if authenticator is None.


ZohebShaikh:
Yes,I'm quite excited as well, In my mind I thought it might take quite a lot more effort to get this working but it looks to be quite simpler than I imagined. 

For the realm-config, The actual useful configuration is just ~250 lines the remaining is the default configuration. I just wanted one command `docker-compose up` to set everything up. I will try to find if there is a script I could write to reduce the config file.

The only thing  remaining to figure out is the access policy (scopes) and how to make it work with [OPA](https://www.openpolicyagent.org/)  (authz).

tiled/server/schemas.py
```diff
@@ -396,6 +397,7 @@ class Principal(pydantic.BaseModel):
     api_keys: List[APIKey] = []
     sessions: List[Session] = []
     latest_activity: Optional[datetime] = None
+    access_token: Optional[str] = None
```
DiamondJoseph, line 400: Shouldn't the principal be identifiable from the access token, rather than holding the latest(?) access token?

tiled/server/schemas.py
```diff
@@ -294,6 +294,7 @@ class DeviceCode(pydantic.BaseModel):
 class PrincipalType(str, enum.Enum):
     user = "user"
     service = "service"
+    jwt_token = "jwt_token"
```
DiamondJoseph, line 297: Shouldn't the jwt_token identify the `user` or `service` to which it belongs? In the `sub` of the access token or the identity token? We may need some predicate that determines if a particular subject (`k8s-i22-beamline` | `qwe67581`) is a `user` or a `service` - there may be insight in this thread? We might need to check with LDAP.... https://diamondlightsource.slack.com/archives/C03P6QB9589/p1758275763418459

tiled/server/authentication.py
```diff
@@ -144,27 +145,41 @@ def create_refresh_token(session_id, secret_key, expires_delta):
...
-            continue
+    payload = ""
+    if proxied_authenticator:
+        payload = jwt.decode(
+            token,
+            key=proxied_authenticator.keys,
+            algorithms=proxied_authenticator.id_token_signing_alg_values_supported,
+            audience=proxied_authenticator.audience,
+            issuer=proxied_authenticator.issuer,
+        )
```
DiamondJoseph, line 166:
```suggestion
    if proxied_authenticator:
        return jwt.decode(
            token,
            key=proxied_authenticator.keys,
            algorithms=proxied_authenticator.id_token_signing_alg_values_supported,
            audience=proxied_authenticator.audience,
            issuer=proxied_authenticator.issuer,
        )
```

nit: assigning to payload prior to the secret_keys loop is a change in behaviour if secret_keys contains only one element and it raises a JWTError, previously it would try and return payload before it is assigned and raise an exception, now it would return an empty string.

tiled/server/app.py
```diff
@@ -467,13 +468,20 @@ def override_get_settings():
...
+            # If we support authentication providers, we need a database, so if one is
+            # not set, use a SQLite database in memory. Horizontally scaled deployments
+            # must specify a persistent database.
+            settings.database_settings.uri = (
+                settings.database_settings.uri or "sqlite://"
+            )
+            # If the only authentication provider is a ProxiedOIDCAuthenticator,
+            # set it on the settings object for downstream use.
+            if len(authenticators) == 1:
+                _, authenticator = next(iter(authenticators.items()))
```
DiamondJoseph, line 481:
```suggestion
                authenticator = next(iter(authenticators.values()))
```

nit, it's a shame there's not an easier way to do this

tiled/authenticators.py
```diff
@@ -219,6 +220,54 @@ async def authenticate(self, request: Request) -> Optional[UserSessionState]:
...
+        )
+        self.scopes = scopes
+        self._oidc_bearer = OAuth2AuthorizationCodeBearer(
+            authorizationUrl=str(self.authorization_endpoint),
+            tokenUrl=self.token_endpoint,
+        )
+
+    @functools.cached_property
+    def keys(self):
+        return httpx.get(self.jwks_uri).raise_for_status().json().get("keys", [])
```
DiamondJoseph, line 264: This should have a time-to-live or other way of the cache being invalidated, as the oauth server should be rotating keys periodically.

DiamondJoseph: This cache also isn't being used in the superclass authenticate method, it's calling it each time it authenticates

tiled/access_control/scopes.py
```diff
@@ -20,6 +20,7 @@
     },
 }
 
+
```
DiamondJoseph, line 23:
```suggestion
```

nit: reduce the size of an already large PR to touch fewer files

tiled/access_control/access_policies.py
```diff
@@ -360,3 +367,54 @@ async def filters(self, node, principal, authn_access_tags, authn_scopes, scopes
...
+
+class Input(BaseModel):
+    input: Data
+
+
+class Decision(BaseModel):
+    result: bool
+
+
+class ExternalPolicyDecisionPoint:
```
DiamondJoseph, line 387: Extract these changes (the ExternalPolicy access_policy and OPA changes) into their own PR. That PR can be into this branch if it's required for merging this branch into main, but the external AuthN and the external AuthZ are orthogonal to each other and are making it harder to review this

tiled/server/authentication.py
```diff
@@ -362,7 +379,10 @@ async def get_current_scopes(
             api_key, settings, request.app.state.authenticated, db
         )
     elif decoded_access_token is not None:
-        return decoded_access_token["scp"]
+        if isinstance(settings.authenticator, ProxiedOIDCAuthenticator):
+            return set(decoded_access_token["scope"].split(" "))
```
DiamondJoseph, line 383: :+1: According to StackOverflow, [this is the standard format of scopes in an access token](https://stackoverflow.com/a/70184278)

tiled/server/authentication.py
```diff
@@ -1159,8 +1190,13 @@ async def revoke_session_by_id(
         return Response(status_code=HTTP_204_NO_CONTENT)
 
     async def slide_session(refresh_token, settings, db):
+        if isinstance(settings.authenticator, ProxiedOIDCAuthenticator):
+            return
```
DiamondJoseph, line 1194: Should: Add a comment about refreshing the token being the responsibility of the client/proxy in this case

example_configs/keycloak_oidc/oauth2-proxy.cfg
```diff
@@ -0,0 +1,14 @@
+## OAuth2 Proxy Config File
+## https://github.com/oauth2-proxy/oauth2-proxy
+
+email_domains = [
+    "*"
+]
+
+skip_jwt_bearer_tokens = true
+
+cookie_secret = "Dhf3pGspLQ5DjtzIz_la8mq2MkCsXzeV-js-HpeRXtA="
```
danielballan, line 10: Should this be set to some value that makes it clear this is a public dummy credential?

tiled/server/app.py
```diff
@@ -474,6 +475,12 @@ def override_get_settings():
                 settings.database_settings.uri = (
                     settings.database_settings.uri or "sqlite://"
                 )
+        if authenticators:
+            # ProxiedOIDCAuthenticator cannot be used alongside other authentication providers
+            if len(authenticators) == 1:
+                authenticator = next(iter(authenticators.values()))
+                if isinstance(authenticator, ProxiedOIDCAuthenticator):
+                    settings.authenticator = authenticator
```
danielballan, line 483: This passes silently if `ProxiedOIDCAuthenticator` is used alongside other authentication providers. It should raise a clear exception here, I think.

As an aside, it may be that Tiled should drop support for more using than one authenticator in the future, at which point this could be simplified.

tiled/authenticators.py
```diff
@@ -152,7 +153,7 @@ def _config_from_oidc_url(self) -> dict[str, Any]:
         response.raise_for_status()
         return response.json()
 
-    @functools.cached_property
+    @property
```
danielballan, line 156: Why remove `cached_property` here? While the `client_id` could potentially be changed, the new configuration won't take effect until you restart the server anyway, so I think there is no harm in caching it here.

tiled/authenticators.py
```diff
@@ -181,6 +182,19 @@ def authorization_endpoint(self) -> httpx.URL:
             cast(str, self._config_from_oidc_url.get("authorization_endpoint"))
         )
 
+    # TODO: Cache with expiration to allow for key rotation
```
danielballan, line 185: A `cachetools.TTLCache` might be appropriate here. (We use it elsewhere in the codebase, so we already have the dependency.)

tiled/server/authentication.py
```diff
@@ -1220,6 +1243,9 @@ async def revoke_session_by_id(
         return Response(status_code=HTTP_204_NO_CONTENT)
 
     async def slide_session(refresh_token, settings, db):
+        if isinstance(settings.authenticator, ProxiedOIDCAuthenticator):
+            # Session management should be handled by the client, not by Tiled.
+            return
```
danielballan, line 1248: I'm not sure that passing silently is the way to go here. If we reach this codepath, something has gone wrong, right? Should this be a RuntimeError?

tiled/server/schemas.py
danielballan, line 297: Can we workshop this name? Principals that are `user` type can also use JWT, so this might be confusing. What if we called it `external` or something like that?

danielballan (changes requested):
Awesome! This looks solid, just a couple suggestions and questions.

danielballan (approved):

PR #1056, opened 2025-07-31
UI/add unit tests

AditiChikkali:
This PR adds unit tests (using Vitest)
- Restructured file/folder structure to group components and their tests together for better maintainability and clarity.
- Added unit tests for individual components, covering pagination, page size controls, tab navigation, breadcrumb navigation, and download tab features


web-frontend/src/App.tsx
```diff
@@ -1,7 +1,7 @@
 import Container from "@mui/material/Container";
-import ErrorBoundary from "./components/error-boundary";
+import ErrorBoundary from "../src/components/error-boundary/error-boundary";
```
danielballan, line 2: For my general education, what's the thinking behind going up two levels `..` to then navigate back down into `/src`? Is it generally considered best practice to anchor to the project root?

AditiChikkali: The ../src/ path was a mistake, which is confusing and unnecessary. 
I'll fix this in the next commit. Thanks for the review!

danielballan: Oh, OK, great. The only other instance I see is just below here, in the same file.

Issue #1033, opened 2025-07-24
Container image 0.1.0-b30 failed to publish


DiamondJoseph:
```
2025-07-24T10:17:27.9335236Z #39 237.3 Building wheels for collected packages: tiled, numcodecs, ndindex
2025-07-24T10:17:28.0902085Z #39 237.3   Building wheel for tiled (pyproject.toml): started
2025-07-24T10:17:28.5419204Z #39 237.9   Building wheel for tiled (pyproject.toml): finished with status 'done'
2025-07-24T10:17:28.6477320Z #39 237.9   Created wheel for tiled: filename=tiled-0.1.0b30-py3-none-any.whl size=1758947 sha256=75366b2f2848e923765203ce7b0886c1faeea5eb7c26711b5021a509f32ce0ee
2025-07-24T10:17:28.6478988Z #39 237.9   Stored in directory: /tmp/pip-ephem-wheel-cache-oxjh3cn_/wheels/f1/48/1b/a62d5561052e80c3dcfa2c21d0074268a0e9cbbe483c0d680a
2025-07-24T10:17:28.6480055Z #39 238.0   Building wheel for numcodecs (pyproject.toml): started
2025-07-24T10:17:49.5820405Z #39 258.9   Building wheel for numcodecs (pyproject.toml): finished with status 'error'
2025-07-24T10:17:49.9144205Z #39 259.2   error: subprocess-exited-with-error
2025-07-24T10:17:49.9145342Z #39 259.2   
2025-07-24T10:17:49.9149269Z #39 259.2   × Building wheel for numcodecs (pyproject.toml) did not run successfully.
2025-07-24T10:17:49.9155146Z #39 259.2   │ exit code: 1
2025-07-24T10:17:49.9156390Z #39 259.2   ╰─> [298 lines of output]
2025-07-24T10:17:49.9157263Z #39 259.2       [numcodecs] setting up Blosc extension
2025-07-24T10:17:49.9157920Z #39 259.2       [numcodecs] compiling Blosc extension with SSE2 support
2025-07-24T10:17:49.9158590Z #39 259.2       [numcodecs] compiling Blosc extension with AVX2 support
2025-07-24T10:17:49.9159216Z #39 259.2       [numcodecs] setting up Zstandard extension
2025-07-24T10:17:49.9159840Z #39 259.2       [numcodecs] setting up LZ4 extension
2025-07-24T10:17:49.9160299Z #39 259.2       [numcodecs] setting up compat extension
[...]

PR #1057, opened 2025-08-01
Fix build

danielballan:
- In April we [pinned back numcodecs](https://github.com/bluesky/tiled/pull/950) to address a transient incompatibility with Zarr: numcodecs removed some APIs and Zarr took a moment to react. We should have opened an issue (and set a calendar reminder or something) to follow up. This PR removes the pin. Now pip is able to find a compatible wheel.
- Our CI was not exercising the multi-architecture Docker build, and thus this issue only revealed itself upon _release_. This PR now makes our container build on CI more closely mirror our build process on release. There is some subtlety in this, which is the cause of much of the commit thrashing.

Closes https://github.com/bluesky/tiled/issues/1033

### Checklist
- [ ] ~~Add a Changelog entry~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,11 +35,7 @@ FROM developer AS builder
 # Include server and client dependencies here because this container may be used
 # for `tiled register ...` and `tiled server directory ...` which invokes
 # client-side code.
-RUN if [ "$(uname -m)" = "aarch64" ]; then \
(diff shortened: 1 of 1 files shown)


danielballan:
Commit [4a4f434](https://github.com/bluesky/tiled/pull/1057/commits/4a4f4349e35981703a6fb615189c87a94e71ce60) worked:

image

In subsequent commits I reverted the exploratory changes that turned out to be not necessary.

PR #1060, opened 2025-08-04
Added new registries to export tables and arrays as csv files on Windows

jmaruland:
### Checklist
- [x] Add a Changelog entry

While running some tests for PR #991, I noticed that some export tests failed when I tried to run them in a Windows machine. After debugging the problem, I noticed that Windows uses application/vnd.ms-excel as a valid media type. I registered this media type for arrays and tables to fix the problem


danielballan:
Great! Can you fashion a unit test, perhaps in [`test_export.py`](https://github.com/bluesky/tiled/blob/main/tiled/_tests/test_export.py) that _would_ fail without this fix in place?

PR #1064, opened 2025-08-05
ENH: human-friendly repr for chunks

genematx:
This changes the `__repr__` for ArrayClient to avoid printing an excessively long string for large arrays with high number of small chunks, for example:

```
((1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1), (n,), (m,)) -> ((1, 1, ..., 1), (n,), (m,))
((1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2), (n,), (m,)) -> ((1, 1, ..., 1, 2), (n,), (m,))
((1, 2, 3, 4, 5, 6, 7, 8, 9, 10), (n,), (m,)) -> (variable, (n,), (m,))
```

### Checklist
- [x] Add a Changelog entry


tiled/client/utils.py
```diff
@@ -395,3 +395,27 @@ def get_asset_filepaths(node):
...
+    result = "("
+    for dim in chunks:
+        if len(dim) <= 5:
+            # Short dimensions, e.g. (1, 1, 1)
+            result += str(tuple(dim)) + ", "
+        elif len(set(dim)) == 1:
+            # All chunk sizes are the same, e.g. (1, 1, ..., 1)
+            result += f"({dim[0]}, {dim[0]}, ..., {dim[0]}), "
+        elif len(set(dim[:-1])) == 1:
+            # All chunk sizes but the last are the same, e.g. (1, ..., 1, 3)
```
danielballan, line 415: It very slightly tripped me up that this shows the first two elements if all are the same but only first one if the last is different.

Perhaps you should ignore this comment. Just a thought.

danielballan (approved):
Great idea. One note for consideration.

tiled/client/utils.py
genematx: hah, just wanted to make them same size, but I agree -- it also looked a bit off to me. Added one more.
In case of 6 elements, `(1, 1, ..., 1, 2)` doesn't give us much improvement though compared to `(1, 1, 1, 1, 1, 2)`.

danielballan (approved):

PR #1067, opened 2025-08-06
Use timedeltas for settings time values

tpoliaw:


tpoliaw:
Interesting, I'd not heard of a preference for division. Can't see that the `total_seconds` method will be going away any time soon though and looking at the implementation it's definitely not any less efficient.

PR #1070, opened 2025-08-06
Use pixi for GH workflows

danielballan:
For installation on GH Actions, use pixi instead of pip.

Update: This reduces the installation time from 90 seconds to 35.

### Checklist
- [ ] Add a Changelog entry


danielballan:
There is a previously-reported issue in pixi https://github.com/prefix-dev/pixi/issues/4080.

tacaswell:
A very short comment: I think it is better to use a pixi.toml to keep from mixing pypa packing with pixi packaging.  The cost of having to keep the dependencies 2x is worth it to avoid the inevitable chaos wheels will bring to the party.

@thomashopkins32 and I had a long discussion about this on another project.

.github/workflows/ci.yml
```diff
@@ -29,32 +29,20 @@ jobs:
...
-      shell: bash -l {0}
-      run: source continuous_integration/scripts/install.sh
-
-    - name: Install test requirements
-      shell: bash -l {0}
-      run: |
-        set -vxeuo pipefail
-        python -m pip install .[dev]
-        python -m pip list
+        environments: ${{ matrix.environment }}
```
thomashopkins32, line 45: I would add
```
cache: false
```
here. This way, the environment gets rebuilt for each CI run rather than re-use a different commit for `tiled`. I think this is an issue when using the following combination:
```
cache: true (the default)
```
and
```
tiled = { path = ".", editable = true }
```

Also, you can add
```
  activate_environment: true
```
so you don't have to use `pixi run -e dev` everywhere after this.

thomashopkins32 (commented):
One comment on the `setup-pixi` action.

I also agree that a separate `pixi.toml` is cleaner. Mixing the PyPI and conda-forge dependencies has lead to a lot of packaging headaches for me, personally.

jennmald (approved):

pyproject.toml
```diff
@@ -231,18 +233,19 @@ server = [
...
     "pandas",
     "pillow",
     "prometheus_client",
-    "pyarrow",
+    "pyarrow >=14.0.1",  # includes fix to CVE 2023-47248
     "python-dateutil",
     "python-jose[cryptography]",
     "python-multipart",
-    "sparse",
+    "sparse >=0.15.5",
```
genematx, line 248: at least in some of my environments I had to install `sparse` from pypi, because otherwise pixi couldn't resolve the versions (perhaps a mismatch with conda-forge).

danielballan:
I'm putting this on ice for now. We moved GH workflows from pip to uv. I would like to see us move to pixi eventually, but not with a high priority.

PR #1072, opened 2025-08-07
Remove registry overrides when building app kwargs

tpoliaw:
These were (per the comments) intended to be used for testing but do not
appear to be used anywhere.

### DIFF
--- a/tiled/config.py
+++ b/tiled/config.py
@@ -11,21 +11,18 @@ from collections import defaultdict
 from datetime import timedelta
 from functools import cache
 from pathlib import Path
-from typing import Optional
 
 import jsonschema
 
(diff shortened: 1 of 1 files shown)


danielballan:
I'll need to take a close look at this to recall what the intent what here. Will try to get to it today or tomorrow.

tpoliaw:
The registries can still be overridden in tests using the `build_app` method directly (as in `test_writing.py`).

For a bit of wider context, I'm hoping to incrementally get to the same state that #928 aimed for without being in one 1k line PR.

danielballan (approved):

danielballan:
OK, agreed. Thanks!

PR #1074, opened 2025-08-07
Add initial typing to config module

tpoliaw:


tpoliaw:
I think there is more typing that could be added but while the rest of the module is a bit up in the air, I think it's probably useful to include this as it is for now. We can always expand the typing later.

danielballan:
Works for me. Would you resolve conflcits?

danielballan:
This was failing static linter checks. I misread the output and merged it with that issue. It was fixed in https://github.com/bluesky/tiled/pull/774/commits/944875cc17897d6d77dc8bfe038cc9327b428df8. All good now.

PR #1086, opened 2025-08-13
Rewrite Dockerfile

### LINKED ISSUE #1059: Refurbish `Dockerfile` to use modern best practice
We should look at https://hynek.me/articles/docker-uv/ and also consider integrating pixi.

danielballan:
Motivation:
- Switch from pip to uv for speed. (When we get pixi working well I'd be happy to consider switching.)
- Separate installation of _dependencies_ from installation of tiled itself into separate steps, for better cache reuse.
- Byte-compile the Python code at build time for faster startup.

Recommendations taken from: https://hynek.me/articles/docker-uv/

Closes #1059 

Some of the recommendations are compatible with `docker build` but not yet supported by `podman build`, because of their use of BuildKit features. ~~We need a decision on whether it's worth forgoing support for `podman build` at least for now.~~ Decision: We'll use only podman-compatible features.

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -14,7 +14,7 @@
 	"features": {
 		"ghcr.io/devcontainers/features/git:1": {}
 	},
-	"postCreateCommand": "set -ex && uv sync --extra all --extra dev",
+	"postCreateCommand": "set -ex && uv sync --extra all --extra dev --project src && uv tool install pre-commit --with pre-commit-uv && SHELL=bash uv tool update-shell && pre-commit install",
 	"customizations": {
 		"vscode": {
 			"settings": {


DiamondJoseph:
DLS requires support of podman for at least devcontainers, so we would pretty please request a podman compatible build

danielballan:
We are a podman-only shop as well, so I'm on board with this. I hope podman catches up soon; these BuildKit features are nice!

danielballan:
Now `podman build --no-cache .` run locally—i.e. on slow-ish home WiFi—in under 5 minutes. Tolerable.

Crucially, the installation of the dependencies is cached in a separate layer from the installation of the app itself, so rebuilding with changes to the app only is cheap.

danielballan:
CI has been "stuck" here for several minutes....

image

danielballan:
Thanks. I have updated the config accordingly.

danielballan:
Builds have been typically taking more than hour; [a recent example](https://github.com/bluesky/tiled/actions/runs/17095336746/job/48478355550) took 1h 7m. The latest build on this PR  branch finished in under 27 minutes.

Issue #1088, opened 2025-08-14
Support only single configuration file


tpoliaw:
The current `config.py` module supports loading configuration from multiple files (eg if passed a directory it will load all the yaml files it contains). Is this functionality still required or would reducing the configuration to a single file (as suggested by the proposed CLI in #1069) be sufficient?

PR #1091, opened 2025-08-15
Use pydantic for config file loading

tpoliaw:
- **WIP First pass at using pydantic to parse config file**
- **Deal with the fallout of rewriting the config loading**

### DIFF
--- a/tiled/config.py
+++ b/tiled/config.py
@@ -8,7 +8,7 @@ import copy
 from datetime import timedelta
 from functools import cached_property
 from pathlib import Path
-from typing import Annotated, Any, Iterator, Optional, TypedDict, Union
+from typing import Annotated, Any, Iterator, Optional, Union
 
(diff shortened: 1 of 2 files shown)


danielballan:
Regarding #1088, if supporting only one config file makes this substantially easier, let's do it now.

danielballan:
I too really like the direction of build_app taking a single config instance. I will review the progress so far, today.

danielballan:
Sorry, I will not get to this in time for the start of the workday at DLS. It is on my list for Wednesday.

tpoliaw:
No problem, I'll fix up the last windows test, then I'm probably going to be working on other stuff for the next week or so.

danielballan:
I read through. This reads well and seems on track. :+1:

Do I understand that pydantic's config parsing can give us env var parsing for free? Is this in scope for this PR?

danielballan:
OK, that sounds good. We'll take this for a test drive and give it a round of line review, then put it in.

Issue #1081, opened 2025-08-12
Remove `Composite` Structure


genematx:
It turned out that the features of the `Composite` structure introduced in #668 -- namely, the flat name space across all arrays and table columns in a container -- create more complexity then bring benefits. The same functionality can be implemented via specs on the container objects and enforced by the Python client.

genematx:
Great, thank you! configurable validators sound like exactly what we need.

PR #1093, opened 2025-08-15
WIP: Remove Composite Structure Family

genematx:
WIP

Issue: #1081

### Checklist
- [ ] Add a Changelog entry


tiled/client/composite.py
```diff
@@ -53,8 +53,11 @@ def _flat_keys_mapping(self):
         return result
 
     @property
-    def parts(self):
-        return CompositeParts(self)
+    def base(self):
```
genematx, line 56: We can call something else, of course. `base` is just shortest relevant name I could think of. In future, we can also put it in the base class itself and make it a general accessor to the relevant Tiled client without specs -- it shouldn't break the current implementation.

PR #1101, opened 2025-08-20
Add Python 3.14 RCs to CI matrix

danielballan:
This adds two builds, `3.14.0rc2` and the free-threaded variant, `3.14.0rc2t`.

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,8 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
+          - "3.14.0rc2"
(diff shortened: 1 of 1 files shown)


danielballan:
Blocked by this, which is not a surprise:

```
  × Failed to build `llvmlite==0.44.0`
```

It might be useful to make the tests running _sans_ `sparse`, with the relevant tests just skipped.

ZohebShaikh:
It is time for [3.15](https://docs.python.org/3.15/whatsnew/3.15.html)

Issue #801, opened 2024-10-27
Add a `ragged` structure family


danielballan:
**tl;dr** Let's add first-class support for `ragged` structures in Tiled, remixing some patterns we have used for `array` and `awkward` structures, to properly represent data from hardware that can give us irregularly-sized arrays.

## Motivation

Sometimes hardware generates "ragged" array data: array data that cannot be represented in a rectangular array.

```py
[
    [1, 2, 3],
    [4, 5],
    [7, 8, 9, 10],
]
```

In "Databroker v0" (2015, pre-dating Tiled) we gave users either an iterable of arrays or a `pandas.Series` of arrays. Both of these were able to accommodate irregularly sized arrays. By placing each item in its own Python object, this approach was not as efficient---in memory footprint or speed of common operations---as a single array. For the common case where data _was_ regularly shaped, it was unnecessarily clumsy and slow.

Now, in the pre-release "Databroker v2", we currently always give users a single array. For regularly shaped data, this is both more usable and more efficient than a generator or a `pandas.Series`. But in the case where data is irregularly shaped (i.e. ragged), often due to intermittently flaky hardware, the data is now un-representable. Scientists sometimes elect to pad or trim into a regular shape in order to access the data. This situation is not good: it ought to be possible to represent the actual measurements.

## Ragged and Awkward

A new library, [ragged](https://github.com/scikit-hep/ragged?tab=readme-ov-file#ragged), enables representing arrays with irregularly-sized dimensions. It is more space- and time-efficient than using nested Python lists. Like numpy, it places all the data in a contiguous block of memory and holds metadata describing how a logical step through the N-dimensional array corresponds to a stride length through the linear block of memory. Numpy describes the stride-length along each dimension with a single number:

```py
>>> a = numpy.ones((3, 5))

>>> a.strides
(40, 8)
```

To describe the more complex structure of a ragged array, `ragged` needs to use a _list_ of offsets for each ragged dimension. This is implemented by wrapping an AwkwardArray. (AwkwardArrays can describe tree-like (JSON-like) structures. Ragged arrays are a subset of AwkwardArrays.)

```py
>>> x = ragged.array([[1,2,3], [4,5], [7,8,9,10]])

>>> x
ragged.array([
    [1, 2, 3],
    [4, 5],
    [7, 8, 9, 10]
])

>>> x.shape
(3, None)

>>> x._impl  # AwkwardArray tucked inside
[...]

PR #1104, opened 2025-08-23
Add support for `ragged` arrays

cjboyle:
This adds client and backend support for reading/writing irregular arrays using the the `ragged` package. As `ragged` is more or less a wrapper around `awkward`, ~~this PR reuses, or adds similar implementations from that structure family (e.g. serialization)~~.

Implements #801.

### Checklist
- [x] Add a Changelog entry
- [x] Writing ragged data from client (file storage)
- Reading ragged data to client (file storage)
  - [x] in full
  - [x] sliced
  - [x] from block/chunk
  - [x] from block/chunk, sliced
- [x] Writing `[None]` shaped data from Bluesky/TiledWriter into SQL storage
- [x] Reading data from `RaggedAdapter` returned by `SQLAdapter` (SQL storage)
- Reading ragged data to client (SQL storage)
  - [x] in full
  - [x] sliced
  - [ ] from block/chunk
  - [ ] from block/chunk, sliced
 - Serialization
   - [x] JSON
   - [x] Arrow
   - [x] Parquet
   - [x] w/ slicing (except scalar results)
   - [ ] others?


danielballan:
Awesome!

It looks like you found all the modules that need to be touched to add this.

The aspect that will need the most careful thought is the structure description and the HTTP APIs. These are designed to be used not only from the built-in Python client, but also from `curl` with tools like `jq`, browser-based applications, maybe Julia or Rust someday....

The Awkward `form` is quite complex. I suspect that only Python and C++ based clients, with access to `awkward` / `awkward-cpp` libraries, will be able to parse the `form` and engage with Tiled's Awkward structures in detail. (Unless, that is, IRIS-HEP builds Awkward libraries in other languages.) Clients without knowledge of Awkward can still get the data—exporting it to JSON, for example—but they probably cannot introspect or slice it in sophisticated ways.

If we were willing to similarly restrict `ragged` to clients with access to an `awkward` implementation, we wouldn't even really need to add a new structure family. We could implement it fully client-side, as a wrapper of the `awkward` client. But I see advantages in using the comparative simplicity of `ragged` to make it more accessible to simple clients.

This `form` construct is more flexible than `ragged` requires:

```
{'class': 'ListOffsetArray',
 'offsets': 'i64',
 'content': {'class': 'NumpyArray',
  'primitive': 'int64',
  'inner_shape': [],
  'parameters': {},
  'form_key': 'node1'},
 'parameters': {},
 'form_key': 'node0'}
```

A `ragged` form is always composed of one numpy "content" array and some number of "offset" arrays—full stop. It can be described thus (from #801):

```py
class RaggedStructure(ArrayStructure):
    shape: Tuple[None | int, ...]  # override base class which has this as Tuple[int, ...]
```

I'm not sure whether `ragged` _always_ puts offset arrays in `int64` dtype. If other uint types may be needed, then we will need a supplemental `offset_datatype`, similar to the supplemental `coord_datatype` in `sparse` structures.

https://github.com/bluesky/tiled/blob/f6a9509f213824a32a385f3277d320bd85fd2370/tiled/structures/sparse.py#L19-L23

Although reusing the awkward form keeps things simple _assuming your client already consumes awkward_ I think having a custom, much more constrained structure JSON, is worthwhile, to make ragged arrays a more portable and accessible concept.

cjboyle:
Currently using flattened numpy streams over the wire, and for file storage, which I think is working well for full arrays.

For fetching data, I need to rethink either slicing or server-to-client serialization of sliced data, partly because `ragged.reshape` is not yet implemented, mostly because the new shape must be known by the client. I'm also testing determining the new shape by applying the slice to the known structure shape property.

Likely a solution could be to return a JSON blob of form
```json
{
  "sliced_shape": [ <int>, <int|null>, ... ],
  "sliced_offsets": [ [ <int>, ... ], ... ], 
  "stream": <flattened octet-stream>
}
```
For short-term, could also just use the `to_json`/`from_json` which would keep structure, but sacrificing bandwidth.

cjboyle:
@danielballan I think this is just about ready for review. I guess a few items of note:

* Currently I'm storing file data as the flattened numpy byte-string representation. I think I will change this in favour of Parquet or JSON to support other clients that have file-local access.
* I haven't implemented a `PATCH` endpoint yet, as, unlike with NumPy, indexed/sliced data can't be edited in place. On the other hand I don't think it would be too difficult to allow appending to a dataset row-wise. Use of TiledWriter isn't affected by the lack of this.
* I may need some guidance on defining chunks for variable-sized axes, to use with the `/ragged/block` endpoints. I'm considering whether [Dask-Awkward](https://dask-awkward.readthedocs.io/en/stable/index.html) can help with this. Otherwise, I'll remove and disable those routes for now.

Otherwise, I think all the high-level features for ingestion and readout are completed.

tiled/client/ragged.py
```diff
@@ -0,0 +1,194 @@
...
+        return structure.size
+
+    @property
+    def dtype(self) -> np.dtype:
+        structure = cast("RaggedStructure", self.structure())
+        return structure.data_type.to_numpy_dtype()
+
+    @property
+    def nbytes(self) -> int:
+        return self.size * self.dtype.itemsize
```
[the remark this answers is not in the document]
cjboyle: Expanding on this a bit: the implication of this is that the size must be stored in the structure as well, which shouldn't be a big deal.

tiled/structures/ragged.py
```diff
@@ -0,0 +1,158 @@
...
+@dataclass(kw_only=True)
+class RaggedStructure(Structure):
+    data_type: BuiltinDtype | StructDtype
+    """Serializable representation of the array's data type"""
+    shape: tuple[int | None, ...]
+    """The shape of the array, where the first dimension is always a known integer,
+    and any variable dimensions are represented by None."""
+    size: int
+    """The total number of elements in the array."""
+    partitions: tuple[int, ...]
```
danielballan: Following convention, we use "chunks" to describe N-dimensional array chunks. The term "partitions" is applied to tables, where the chunking is inherently 1-dimensional, along the rows of the table.

tiled/structures/ragged.py
```diff
@@ -0,0 +1,158 @@
...
+    """Defines the boundaries for partitioning the array.
+    Note that the final value is the row count from `shape[0]`."""
+    nbytes: int
```
danielballan: This can be derived from the `size` and the `data_type`. I think that storing it separately over-determines the structure. Should it become a `property`?

danielballan (commented):
I leave it to @genematx to do a detailed review. I have just a couple comments on the structure, which I think it's important to get as right as we can from the start.

tiled/structures/ragged.py
cjboyle: That is how I originally had it (until yesterday :slightly_smiling_face:). The difference is that the underlying size from Awkward includes the size of the `np.int64` offsets data.

tiled/structures/ragged.py
danielballan: Oh wait, this might be the offsets? In that case I would suggest aligning with Awkward terminology and calling it `offsets`.

cjboyle: No your first comment was correct (chunks vs partitions). I was essentially following the terminology that `dask-awkward` uses for partitioning, because it is only being computed over the first dimension.

genematx: if I understand correctly, `partitions` here define the boundaries of splits along the left-most dimension; each partition is stored in its own parquet file (via `awkward.to_parquet`) -- so primarily they are needed to keep track of tiled's assets (if there are several files). I wonder if this can be done by awkward itself?

genematx:
@cjboyle I'm still going through my review and making some small changes, but I thought I'd push what I have so far, so we don't diverge in case you're working on this too. Here're my suggestions: https://github.com/cjboyle/tiled/pull/1 (mostly around tests and alembic migration; I've also updated it with main, so it may appear there are more unrelated changes)

tiled/structures/ragged.py
genematx: A `property` seems like a better fit for this, but do we even need it at all?

tiled/structures/ragged.py
cjboyle: That is correct. I don't believe this can be done in `awkward` proper. `dask-awkward` does provide some limited partitioning functionality, though see Dan's previous comment.

tiled/structures/ragged.py
cjboyle: I don't think it is strictly needed, no.

tiled/client/ragged.py
```diff
@@ -0,0 +1,273 @@
...
+                        params=url_params,
+                    ),
+                ).read()
+
+        return from_zipped_buffers(
+            buffer=content,
+            dtype=self.dtype,
+        )
+
+    def read_block(self, block: Any, slice: Any | None = None) -> ragged.array:
```
genematx: We probably don't need a separate `.read_block` method; it currently exists for the `ArrayClient`, but it likely will become deprecated in the near future (just the simple read + slice covers all necessary cases, especially since dask is not yet supported).

tiled/structures/ragged.py
```diff
@@ -0,0 +1,158 @@
...
+    # The total number of elements in the array
+    size: int
+    # Defines the boundaries for partitioning the array.
+    # Note that the final value is the row count from `shape[0]`.
+    partitions: tuple[int, ...]
+    # The size of the underlying dataset in bytes.
+    nbytes: int
+    # Optional tuple of dimension names, e.g. ("time", "x"), or None for unnamed dimensions
+    dims: tuple[str, ...] | None = None
+    resizable: bool | tuple[bool, ...] = False
```
genematx, line 77: i'm not sure if need `resizable` in the structure -- I don't think we use it, even for the usual arrays @danielballan

tiled/structures/ragged.py
genematx: What if we call it `chunks` and make it `tuple[tuple[int, ...] | None ...]`, like the usual array chunks? Of course, in case of `ragged`, `chunks` always be non-None only along the first dimension, but then the structure would be consistent with what we have for the usual arrays (and from what I understand the main idea is to make ragged arrays look and behave as normal arrays as much as possible, right?).

tiled/adapters/ragged_parquet.py
```diff
@@ -0,0 +1,118 @@
...
+from tiled.structures.ragged import RaggedStructure
+from tiled.utils import path_from_uri
+
+if TYPE_CHECKING:
+    from tiled.catalog.orm import Node
+    from tiled.structures.core import Spec
+    from tiled.type_aliases import JSON
+
+
+class RaggedParquetAdapter(Adapter[RaggedStructure]):
```
genematx: I've been thinking about the backend storage for this. While parquet is simple and is working, I wonder if it is the optimal choice here.

First, it raises the question, why don't we use parquet for the usual AwkwardArrays but instead store them in lower-level byte-arrays representation? Both methods have their pros and cons (mainly around the standardization of the interface vs storage size and performance), but it probably would make sense to keep them consistent, since ragged is a derivative of awkward (I think in a way it would be nice if we could read ragged storage with `AwkwardAdapter` -- even thought we probably would never do this). I have made some refactoring changes to `AwkwardAdapters` [here](https://github.com/bluesky/tiled/pull/1362) to make it easier to extend and adopt it (@cjboyle somehow I couldn't add you as a reviewer, but your comments are very welcome!).

On the other hand, the main advantage of ragged structures, as I see it, is the ability to append variably-sized data dynamically. (I believe this is how they are intended to be used at one of our applications at NSLS2.) I think we should consider a possibility for designing an appendable storage -- e.g. similar to what we have for appendable tables -- and this could be another difference from the base awkward class.

tiled/adapters/ragged.py
```diff
@@ -0,0 +1,81 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    import awkward
+    from numpy.typing import NDArray
+
+    from tiled.type_aliases import JSON
+
+
+class RaggedAdapter(Adapter[RaggedStructure]):
```
genematx, line 39: I'm also tempted to combine the in-memory and storage-backed adapters. Will work on this.

tiled/structures/ragged.py
cjboyle: That works for me. And just for confirmation, this would change from a list of bounding indices to a list of shapes?

```python
# old
partitions = [
    0,
    10,
    50,
    60,
    75,
    ...
]

# new
chunks = [
    (10, None, ...),
    (40, None, ...),
    (10, None, ...),
    (15, None, ...),
    ...
]

genematx: yeah, that's almost what I was thinking, more specifically
`chunks = [[10, 40, 10, 15], None, None, ...]`

tiled/adapters/ragged_parquet.py
cjboyle: Yeah, I was going both ways on this. In the first revision the adapter was storing the flattened data as `.npy` files for efficiency (with offsets stored in the structure object). I'm okay with reusing the Awkward storage implementation for less overhead.

I definitely had "implement append" on a to-do list somewhere... though this already works with the SQL storage for lists of 1D arrays. Yep, this could be done with either `ragged.concat` or `awkward.concatenate` with axis=0, or if you have a better idea using the file-backed containers directly.

tiled/structures/ragged.py
cjboyle: Ah right. I can get started on that.

tiled/client/ragged.py
cjboyle: Is the plan to deprecate this from just the clients, or from the adapters and router endpoints as well?

tiled/structures/ragged.py
cjboyle: I've refactored this in a way that I *think* is sane (I've added into the structure its own version of `shape_from_chunks` that helps with this). This also updates the parquet file output to be in line with the `sparse` implementation (e.g. `block-5.0.0.parquet`), bearing in mind that this may soon be switched to `DirectoryContainer`.

genematx: Thank you very much, Connor! I'll have a look at this asap; likely some time next week -- we have a really busy next couple of days here.

tiled/client/ragged.py
danielballan: It is not totally thought through yet, but I think we could support this on the client and in HTTP for back-compat, while dropping the need for it on the Adapters (implementing support just in terms of the read method).

genematx:
Hi @cjboyle , I've finally made [another PR](https://github.com/cjboyle/tiled/pull/2) into your branch with quite a few changes. I think the `RaggedStructure` looks solid now; thank you very much for implementing the chunks for it! I have added the functionality necessary for appendable writes. It uses a new `RaggedSQLAdapter` and required some changes to the underlying `SQLadapter` (already merged). I have also removed the parquet-based adapter for now; mainly to keep it simple, but also maybe we should explore other options for long-term (non-appendable) storage -- saving tables with array-typed columns is probably not ideal. There are just a couple of minor things left to do (more efficient reading of chunks, possibly extending the websockets, etc).

I might get a chance to look into this later this week, but wanted to get it to you anyway, so you can review. It looks like a lot of changes, but most of them are refactoring of what you already had there (+ SQL-based adapter, of course). Also, you may need to merge the `main` branch first to see a better diff. I'm happy to have a chat if you have any questions, and thank yiu again for your work on this!

PR #1115, opened 2025-08-28
FIX: set default None for optional patch

genematx:
Fixes an issue with PUT /data_source endpoint that requires a default values for the optional `patch` parameter to be set.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ Write the date in place of the "Unreleased" in the case a new version is release
 # Changelog
 
 
+## v0.1.0-b39 (Unreleased)
+
+### Fixed
+
+- Default paraneter (`None`) for the `patch` parameter in `PUT /data_source` endpoint.
+
+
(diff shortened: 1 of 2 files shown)


danielballan:
Per discussion with @dylanmcreynolds and @genematx, we will move the patch information into a query parameter, to stay truer to `PUT` semantics, where the body really should _be_ the resource, not have extraneous info like this.

Given that, we'll make this soon-to-be-removed field optional, which will un-break some experimental clients.

dylanmcreynolds:
This is probably too late, but I started thinking about [fragments](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment) for this info.

danielballan:
Not too late. But this limitation seems important:

>The fragment is not sent to the server when the URI is requested

dylanmcreynolds:
Oh... I see. I've seen it in some cases where tokens are placed in the fragment, I guess for the browser app to grab them. Makes sense.

Issue #1114, opened 2025-08-28
Failure in migrating database when upgrading past Tiled v0.1.0b32


canismarko:
I have a catalog of scans in a postgresql database.

This had previously been upgraded to match Tiled v0.1.0b32. Upon trying to upgrade to anything 0.1.0b33 -- 0.1.0b38 I get the following exception:

```python
IntegrityError: (sqlalchemy.dialects.postgresql.asyncpg.IntegrityError) <class
'asyncpg.exceptions.UniqueViolationError'>: duplicate key value violates unique constraint "nodes_closure_pkey"
DETAIL:  Key (ancestor, descendant)=(81547, 81547) already exists.
[SQL:
            INSERT INTO nodes_closure (ancestor, descendant, depth)
            SELECT parent.id, child.id, jsonb_array_length(child.ancestors) - 2
            FROM nodes AS child
            JOIN nodes AS parent ON parent.id = child.parent
            WHERE jsonb_array_length(child.ancestors) >= 3;
        ]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
```

The node with primary id 81547 is:

```sql
catalog-alembic=# SELECT * FROM nodes WHERE id = 81547;
  id   |     key      |                              ancestors                              | structure_family | metadata | specs | access_blob |        time_created        |        time_updated
-------+--------------+---------------------------------------------------------------------+------------------+----------+-------+-------------+----------------------------+----------------------------
 81547 | ge_13element | ["7e5cd568-6a87-4995-bfcf-81903b0dcd9c", "streams", "ge_13element"] | array            | {}       | []    | {}          | 2025-07-27 00:51:25.080753 | 2025-07-27 00:51:25.080753
(1 row)
```

My one thought so far is maybe the routine for parsing the nodes to generate the closure table is not correctly interpreting a data key with the same name as it's parent stream? Though it all seems to be handled in the one SQL INSERT statement.

Here is the full output and some additional info: https://gist.github.com/canismarko/4b6244f6ee9bd95b473db20cbb88d78f

canismarko:
I found a way to change the migration so it works with my catalog. I *think* this has the same effect as originally, but not entirely sure.

[...]

PR #1116, opened 2025-08-28
Fix for nodes_closure migration when node name matches parent

canismarko:
Modifies the alembic migration that builds the initial `nodes_closure` table to handle a catalog with edge cases. See the linked issue for details.

Fixes:
- https://github.com/bluesky/tiled/issues/1114

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- Handling for certain catalog edge cases when building the nodes_closure table.
 
 ## v0.1.0-b39 (2025-08-28)
 
--- a/tiled/catalog/migrations/versions/e05e918092c3_add_closure_table.py
+++ b/tiled/catalog/migrations/versions/e05e918092c3_add_closure_table.py
@@ -82,6 +82,7 @@ The `parent` column is introduced to represent the parent-child relationships be
 The `ancestors` column is removed, as it is no longer needed with the closure table structure.
 
 """
+
 import logging
 
 import sqlalchemy as sa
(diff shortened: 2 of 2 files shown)


canismarko:
Need to update so that SQLite still works.

canismarko:
This worked on my database:

```sh
$ tiled catalog upgrade-database "postgresql+asyncpg://***:***@***/catalog"
Starting migration to add closure table...
Added 'parent' column and foreign key to 'nodes' table.
Created 'nodes_closure' table with uniqueness constraint.
Inserted root node with id=0.
Maximum depth of ancestors found: 3
Initialized parent of each node as 0 and set depth in 'nodes_closure'.
Updated 'parent' column and 'nodes_closure' for depth 1.
Updated 'parent' column and 'nodes_closure' for depth 2.
Updated 'parent' column and 'nodes_closure' for depth 3.
Completed updating 'parent' column recursively.
Inserted self-referential records into 'nodes_closure' for each node.
Updated index in the 'nodes' table.
Created unique constraint on (key, parent) pairs in 'nodes' table.
Dropped 'ancestors' column from the 'nodes' table.
Added triggers to maintain the closure table.
Migration to add closure table completed successfully.
```

genematx:
Thanks for the update, @canismarko! Just to be clear, the node that caused problems has the following path in the catalog: `7e5cd568-6a87-4995-bfcf-81903b0dcd9c/streams/ge_13element/ge_13element`, right? (i.e. an array within a container with the same name). I need to double-check if changing the order of operations would fix this, or if there's a deeper issue here.

genematx:
I think [this](https://github.com/spc-group/tiled/pull/2) should fix it, @canismarko. We don't need to change the order of migration steps, but we do need to check if the child has the same name (key) as its parent.
The self-referential rows added in Step 4 are meant to support the closure table logic; I am afraid that when we simply change the order of operations, even though the error is not being raised anymore, the resulting database is not correct with parent and child ids being mixed up.
Thank you again for noticing this and helping to fix it!

canismarko:
@genematx I restored to original DB from backup and ran the updated migration you sent. It ran without errors. ✅

I think this is ready to review and/or merge.

Issue #885, opened 2025-02-13
Swap argument order for `*partition` methods


danielballan:
Methods `write_partition(data, N)` and `append_partition(data, N)` should have the argument order swapped (`N` first) to match common usage in Python (e.g. `list.insert`) and scipy libraries. This is easy to do in a practically background-compatible way with an `isinstance(..., int)` check and a warning.

danielballan:
Closed by #1117

PR #1117, opened 2025-09-03
Changing the Order of Arguments

genematx:
The order of arguments in the `write_partition` and `append_partition` methods has been changed to put the number of partition first.

Issue: #885 

### Checklist
- [x] Add a Changelog entry


genematx:
> I think this idea is worth implementing. Easy enough, avoids user annoyance.
> 
> > This is easy to do in a practically background-compatible way with an isinstance(..., int) check and a warning.

sorry, missed that comment. Added the checks and warnings now.

PR #1121, opened 2025-09-03
Streaming (websocket) tutorial

danielballan:
This documentation PR includes some "documentation-driven development":

- Fixed missing `sequence` key in array update
- Added convenience method `x.subscribe()` for creating `Subscription` from client object
- Added a nice repr to `Subscription`
- Changed API of `Subscription` to specify where to start when calling `start()` instead of at `Subscription` instantiation time.
- Handled write-initiated close gracefully (no exception printed in the client) and exposed `sub.closed` so clients can check if the stream is over. (We might actually want a callback registry for that, so clients can subscribe to be _notified_ when the stream is over, but that's more than a tweak so I'll leave it for a future PR.)

We need "explanation" docs but I don't want to hold v0.1.0 over it.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/docs/source/tutorials/streaming.md
+++ b/docs/source/tutorials/streaming.md
@@ -85,7 +85,7 @@ where `sub` is our Subscription and `data` is a dict with information
 about what is new. In this case, it contains a detailed description of the
 new array named `x` that was created by the other client.
 
-If we are interested we can subscribe to updates about `x` and its data.
+If we are interested, we can subscribe to updates about `x` and its data.
 
 ```py
 x_sub = c['x'].subscribe()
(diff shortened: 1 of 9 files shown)


docs/source/tutorials/streaming.md
```diff
@@ -0,0 +1,174 @@
...
+the number 1. Subscribers can use this if, for example, they need to recover
+from an interruption. They can subscribe from a specific counter index `sub.start(N)`.
+As already mentioned above, `x.start(0)` means, "Start from the oldest record retained."
+
+Of course, clients can always fetch _all_ the data via the non-streaming
+interface.
+
+## Disconnecting or Closing
+
+As subscriber can disconnect from a stream at any point, like so:
```
genematx, line 135:
```suggestion
A subscriber can disconnect from a stream at any point, like so:
```

vshekar (approved):

PR #1124, opened 2025-09-04
Handle sequence counter lifecycle

### LINKED ISSUE #1123: Streaming lifecycle management
## Current Behavior

This is what happens when a streaming cache is enabled:

1. When a `POST` request creates an array or container (the only structure families we support yet, for streaming), we allocate a key in Redis named `sequence:{node_id}` with the value 0.
2. When a subsequent request writes data—or, in the case of containers, creates a _child_ container—we increment the value in `sequence:{node_id}` and note what the new value is. Then, we stash the new fragment of data (or pointer to data) in a key named `data:{node_id}:{sequence}`. That data key is assigned a TTL, so that Redis will cull it after some period (an hour, by default). Redis can cull it sooner if it is under memory pressure. Notable the `sequence:{node_id}` does not have a TTL. We do not want Redis to cull it, whether under pressure or not. Also, it's small.
3. The client may issue a `DELETE /stream/close/{path}` request to indicate that no more data is coming. If it does that, we slap a TTL on the `sequence:{node}` key so that it, too, will be culled—slightly after the oldest data that it is tracking.

## Problem

If the client does not bother to `DELETE /stream/close/{path}`, this system leaks keys. It will be a slow leak: again, the sequence counters are just one integer. But it will be a leak.

We have had some ideas for handling this more properly:
- A periodic background job, perhaps launched by the server, could cull `sequence:` keys that have no active data.
- We could assign a TTL to the `sequence:` keys from the start, but extend it each time new data is inserted for that sequence.

I think both fail to account for the use case of a long-lived subscriber to a stream that may have large gaps.

Consider, for example, some screen that shows a live feed of which beamlines are taking data. It might represent each Bluesky Run as a dot or something. That feed would experience a weeks-long (shutdown) but then want to resume where it left off. Should the `sequence` restart from 1? Would that make it harder for the client to rely on the sequence number, if it could reset to 1 mid-stream? Would it weaken the use of the sequence number for integrity checks? How hard would it be to persist the latest sequence number in "cold storage" if a stream goes on hiatus but then resumes?

## Possible Solution

As now ,`sequence:` keys in Redis should have no TTL. Redis should not be allowed to cull them ever, even when memory is tight. When a client indicates that no new data is coming (for now) with `DELETE /stream/close/{path}`, or when a periodic job detects that a sequence has not got any new updates for an extended period, the `sequence:` should be removed from Redis to clear space. However, its latest value should be then stashed in persistent storage. I think the cleanest way to do this would be a new table in a catalog database, something like:

```SQL
CREATE TABLE sequence_counters (
    FOREIGN KEY node_id REFERENCES nodes.id,
    sequence_number integer,
)
```

Crucially, the server would only need to interact with this table when closing a sequence or when _reopening_ it after some extended period of inactivity. It would not be necessary to interact with the table (and slow things down...) on each update. If the server finds a `sequence:{node_id}` key already in Redis, it knows Redis holds the ground truth. If it does not find it, only then does it consult the table to check whether there this data was once already streamed, but with its sequence counter moved to "cold storage". It would move the key back into Redis and proceed where it left off.

We could add a time of last use column if we want to cull  sequences completely (i.e. reset counter to 1) after a year or something.

We wouldn't have to use the catalog database for this. We could, for example, use another Redis. But there is something nice, I think, about treating the Redis data as largely transient and short term. I like the idea that we could blow it away and start over during a shutdown or maintenance period, and never have to worry about _migrating" anything in there. "If it's important for long-term retention, it's Postgres."

danielballan:
Stacked on top of #1121 commits. Closes #1123.

### Checklist
- [ ] Add a Changelog entry


docs/source/tutorials/streaming.md
```diff
@@ -0,0 +1,153 @@
...
+the number 1. Subscribers can use this if, for example, they need to recover
+from an interruption. They can subscribe from a specify number `sub.start(N)`.
+As already above, `x.start(0)` means, "Start from the oldest record retained."
+
+Of course, clients can always fetch _all_ the data via the non-streaming
+interface.
+
+The writer can let subscribers know that no more data is expected (for now).
+This does not prevent them from writing more later; it just signal that
+nothing more is expected. Live data data processing jobs may use this as
```
gwbischof, line 133:
```suggestion
nothing more is expected. Live data processing jobs may use this as
```

docs/source/tutorials/streaming.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+and, from there, any new updates as well.
+
+Notice that content of the updates includes a `sequence` counter, starting from
+the number 1. Subscribers can use this if, for example, they need to recover
+from an interruption. They can subscribe from a specify number `sub.start(N)`.
+As already above, `x.start(0)` means, "Start from the oldest record retained."
+
+Of course, clients can always fetch _all_ the data via the non-streaming
```
gwbischof, line 128:
```suggestion
Of course, clients can always fetch _all_ of the data via the non-streaming
```

PR #1140, opened 2025-09-05
fix: The built UI app was placed in the wrong location.

### LINKED ISSUE #1139: Web UI not available from container
When running tiled from the container, the web interface is missing

Can be seen from a browser by following the "try it" link from the root landing page, or from a terminal via:

```
$ podman run -d --rm -e TILED_SINGLE_USER_API_KEY=foobar -p8000:8000 ghcr.io/bluesky/tiled:0.1.0
a21d10f40805ee7036b8969a4d75a83b5a73221d591a3a40e629b333a859c1e0
$ curl 'http://127.0.0.1:8000/ui/browse/?api_key=foobar'
{"detail":"Not Found"}
```

I think the web-frontend directory is being copied to the wrong location in the Containerfile but not sure where it should be instead - `/app/share/`?

danielballan:
Closes #1139 

The app loads, but now I'm seeing issues with the single-user auth. It may be that this may not be operating as expected; it was recently refactored slightly as part of enabling the WebSockets work. I tested it at the time, but could have been mistaken.

https://github.com/bluesky/tiled/blob/6b47006419c1bf88cb952b9c843cc5ed7114f52a/tiled/server/authentication.py#L300-L313

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -108,7 +108,7 @@ ENV TILED_CONFIG=/deploy/config
 # Copy the pre-built `/app` directory to the runtime container
 # and change the ownership to user app and group app in one step.
 COPY --from=app_build --chown=app:app /app /app
-COPY --from=web_frontend_build --chown=app:app /src/dist /src/share/tiled/ui
+COPY --from=web_frontend_build --chown=app:app /src/dist /app/share/tiled/ui
 
 USER app
 WORKDIR /app


danielballan:
Now I understand why our builds were so slow: we were copying `.git` into the one of the build stages. I think this may have also affected the byte-compiling time.

padraic-shafer (approved):
Changes look reasonable to me.  Merge when ready.

PR #1145, opened 2025-09-17
Filter kwargs for CSV Adapter

genematx:
This allows additional kwargs in CSV adapter but filters them to ensure that `pandas.read_csv` wouldn't raise an error.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/structures/table.py
+++ b/tiled/structures/table.py
@@ -24,6 +24,7 @@ class TableStructure:
     resizable: Union[bool, Tuple[bool, ...]] = False
 
     def __post_init__(self):
+        object.__setattr__(self, "columns", list(map(str, self.columns)))
         for column in self.columns:
             if column.startswith("_"):
                 raise ValueError(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- Pass additional kwargs to csv adapter with an extra validity check.
+
 
 ## v0.1.1 (2025-09-10)
 
--- a/tiled/adapters/csv.py
+++ b/tiled/adapters/csv.py
@@ -4,6 +4,7 @@ from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Tupl
 from urllib.parse import quote_plus
 
 import dask.dataframe
+import inspect
 import pandas
 
 from ..catalog.orm import Node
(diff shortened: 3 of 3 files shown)


tiled/adapters/hdf5.py
```diff
@@ -222,20 +227,17 @@ def from_catalog(
         dataset: Optional[str] = None,
         slice: Optional[Union[str, NDSlice]] = None,
         squeeze: Optional[bool] = False,
-        swmr: bool = SWMR_DEFAULT,
-        libver: str = "latest",
         **kwargs: Optional[Any],
     ) -> "HDF5ArrayAdapter":
+        kwargs = filter_kwargs(ALLOWED_KWARGS, **{**DEFAULT_KWARGS, **kwargs})
```
danielballan: In the original, in `main`, we accept arbitrary `**kwargs` and then silently ignore them. I am not sure how/when that slipped in to the code base, but it is a known anti-pattern we avoid. User input should not be silently ignored.

This PR stops ignoring _valid_ arguments—passing them through—but it still silently ignores any invalid ones. I think in general it's better to raise an exception: we should alert the user to a misspelling or other mistake that should not pass silently.

Additionally, I have mixed feelings about proactive passing _all_  valid arguments through and exposing the full parameter set to HTTP  clients without individual review. The `h5py.File` constructor takes [a lot of parameters](https://docs.h5py.org/en/stable/high/file.html#h5py.File) and the friction of a PR to expose them individually as needed might be a good guardrail, prompting a conversation like, "Why are you trying to do that?" which might or might not lead to exposing the parameter. On the other hand, maybe we'd just be signing ourselves up for many tiny PRs. Again—mixed feelings, not sure what is best here.

danielballan (changes requested):

tiled/structures/table.py
```diff
@@ -24,6 +24,7 @@ class TableStructure:
     resizable: Union[bool, Tuple[bool, ...]] = False
 
     def __post_init__(self):
+        object.__setattr__(self, "columns", list(map(str, self.columns)))
```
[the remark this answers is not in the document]
genematx: ah, good call, thanks! this is something I copied from a frozen dataclass, but of course, it's not needed in this simple case.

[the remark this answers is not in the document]
genematx: Fixed!

tiled/adapters/hdf5.py
genematx: I agree with concerns, and likewise, I am torn between these alternatives. Ignoring kwargs is definitely wrong, but this is not a user-facing code; the invalid parameters would be often passed from `resource_kwargs` stored in the catalog, and the user, or even a Tiled admin, may not have an easy (or any) control over them. The idea to "accept all and pick needed" in this PR comes from the fact that we treat `resource_kwargs` as parameters for an adapter (I think we had a discussion about this long time ago, but I might be mistaken).

I would prefer a set of arguments clearly defined in the signature, like Tom has mentioned below -- I've always found kwargs too hard parse when reading docs, but this does set us up for a continuous updates like we're doing already here. Specifically, HXN need to set `locking=False` when they read HDF5 files (why -- is a different story); tbh, I've never used it myself before and thought that simply setting `swmr=True` would take care of this. Even worse, `pandas.read_csv` has a lot of potentially useful parameters that we may need to rely on when reading esoteric CSV formats.

On the other hand, I don't like allowing ALL (possibly unneccessary) kwargs from `hdf5.File` and `pandas.read_csv`. A potential workaround could be to go through the list of all possible kwargs and pick those that we _think_ would be useful?

danielballan: That is helpful context.

Starting with a curated set seems reasonable. That would address @tacaswell’s comment, as we would avoid exposing everything. But we would also give ourselves the tools we may need to deal with what we find around the ring.

PR #1146, opened 2025-09-17
ENH: configure recycle and pre_ping for storage ADBC connections

genematx:
Fine-tuning DB Connection Pooling:
1. Allow recycling in Storage connection Pool;
2. Explicitly configure `pre_ping` (False by default) in Storage connection pool;
3. ...

### Checklist
- [ ] Add a Changelog entry


danielballan:
Per discussion on Slack, we want to observe the pool monitoring #1163 and learn more about whether this is necessary and appropriate for our use case.

Issue #1144, opened 2025-09-17
Investigate change in duckdb


danielballan:
DuckDB issued a [release](https://github.com/duckdb/duckdb/releases) last night, which is likely the cause of the wall of these errors on last night's scheduled [CI run](https://github.com/bluesky/tiled/actions/runs/17786589157/job/50555319933).

```
FAILED tiled/_tests/test_writing.py::test_append_partition[orig_file2-file_to_append2-expected_file2] - ModuleNotFoundError: No module named 'adbc_driver_duckdb'
```

danielballan:
Confirmed with a [simple example](https://gist.github.com/danielballan/11db4842a8a73213019866be21c35970) that this is a regression in DuckDB. Reported [upstream](https://github.com/duckdb/duckdb/issues/19038).

danielballan:
Thanks @evertlammerts! It was generous of you to follow up here. Appreciated.

PR #1149, opened 2025-09-17
Pin back duckdb to avoid regression in 1.4.0

danielballan:
Works around #1144 as a temporary fix while we wait for word (and hopefully a fix) from upstream.

### Checklist
- [x] Add a Changelog entry


danielballan:
Self-merging as a hot fix.

PR #1152, opened 2025-09-19
Subscription.start() blocks the current thread.

### LINKED ISSUE #1151: Make `Subscription.start()` blocking
Consider this example:

```py
import time

def on_new_array(sub, data):
    print(data)

client = from_uri(...)
sub = client.subscribe()
sub.add_callback(on_new_array)
print("Listening for updates. Use Ctrl+C to stop....")
sub.start()  # launch background thread, listening for updates

# Block with a sleep loop.
try:
    while True:
        time.sleep(1)
except KeyboardInterrupt:
    pass
```

Feedback from @tacaswell and @mrakitin: It's surprising that `sub.start()` does not block. A user might run the script without the sleep loop and be surprised that the script immediately exits. (To be honest, I did this exactly thing yesterday, even as the author of the API!) As a comparison, the pyzmq API blocks on `start()` as does Bluesky's `RemoteDispatcher`.

Perhaps the existing behavior should be renamed `start_on_thread()` or something, and `start()` should block on the current thread.

danielballan:
Closes #1151. Needs updates to unit tests.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/client/stream.py
+++ b/tiled/client/stream.py
@@ -21,6 +21,13 @@ Callback = Callable[["Subscription", dict], None]
 "A Callback will be called with the Subscription calling it and a dict with the update."
 
 
+API_KEY_LIFETIME = 30  # seconds
+RECEIVE_TIMEOUT = 0.1  # seconds
+
+
+__all__ = ["Subscription"]
+
+
 class _TestClientWebsocketWrapper:
     """Wrapper for TestClient websockets."""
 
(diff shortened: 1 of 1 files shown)


tiled/client/stream.py
```diff
@@ -200,8 +199,33 @@ def remove_callback(self, callback: Callback) -> Self:
...
+        if needs_api_key:
+            # The connection is made, so we no longer need the API key.
+            # TODO: Implement single-use API keys so that revoking is not
+            # necessary.
+            self.context.revoke_api_key(key_info["first_eight"])
+
     def _receive(self) -> None:
-        "This method is executed on self._thread."
+        "Blocking loop that receives and processes updates"
         TIMEOUT = 0.1  # seconds
```
genematx: Is it worth pulling this variable to the global config, or at leat to the top of the module?

genematx (approved):

PR #1160, opened 2025-09-24
Accept 'trees' and 'media_types' config in multiple files.

### LINKED ISSUE #1148: Address regressions in configuration parsing
Two knows issues:
1. If multiple files define `trees:` they were formally merged. Now, an error is raised.
2. The `databroker.mongo_normalized` config does not seem parseable. ~~Needs investigation, but I suspect it's `package.module.object` versus `package.module:object` syntax.~~

Reproducer for 2:

```
# config.yml
trees:
  - tree: databroker.mongo_normalized:Tree.from_uri
    path: /
    args:
      uri: "mongodb://localhost:27017/test"
```

```
❯ uv run --with "databroker[all] ==2.0.0b68" tiled admin check-config config.yml
Installed 131 packages in 131ms
1 validation error for Config
trees.0.tree
  Invalid python path: cannot import name 'Tree.from_uri' from 'databroker.mongo_normalized' [type=import_error, input_value='databroker.mongo_normalized:Tree.from_uri', input_type=str]
```

danielballan:
Closes #1148

Testing the NSLS-II tiled configuration (as I should have done before merging #1091)...

On `main`:

```
❯ uv run tiled admin check-config ../tiled-site-config
Duplicate configuration for {'trees', 'specs'} in ../tiled-site-config/bluesky_prod_mongos.yml
```

On this PR branch:

```
❯ uv run tiled admin check-config ../tiled-site-config
No errors found in configuration.
```

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ Write the date in place of the "Unreleased" in the case a new version is release
 # Changelog
 
 
+## Unreleased
+
+### Fixed
+
+- A regression in v0.1.1 disallowed specifying `trees` and `media_types`
+  across multiple files. (This can be useful for config.d-style configuration
+  where different config files are managed by different stages of configuration
+  management.)
+
+
 ## v0.1.2 (2025-09-17)
 
 This release temporarily pins backs the version of the depedency DuckDB to
--- a/tiled/_tests/test_config.py
+++ b/tiled/_tests/test_config.py
@@ -114,16 +114,36 @@ def test_extra_files(tmpdir):
     parse_configs(str(tmpdir))
 
 
-def test_multi_file_conflict(tmpdir):
+def test_multi_file_trees(tmpdir):
+    "Test that 'trees' can be specified across more than one file, merged."
     import yaml
 
-    conf1 = {"trees": [{"path": "/", "tree": "tiled.examples.generated_minimal:tree"}]}
-    conf2 = {"trees": [{"path": "/", "tree": "tiled.examples.generated_minimal:tree"}]}
+    conf1 = {"trees": [{"path": "/a", "tree": "tiled.examples.generated_minimal:tree"}]}
+    conf2 = {"trees": [{"path": "/b", "tree": "tiled.examples.generated_minimal:tree"}]}
     with open(tmpdir / "conf1.yml", "w") as c1:
         yaml.dump(conf1, c1)
     with open(tmpdir / "conf2.yml", "w") as c2:
         yaml.dump(conf2, c2)
-    with pytest.raises(ValueError, match="Duplicate configuration for {'trees'}"):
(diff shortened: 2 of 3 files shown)


tiled/_tests/test_config.py
```diff
@@ -114,16 +114,36 @@ def test_extra_files(tmpdir):
...
+    "Test that 'trees' can be specified across more than one file, merged."
+    import yaml
+
+    conf1 = {"trees": [{"path": "/a", "tree": "tiled.examples.generated_minimal:tree"}]}
+    conf2 = {"trees": [{"path": "/b", "tree": "tiled.examples.generated_minimal:tree"}]}
+    with open(tmpdir / "conf1.yml", "w") as c1:
+        yaml.dump(conf1, c1)
+    with open(tmpdir / "conf2.yml", "w") as c2:
+        yaml.dump(conf2, c2)
+    parse_configs(tmpdir)
```
tpoliaw, line 127: Should this test that both trees are present?

tiled/config.py
```diff
@@ -304,7 +304,16 @@ def parse_configs(src_file: Union[str, Path]) -> Config:
             if f.is_file() and f.suffix == ".yml":
                 new_config = parse(f)
                 if common := new_config.keys() & conf.keys():
-                    raise ValueError(f"Duplicate configuration for {common} in {f}")
+                    # These specific keys can be merged from separate files.
+                    # This can be useful for config.d-style where different
+                    # files are managed by different stages of configuration
+                    # management.
+                    mergeable_lists = {"trees", "specs"}
```
tpoliaw, line 311: From the old config merge function (ln 247-251), it looks like `media_types`, `file_extensions` and `allow_origins` were also merged instead of being exclusive. Should they be allowed here as well? It might be awkward if they need merging in different ways, extend vs update etc

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ Write the date in place of the "Unreleased" in the case a new version is release
 # Changelog
 
 
+## Unreleased
+
+### Fixed
+
+- A regression in v0.1.1 disallowed specifying `trees` and `media_types`
```
tpoliaw, line 11: If the code is left to just `trees` and `specs`, this should be updated to match

tpoliaw (commented):
Thanks for catching and sorting this - I saw the old merge function excluding duplicates and got carried away when I re-added the multi-file loading.

tiled/config.py
danielballan: I agree: extending this to objects (`media_types` and `file_extensions`) makes things more complex. Let's leave these out. (NSLS-II does not rely on them.) If someone yells, we can revisit and re-instate if needed.

genematx (approved):

PR #1161, opened 2025-09-24
Bug: De-dupe profile editor artifacts

### LINKED ISSUE #1157: Tiled Profile should ignore temporary files
After editing my profile with a certain widely-used, open-source text editor, I often get a file that ends in ~, containing autosave information. Then, this results in the following Tiled error:

```bash
tiled profile get-default
/home/jamie/work/xraygui/pymca-pixi-dev/.pixi/envs/default/lib/python3.11/site-packages/tiled/profiles.py:203: UserWarning: More than file in the same directory:

/home/jamie/.config/tiled/profiles/ucal.yml
/home/jamie/.config/tiled/profiles/ucal.yml~

defines a profile with the name 'ucal'.

The profile will be ommitted. Fix this by removing one of the duplicates.
```

It would be nice if Tiled would ignore files like
```
ucal.yml~
#ucal.yml#
ucal.yml.bak
ucal.bak
```
by just doing some basic pattern matching for files that end in "yml" or "yaml", and ignoring the rest. Or explicitly discarding the stuff like ~ and # that you would normally put in your .gitignore, if the desire is really for any other file with any ending to be parsed as a profile.

Kezzsim:
Resolves #1157 by blocking file suffixes identified as editor artifacts. I didn't think to go so far as exclude certain valid alphanumeric filetypes such as `.bak` that might benefit from the creation of an explicit file extension blacklist.

### DIFF
--- a/tiled/profiles.py
+++ b/tiled/profiles.py
@@ -85,10 +85,10 @@ def gather_profiles(paths, strict=True):
             for filename in path.iterdir():
                 filepath = path / filename
                 # Only parse files in the YAML format.
-                if (
-                    filename.name.startswith(".")
-                    or filename.suffix not in {".yml", ".yaml"}
-                ):
+                if filename.name.startswith(".") or filename.suffix not in {
+                    ".yml",
+                    ".yaml",
+                }:
                     continue
                 try:
                     content = parse(filepath)


danielballan:
Another possibility is to do this the other way around.

In this PR we are addressing _client_ config (which tiled calls "profiles"). As a point of comparison, in the _server_ config, we just ignore everything that is not `.yml`:

https://github.com/bluesky/tiled/blob/e0277dcd7bfe5dc04034d367658208276eead162/tiled/config.py#L305

Maybe that's worth considering here. It's pretty simple to explain/understand! We could perhaps relax it to accept also `.yaml`.

What do you think?

Kezzsim:
I agree, it should be a standard that's used both client and server side. I've kept the original convention of walking the entire directory and bypassing invalid files with a `continue` to stay consistent.

danielballan (approved):

PR #1164, opened 2025-09-24
Add 'locking' parameter to hdf5 adapter

genematx:
This adds `locking` key-word parameter to HDF5Adapter and HDF5ArrayAdapter and removes the (otherwise ignored) `**kwargs`.

Related: #1145 

### Checklist
- [x] Add a Changelog entry


danielballan (approved):
Looks good, just need CHANGELOG update

PR #1170, opened 2025-09-26
Add opa

ZohebShaikh:
### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/example_configs/keycloak_oidc/policy/rbac/rbac.rego
+++ b/example_configs/keycloak_oidc/policy/rbac/rbac.rego
@@ -1,0 +1,37 @@
+package rbac
+
+import rego.v1
+
+import data.token
+
+# role-permissions assignments
+role_permissions := {
+	"user": [
+		{"action": "read:metadata"},
+		{"action": "read:data"},
+		{"action": "write:metadata"},
+		{"action": "write:data"},
+	],
+	"admin": [
+		{"action": "read:metadata"},
+		{"action": "read:data"},
+		{"action": "write:metadata"},
+		{"action": "write:data"},
+		{"action": "create"},
+		{"action": "register"},
+	],
+	"service": [{"action": "metrics"}],
+}
+
+default allow := false
+
+# METADATA
+# description: Check permission for each user
+# entrypoint: true
+allow if {
+	every action in input.actions {
+		some r in token.roles
+		some p in role_permissions[r]
+		p == {"action": action}
+	}
+}
--- a/example_configs/keycloak_oidc/policy/rbac/rbac.rego
+++ b/example_configs/keycloak_oidc/policy/rbac/rbac.rego
@@ -12,14 +12,6 @@ role_permissions := {
 		{"action": "write:metadata"},
 		{"action": "write:data"},
 	],
-	"admin": [
-		{"action": "read:metadata"},
-		{"action": "read:data"},
-		{"action": "write:metadata"},
-		{"action": "write:data"},
-		{"action": "create"},
-		{"action": "register"},
-	],
 	"service": [{"action": "metrics"}],
 }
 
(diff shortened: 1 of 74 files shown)


tiled/access_control/access_policies.py
```diff
@@ -412,3 +415,178 @@ async def filters(
...
+        self._scopes_access = urljoin(
+            str(self.authorization_provider),
+            urlsplit(scopes_access).path,
+        )
+
+        self.audience = audience
+        self.provider = provider
+
+    def _identifier(self, principal) -> str:
+        if principal.type == PrincipalType.service:
```
DiamondJoseph: (not knowing enough about AccessPolicy)- should we be returning anything if it's not an external Principal? As an authorization provider, we can't verify that the principal is authenticated if it's not a token

tiled/access_control/access_policies.py
```diff
@@ -412,3 +415,178 @@ async def filters(
...
+            )
+        )
+        async with httpx.AsyncClient() as client:
+            response = await client.post(
+                str(self._node_access), content=input.model_dump_json(exclude_none=True)
+            )
+        response.raise_for_status()
+        decision = Decision.model_validate_json(response.text)
+        if not decision.result:
+            raise ValueError("Permission denied not able to add the node")
```
DiamondJoseph:
```suggestion
        if not decision.result:
            return (False, None)
```

?

tiled/access_control/access_policies.py
```diff
@@ -412,3 +415,178 @@ async def filters(
...
+            str(self.authorization_provider),
+            urlsplit(filter_nodes).path,
+        )
+        self._scopes_access = urljoin(
+            str(self.authorization_provider),
+            urlsplit(scopes_access).path,
+        )
+
+        self.audience = audience
+        self.provider = provider
```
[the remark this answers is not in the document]
ZohebShaikh: I don't think having audience and providers as a property is required

example_configs/keycloak_oidc/policy/rbac/rbac.rego
```diff
@@ -0,0 +1,98 @@
...
+		"delete:node",
+		"delete:revision"
+	],
+	"beamline_x_user": [
+		"read:data",
+		"read:metadata",
+	],
+	"public": [
+		"read:data",
+		"write:data",
```
danielballan: This tag configures what unauthenticated requests can do. Generally I would expect those would never be allowed to write (or create, register, delete).

tiled/access_control/access_policies.py
```diff
@@ -412,3 +414,167 @@ async def filters(
...
+
+class Input(BaseModel):
+    input: Data
+
+
+class Decision(BaseModel):
+    result: Union[List[str], bool]
+
+
+class ExternalPolicyDecisionPoint(AccessPolicy):
```
danielballan: @nmaytan and I wonder if the `TagBasedAccessPolicy` could be reused, using the `tag_parser` argument to inject OPA-specific integration (`_get_external_decision`). Most of the `AccessPolicy` interface will be the same across our local solution, OPA, OpenFGA, and others. The `TagParser` abstraction might be a more tightly-scoped way of injecting framework-specific integration.

Nate will aim to find the bandwidth to implement this suggestion as a PR into your PR. But I mention this now in case you have any immediate thoughts on this.

danielballan:
Plan:
1. @ZohebShaikh will remove `dls.py` which can live in local config (put on the `PYTHONPATH`).
2. @danielballan will merge and release a beta by end of US workday on Wednesday.
3. :christmas_tree: 
4. Evaluate whether we can consolidate on one plugin interface (i.e. `build_input` + `_get_external_decision` versus `tag_parser`). This should enable us to share implementations of `init_node`, `modify_node`, and `filter`.
5. Decide, item by item, how much of the authorization logic in the `TagParser` would be better delegated to an external system: OPA or NSLS-II own's homegrown (temporary) authorization system.
6. Consider whether custom integrations with authorization systems should plug in to Tiled by implementing a custom AccessPolicy class, or whether they should all use the same Tiled AccessPolicy and inject customizations some other way (e.g tag_parser).

danielballan (approved):

Issue #1128, opened 2025-09-04
Stream updates to containers


danielballan:
If a child of a container is updated or delete, publish an update over WS.

This is useful in general, but has particular utility for Bluesky. When TiledWriter receives a RunStart document, it crates a new node with a `POST /metadata/{beamline}/raw/{uid}` request. WS sends out a message to subscribers of the parent container node.

When TiledWriter receives a RunStop document, it updates to the node's metadata with `PATCH /metadata/{beamline}/raw/{uid}` request, adding a `"stop"` key beside the original `"start"` key. Currently, Tiled does not send any notification out.

Once it does send a notification, we can use that notification to kick off end-of-run workflows.

_Note that no Bluesky specifics are needed in this PR; Tiled can treat this as generically, "Metadata has changed; send an update!"_

PR #1176, opened 2025-10-01
🛜 Stream container updates to WebSocket subscribers

Kezzsim:
Addresses #1128 

Previously, when creating a new node, connected websocket clients were able to receive an update json object via redis which contained information about that new node. This PR adds that functionality to updating metadata on a node, transmitting the metadata object to subscribers so they can react accordingly.

### Checklist
- [ ] Add a Changelog entry
- [ ] Add Testing (either in `_tests/test_websockets.py` or in the metadata update test)

### DIFF
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -1097,6 +1097,31 @@ class CatalogNodeAdapter:
                 update(orm.Node).where(orm.Node.id == self.node.id).values(**values)
             )
             await db.commit()
+            # Upon successful update, inform websocket subscribers through redis
+            if self.context.cache_client:
+                sequence = await self.context.cache_client.incr(
+                    f"sequence:{self.node.id}"
+                )
+                metadata = {
+                    "sequence": sequence,
+                    "timestamp": datetime.now().isoformat(),
+                    "specs": [spec.model_dump() for spec in (specs or [])],
+                    "metadata": metadata,
+                    "revision_number": next_revision_number,
+                }
+                pipeline = self.context.cache_client.pipeline()
+                pipeline.hset(
+                    f"data:{self.node.id}:{sequence}",
+                    mapping={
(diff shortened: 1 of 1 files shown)


tiled/catalog/adapter.py
```diff
@@ -1097,6 +1097,31 @@ async def replace_metadata(
                 update(orm.Node).where(orm.Node.id == self.node.id).values(**values)
             )
             await db.commit()
+            # Upon successful update, inform websocket subscribers through redis
+            if self.context.cache_client:
+                sequence = await self.context.cache_client.incr(
+                    f"sequence:{self.node.id}"
```
danielballan, line 1103: If we go with the design of notifying the _parent_ node, every instance of `node.id` in here should be `node.parent`.

I think that would be a good design to start with; otherwise metadata updates and _data_ updates will be mixed and that could force more complexity on the consumers. We can revisit later on and consider notifying _both_. But notifying the parent feels like a safe design initially.

tiled/catalog/adapter.py
```diff
@@ -1097,6 +1097,31 @@ async def replace_metadata(
...
+                )
+                metadata = {
+                    "sequence": sequence,
+                    "timestamp": datetime.now().isoformat(),
+                    "specs": [spec.model_dump() for spec in (specs or [])],
+                    "metadata": metadata,
+                    "revision_number": next_revision_number,
```
danielballan, line 1110: Nice! This is good idea to include.:-)

danielballan (dismissed):
Awesome! Nice job extending the patterns. Just a couple notes.

tiled/catalog/adapter.py
Kezzsim: Addressed in 5bd8510 , seems like the most straightforward option

danielballan:
Works interactively!

PR #1178, opened 2025-10-07
Implement /authorize endpoint to initiate OAuth flows with redirects

SeijDeLeon:
Changes to support a web browser client to authenticate with external provider.

- Added a new GET /provider/{provider}/authorize endpoint which redirects to the login page of the external provider with the necessary client_id parameters. This is the same "auth_endpoint" link provided by Tiled in the /api/v1/ route, so it is discoverable without any other information by a client.
- Enabled optional failure and success redirect url parameters for authentication.

### 1. Discover Available Authentication Providers

```http
GET /api/v1/
```

Response includes available authentication providers:
```json
{
  "authentication": {
    "providers": [
      {
        "provider": "orcid",
        "mode": "external",
        "links": {
          "auth_endpoint": "/api/v1/auth/provider/orcid/authorize"
        }
      }
    ]
  }
}
```

### 2. Initiate OAuth Flow

Redirect the user's browser to the auth endpoint, add optional state param that will be included at the final redirect:

```http
GET /api/v1/auth/provider/orcid/authorize?state=abc123
```

### 3. OAuth Flow Execution

Tiled performs authentication procedure

### 4. Final Redirect

Based on your Tiled configuration:

**On Success** (`redirect_on_success` configured):
```http
HTTP 302 Redirect to: http://successURL/?access_token=...&refresh_token=...&identity.id=...&identity.provider=orcid&principal=...&state=abc123
```

**On Failure** (`redirect_on_failure` configured):
```http  
HTTP 302 Redirect to: http://failureURL/
```

**No redirect configured**:
Returns JSON response with tokens or error details (same as before).

## Configuration Requirements

Two new args have been added

```yaml
authentication:
  providers:
  - provider: orcid
    authenticator: tiled.authenticators:OIDCAuthenticator
    args:
      # ... other config ...
      redirect_on_success: http://successURL/ # new arg
      redirect_on_failure: http://failureURL/ # new arg
```

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/_tests/test_authenticators.py
+++ b/tiled/_tests/test_authenticators.py
@@ -9,6 +9,7 @@ from cryptography.hazmat.primitives.asymmetric import rsa
 from jose import ExpiredSignatureError, jwt
 from jose.backends import RSAKey
 from respx import MockRouter
+from starlette.datastructures import URL, QueryParams
 
 from ..authenticators import (
     LDAPAuthenticator,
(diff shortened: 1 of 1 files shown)


dylanmcreynolds:
Talking to @danielballan, we think a great addition to this PR would be that have update the root `metadata` endpoint that gives comprehensive authentication provider information to the client to include the .well-known URL for OICD providers. Then, the client can use the `authorization_endpoint` to redirect the user initially.

tiled/server/authentication.py
```diff
@@ -707,7 +713,41 @@ async def auth_code_route(
                 user_session_state.state,
             )
             tokens = await create_tokens_from_session(settings, db, session, provider)
-        return tokens
+            if authenticator.redirect_on_success:
+                redirect_url = (
```
danielballan, line 717: Using a [`starlette.URL`](https://www.starlette.dev/requests/#url) would be safer here (avoids mistakes in string munging, and properly encodes inputs).

tiled/_tests/test_authenticators.py
```diff
@@ -2,8 +2,10 @@
 import os
 
 import pytest
+from ..authenticators import LDAPAuthenticator, OIDCAuthenticator
+from unittest.mock import AsyncMock, Mock, patch
```
danielballan, line 6: We generally avoid `Mock` or at least adhere to the maxim, "Only mock your own objects." I think it is not used anywhere in the tiled codebase or any of the core bluesky projects.

Would you be open to trying [respx](https://pypi.org/project/respx/) which @ZohebShaikh also introduced in the open PR #1054? The mocking there is more targeted.

tiled/server/authentication.py
SeijDeLeon: Revised to utilize starlette URL

```python
#lines 740-753
if authenticator.redirect_on_success:
                params = {
                    "access_token": tokens["access_token"],
                    "refresh_token": tokens["refresh_token"],
                    "identity.id": tokens["identity"]["id"],
                    "identity.provider": tokens["identity"]["provider"],
                    "principal": tokens["principal"],
                }
                if "state" in request.query_params:
                    params["state"] = request.query_params["state"]
                redirect_url = URL(
                    authenticator.redirect_on_success
                ).include_query_params(**params)
                return RedirectResponse(status_code=302, url=str(redirect_url))
```

tiled/_tests/test_authenticators.py
SeijDeLeon: I was able to take out unittest.mock and use respx, along with some of the fixtures from PR #1054, please take a look and let me know if that's what you were thinking of

danielballan (approved):

PR #1188, opened 2025-10-20
Apply default cache_settings when using CLI

danielballan:
This is a follow-up on #1175, ensuring that the defaults are applied when the configuration is passed via the CLI (as opposed to the config file).

Since this fixes a regression that has not yet been released, we can omit it from the CHANGELOG.

### Checklist
- [ ] Add a Changelog entry


danielballan:
There is already an open issue regarding improving CLI test coverage. It's "expensive" because it requires launching a subprocess that starts a server.

Issue #1130, opened 2025-09-04
Encapsulate Redis integration into an object


danielballan:
This object, maybe called `StreamingCache`, should take `streaming_cache` configuration as input. We should aim to make its public API "backend-agnostic", in the sense that implementations could be written for databases other than Redis (e.g. NATS).

danielballan:
Closed by #1192

PR #1192, opened 2025-10-21
Abstract redis specific code into a StreamingCache class

vshekar:
Extensible to allow other datastores like NATS
Addresses issue #1130

### Checklist
- [x] Add a Changelog entry


danielballan:
Rebased.

Issue #1155, opened 2025-09-22
Move datasource patch description from `PUT` body to query parameters


danielballan:
_Capturing a discussion from #1115_

Where a writer wants to tell Tiled that some external data has been changed—e.g. that a detector wrote additional frames to an array on disk—it should tell us _which_ data changed, i.e. the coordinates of the new/changed data.

Currently, it does this with a `patch` field in the body of the PUT request:

https://github.com/bluesky/tiled/blob/67a37f899e8c6ef4a65d6f20c2ee312b7377b3cf/tiled/server/schemas.py#L465-L467

But, as @dylanmcreynolds pointed out, this goes against the semantics of PUT. The body of a PUT request really ought to _be_ the replacement resource, not the replacement resource plus some other stuff.

We propose to move this information from the body of the request to query parameters. This ought to be doable in a backward-compatible way.

PR #1201, opened 2025-10-26
Stream tabular data

danielballan:
This partially addresses https://github.com/bluesky/tiled/issues/1125. Uploaded tables are supported, _registered_ external tabular data is not. But the first one is the only kind used by Bluesky, so this alone is great progress.

**This is based on the commits in #1200, with relatively small additions specific to this PR.** I also piled on a small fix to streaming external arrays, fixing bugs discovered while doing end-to-end tests with Bluesky.

## End-to-end demo streaming Bluesky data via WS

1. Run Redis

```sh
podman run -d --rm --name tiled-test-redis -p 6379:6379 docker.io/redis:7-alpine
```

2. Run a Tiled server with permission to serve files from `/tmp` (where `ophyd.sim.img` wheres its external data).

```sh
uv run --all-extras tiled serve catalog --temp --api-key secret --cache redis://localhost:6379 -r /tmp
```

3. Launch a WS consumer.

```py
# bluesky_test_consumer.py
from tiled.client import from_uri

client = from_uri('http://localhost:8000', api_key='secret')

def on_new_child(update):
    # TODO: Teach TiledWriter to call close_stream so we can used that to trigger cleanup.
    child = update.child()
    print(child)
    sub = child.subscribe()
    if child.structure_family == "container":
        # Recursively subscribe to updates.
        sub.child_created.add_callback(on_new_child)
    else:
        sub.new_data.add_callback(on_new_data)
    sub.start_in_thread(1)


def on_new_data(update):
    print(update.data())


catalog_sub = client.subscribe()
catalog_sub.child_created.add_callback(on_new_child)
catalog_sub.start()
```

4. Launch demo data acquisition.

```py
# bluesky_test_producer.py
from bluesky import RunEngine
from bluesky.plans import scan
from bluesky.callbacks.tiled_writer import TiledWriter
from ophyd.sim import det, motor, img
from tiled.client import from_uri

client = from_uri('http://localhost:8000', api_key='secret')
tw = TiledWriter(client, batch_size=1)
RE = RunEngine()
RE.subscribe(tw)

RE(scan([det, img], motor, -1, 1, 3))
```

The consumer outputs the following. There is something funny going on with arrays here—I suspect it's in the details of how TiledWriter deals with the output of `ophyd.sim.det`—but the data comes out in the end.

```
<BlueskyRun v3.0 streams: {'primary'} scan_id=1 uid='c564adb8' 2025-10-26 15:56>
<Container {'primary'}>
[...]

PR #1203, opened 2025-10-28
Move array patch specification from `PUT` body to query parameters

danielballan:
Closes https://github.com/bluesky/tiled/issues/1155.

Piles on a little cleanup (CHANGELOG, docs, minor tweaks) from #1200 and #1201.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/server/router.py
+++ b/tiled/server/router.py
@@ -1641,7 +1641,6 @@ def get_router(
         authn_scopes: Scopes = Depends(get_current_scopes),
         patch_shape: Optional[tuple[int, ...]] = Depends(patch_shape_param),
         patch_offset: Optional[tuple[int, ...]] = Depends(patch_offset_param),
-        patch_extend: Optional[bool] = None,
         _=Security(check_scopes, scopes=["write:metadata", "register"]),
     ):
         entry = await get_entry(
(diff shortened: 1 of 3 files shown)


tiled/server/router.py
```diff
@@ -1636,6 +1639,9 @@ async def put_data_source(
         session_state: dict = Depends(get_session_state),
         authn_access_tags: Optional[AccessTags] = Depends(get_current_access_tags),
         authn_scopes: Scopes = Depends(get_current_scopes),
+        patch_shape: Optional[tuple[int, ...]] = Depends(patch_shape_param),
+        patch_offset: Optional[tuple[int, ...]] = Depends(patch_offset_param),
+        patch_extend: Optional[bool] = None,
```
genematx: I wonder if `patch_extend` (and `extend` in general) is superfluous. Shouldn't we always assume that if the size of the patch + offset goes beyond the current array shape, the user wants to extend it? Is the purpose of this parameter just to catch potential errors with mishaped updates? (I think a far worse error in this case would be someone forgetting to specify (or incorrectly specifying) the `offset` and overwriting the top left corner of the array inadvertently).

PR #1206, opened 2025-10-29
Move util shared by client and adapters out of adapters subpackage.

danielballan:
This superses #1191, fixing the issue by avoiding a sqlalchemy dependency in the client.

This is a backward-compatible minor reorganization to fix an unreleased regression, so I'll leave it off the CHANGELOG.


tpoliaw:
Much better approach than adding the dependency 👍

PR #1213, opened 2025-10-30
Add optional 'persist' query to PUT/PATCH array

### LINKED ISSUE #1153: Support transient data that is streamed but not persisted
If the Tiled server is configured with a Redis cache, it always both (5) streams data to any subscribers and then (6) persists it into storage at rest.

```mermaid
%%{init: {'sequence': {'mirrorActors': false}}}%%

sequenceDiagram
	participant ReadingClient as Reading Client
	participant WritingClient as Writing Client
	participant TiledServer as Tiled Servers
	participant Redis
	participant ManagedStorage as Managed Storage
	
	ReadingClient->>TiledServer: 1. Subscribe over a WebSocket connection
	WritingClient->>TiledServer: 2. HTTP PUT or PATCH data
        activate TiledServer
	TiledServer->>Redis: 3. Cache data in Redis with 1-hour TTL.
	Redis->>TiledServer: 4. Notify all Tiled servers about new data
	TiledServer->>ReadingClient: 5. Transmit new data over WebSocket
	TiledServer->>ManagedStorage: 6. Write to persistent storage
        deactivate TiledServer
```

Sometimes data is only of fleeting value, like a partial reconstruction in a series of ever-improving partial results. In that case, it may not be worth persisting each incremental update, but on the final one. Perhaps the Writing Client should have the option to opt out of (6).

This could be exposed as a query parameter on the `PUT /array/...` and `PATCH /array/...` routes, `?transient=true`.

padraic-shafer:
Resolves #1153.

Adds an optional `persist` query parameter to PUT /array/... and PATCH /array/... routes.

When `True` (default) the data is written as usual; otherwise the data is only streamed and not persisted.

This option was also added to the write(), write_block(), and patch() methods on the DaskArrayClient.

Added new tests for PUT /array/full and PATCH /array/full.

### DIFF
--- a/tiled/catalog/adapter.py
+++ b/tiled/catalog/adapter.py
@@ -1235,13 +1235,6 @@ class CatalogArrayAdapter(CatalogNodeAdapter):
     async def patch(
         self, shape, offset, extend, media_type, deserializer, entry, body, persist=True
     ):
-        if extend and not persist:
-            raise Conflicts(
-                "Cannot PATCH an array with both parameters"
-                " extend=True and persist=False."
-                " To extend the array, you must persist the changes."
-                " To skip persisting the changes, you must not extend the array."
-            )
         if self.context.streaming_cache:
             await self._stream(media_type, entry, body, shape, offset=offset)
         if not persist:
--- a/tiled/client/array.py
+++ b/tiled/client/array.py
@@ -300,10 +300,13 @@ class _DaskArrayClient(BaseClient):
                     headers={"Content-Type": "application/octet-stream"},
                     params=params,
                 )
-                if response.status_code == httpx.codes.CONFLICT:
+                if response.status_code in [
+                    httpx.codes.BAD_REQUEST,
+                    httpx.codes.CONFLICT,
+                ]:
                     raise ValueError(
                         response.json()
-                        .get("detail", "Conflict error.")
+                        .get("detail", "Array parameters conflict.")
                         .replace(
                             "Use ?",  # URL query param
                             "Pass keyword argument ",  # Python function argument
--- a/tiled/server/router.py
+++ b/tiled/server/router.py
@@ -1813,6 +1813,17 @@ def get_router(
         authn_scopes: Scopes = Depends(get_current_scopes),
         _=Security(check_scopes, scopes=["write:data"]),
     ):
+        if extend and not persist:
+            bad_args_message = (
+                "Cannot PATCH an array with both parameters"
+                " extend=True and persist=False."
+                " To extend the array, you must persist the changes."
+                " To skip persisting the changes, you must not extend the array."
+            )
+            raise HTTPException(
+                status_code=HTTP_400_BAD_REQUEST,
+                detail=bad_args_message,
+            )
         entry = await get_entry(
             path,
             ["write:data"],


tiled/client/array.py
```diff
@@ -177,22 +177,24 @@ def read(self, slice=None):
...
-    def write(self, array):
+    def write(self, array, persist=True):
         for attempt in retry_context():
             with attempt:
                 handle_error(
                     self.context.http_client.put(
                         self.item["links"]["full"],
                         content=array.tobytes(),
                         headers={"Content-Type": "application/octet-stream"},
+                        params={"persist": bool(persist)},
```
danielballan: To keep the server logs as readable as possible, we have elsewhere adopted a pattern of only specifying optional params when they are not the default value, to cut down on clutter. I think it makes sense to apply that pattern here too (and in the other writing calls below) because the default value will very frequently be used.

danielballan:
My previous comment does prompt an interesting design question: what to do with `tiled.client.container:Container.write_array`. That is what most users interact with today, and it's convenience method around:

```py
x = Container.new(...)  # POST
x.write(arr)  # PUT
# (or x.write_block if chunked)
```

Above I had in mind that `persist` would b exposed on `write_array` and passed through. But I wonder if this could be too confusing. The POST will _always_ persist. (Without a POST, there's no id to put in Redis, and no schema to provide to subscribers.) So `c.write_array(..., persist=False)` would mean, "Create a node around this data, and broadcast these initial values, but leave the persist data "empty". That may be too non-intutitive.

Maybe it's better to guide users who want to use this feature to call `new(...)` directly and then, separately, `write` or `write_block` with `persist=False`.

tiled/client/array.py
```diff
@@ -177,22 +177,32 @@ def read(self, slice=None):
             dask_array = dask_array[slice]
         return dask_array
 
-    def write(self, array):
+    def write(self, array, persist=None):
```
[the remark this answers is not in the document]
padraic-shafer: Resolved in

padraic-shafer:
In a24a8c4d, added an optional list of self-serializing types to `safe_json_dump()`. This seemed like the cleanest option to avoid a circular import between `tiled.utils`, `tiled.ndslice`, and `tiled.type_aliases`.

tiled/catalog/adapter.py
```diff
@@ -1226,9 +1232,20 @@ async def write_block(self, block, media_type, deserializer, entry, body):
             (await self.get_adapter()).write_block, data, block
         )
 
-    async def patch(self, shape, offset, extend, media_type, deserializer, entry, body):
+    async def patch(
+        self, shape, offset, extend, media_type, deserializer, entry, body, persist=True
+    ):
+        if extend and not persist:
```
danielballan: This can be caught in the endpoint (`router.py`) and is arguably a `400`. I would argue it is less a conflict with the current state of the server and more an incompatibility between parameters in the request. That is, we do not need to check the current state of the server to know that the request is invalid on its own terms.

danielballan (commented):
Seems solid! I love the docs.

Just two comments on error handing: one here in the review and another (accidentally) submitted as a standalone comment.

PR #1214, opened 2025-11-04
UI/auth

AditiChikkali:
- Added user authentication system with JWT tokens
- Added login/logout functionality with protected routes
-  Implemented automatic token refresh before expiration
- Configured axios interceptors to automatically attach auth headers


web-frontend/vite.config.js
```diff
@@ -13,12 +12,17 @@ if (!global.crypto) {
 export default defineConfig({
   base: "/ui/",
   server: {
+    //https: true,
     proxy: {
       "/api": {
-        target: "http://127.0.0.1:8000",
+        target: "https://tiled-staging.nsls2.bnl.gov",
```
danielballan: We should keep this URL out of the codebase, I think. Shall we update the instructions for running a local server to configure that local server to use auth? There are toy examples that run toy authentication like:

```
ALICE_PASSWORD=secret tiled serve config example_configs/toy_authentication.yml
```

danielballan:
Folded into #1350

PR #1217, opened 2025-11-06
Add delete:revision, delete:node scopes

### LINKED ISSUE #742: Add scopes `delete:metadata` and `delete:data`
Summarizing a discussion with @dylanmcreynolds 

---

Currently, the deletion routes require `write:metadata` and `write:data` Scopes:

https://github.com/bluesky/tiled/blob/addcda65680c9a93557244301f8b286b843321e7/tiled/server/router.py#L1216

Let's separate out Scopes for deletion so that Roles and Access Control Policies can enable writing but block deleting if desired. (This issue is not taking a policy on what default/common Scopes should be, only that these should be separated out now so that it is _possible_ to distinguish them.)

Specifically:
- `delete:metadata` --- May remove rows from the Catalog database, dropping references to externally-managed data.
- `delete:data` --- May delete actual files, internally-managed files that were written by the Tiled server and may be the canonical copy of a given dataset. (Deleting the files should also clean up the SQL entities referencing them.)

We may need to add and/or rename routes:

- Existing route `DELETE /metadata/{path}` should change from scopes `["write:metadata", "write:data"]` to scopes `["delete:metadata"]`.
- Existing route `DELETE /nodes/{path}` should maybe be split into `DELETE /nodes/metadata/{path}` (can delete _references_ to externally-managed data, or empty nodes) and `DELETE /nodes/assets/{path}` (can delete actual data), protected by `delete:metadata` and `delete:data` respectively.

nmaytan:
Closes #742. We believe, given that delete operations are distinct and not mirrors of write/create operations, that specific names for these scopes are more appropriate.

### Checklist
- [x] Add a Changelog entry


tiled/access_control/scopes.py
```diff
@@ -3,6 +3,8 @@
     "read:data": {"description": "Read data."},
     "write:metadata": {"description": "Write metadata."},
     "write:data": {"description": "Write data."},
+    "delete:metadata": {"description": "Delete metadata."},
```
genematx: I feel like "delete_metadata" operation itself is not well-defined, as one can't DELETE metadata (i.e. without deleting the entire node), only overwrite or update it. So, probably, there's no need to have a dedicated scope for this; "write_metadata" is enough.

nmaytan: This is an interesting consideration. I could muse on this in different ways, but I think the clearest case is that deleting revisions of metadata seems like something that should have a `delete:metadata` scope.

genematx: Yeah, I think I agree, if `delete:metadata` really means "delete revisions". What's not clear to me is, does someone HAVE TO have the `delete:metadata` to delete data (in addition to `delete:data` of course), or is it assumed by default in this case?
I see one use case for `delete:metadata` though: we have this flag "drop_revision" when we update metadata; we may require someone to have the `delete:metadata` scope if they want to set this flag to True. On the other hand, TiledWriter relies on this flag to keep the things clean, so the writing principal will have to have this scope as well.

nmaytan: Great point about the `drop_revision` flag, this is something I overlooked. It does feel right, given our intention, that deciding to change the metadata without saving a revision is functionally equivalent to something that should require the delete scope.

OTOH, changing the metadata via write is a write, which rightfully requires write scope. You aren't _technically_ deleting any entities if you use `drop_revision`, whereas deleting a revision is.  While I do feel the intuition that it makes sense to lock this with delete scope, I think I land on "this remains just write scope, even with the drop flag".

danielballan:
> Yeah, I think I agree, if delete:metadata really means "delete revisions". 

Thoughts on naming this `delete:revisions`? The asymmetry in naming matches the asymmetry in functionality. While you can `write` or `read` just the metadata, you cannot `delete` just the metadata (of the current node). What you can delete is a `revision`. This name would make that clear.

nmaytan:
> > Yeah, I think I agree, if delete:metadata really means "delete revisions".
> 
> Thoughts on naming this `delete:revisions`? The asymmetry in naming matches the asymmetry in functionality. While you can `write` or `read` just the metadata, you cannot `delete` just the metadata (of the current node). What you can delete is a `revision`. This name would make that clear.

A few thoughts.

The deletion route, when it deletes a node, is possibly deleting both data and metadata. Therefore, I think `delete:metadata` is still a scope we want to impose on the delete route.

Also, making the scope that specific - I don't disagree with it, but it gives me pause if we want to be that specific from the jump & impose having an additional scope to include everywhere (assuming that `delete:metadata` sticks around otherwise).

Lastly, the symmetry of read, write, delete feels correct even **if** the scope were to be initially unused. Sort of a rule of 3/5 kind of gut feeling. That isn't necessarily a reason not to have a more specific scope, but it also would mean keeping the scope anyway.

danielballan:
Good to merge once the CHANGELOG conflicts are fixed.

nmaytan:
After reflecting on this and philosophizing about scopes a lot more (and conversations with @genematx) I think we are convinced that `delete` operations are not mirrors of `write` or `create` operations and that more specific names are appropriate and much more clear.

danielballan:
Great. I thought of suggesting `delete:node` but did not find time to properly consider whether I could defend the suggestion. I am happy with this outcome.

Issue #1198, opened 2025-10-23
Implement automatic reconnection for WS


danielballan:
If a websocket connection is severed due to network issues for a server restart, a `Subscription` should automatically reconnect, starting from the most recently processed `sequence`.

That is, this loop should keep track of the last `sequence` it has successfully received. On disconnect, it should resume from there using `self._connect(start=1 + sequence)`. Since this can be done seamlessly, it is probably _not_ necessary to log a warning for the user.

https://github.com/bluesky/tiled/blob/bdbfe024441244bc842185fadfd0ccb96cd45e96/tiled/client/stream.py#L233-L247

danielballan:
I had two thoughts about this.

First, it would be sensible to make a WS-specific version of our HTTP `retry_context`, perhaps sharing the same env var configuration (at least to start) for simplicity.

https://github.com/bluesky/tiled/blob/6b926ede90b6eca156d6bc66ccc9f7de8df91818/tiled/client/utils.py#L121-L128

This implements the principles of good retries: back off, jitter, and eventually give up to avoid hammering a system that is in distress.

Second, what if the system is trying to resume from sequence number X, but by the time it manages to resume the who sequence has been evicted, or the Redis has been flushed and the counter is restarting from 1? How would it tell? And how should it behave?

How would it tell: maybe the initial `Schema` message should hold a UUID for that series, stashed along with the counter in Redis. If the UUID looks different after reconnection, the client would know that the sequence numbers have reset. (A UUID4 comes to mind first, but there are also UUID types like UUID6 that are time-sorted. It might be useful to know whether the unexpected UUID is "later" or "earlier" than the one we were expecting to see.)

How would it behave the counter reset: it could raise an exception and leave it to user whether to resubscribe or do something.

PR #1218, opened 2025-11-07
Subscription Retry

gwbischof:
Fixes https://github.com/bluesky/tiled/issues/1198

There is a trick here that fixed the hanging tests. (that I got help from Eugene to figure out.)
We had to add a lock on the Websocket connection, so that disconnect doesn't get called from the main thread while the connection is in progress.

### DIFF
--- a/tiled/client/stream.py
+++ b/tiled/client/stream.py
@@ -462,7 +462,7 @@ class Subscription(abc.ABC):
             # Websocket may not have been fully connected
             pass
         self.disconnected.process(self)
-        self.executor.shutdown()
+        self.executor.shutdown(wait=True)
 
     def disconnect(self, wait=True) -> None:
         "Close the websocket connection."


danielballan (commented):
Good coverage!

tiled/client/stream.py
```diff
@@ -311,29 +327,60 @@ def _connect(self, start: Optional[int] = None) -> None:
...
         "Blocking loop that receives and processes updates"
-        while not self._disconnect_event.is_set():
-            try:
-                data = self._websocket.recv(timeout=RECEIVE_TIMEOUT)
-            except (TimeoutError, anyio.EndOfStream):
-                continue
-            if data is None:
-                self.stream_closed.process(self)
-                self._disconnect()
+        for attempt in self._websocket_retry_context():
```
danielballan: I see another structural issue here. We're using a single "retry context" for the entire lifecycle of the subscription. Imagine that we run for days and get disconnected once a day. We want a fresh "retry context" each time we get disconnected.

There are three loops:

1. Loop forever (until `self._disconnect_event.is_set()`) where "disconnect" here refers to the Subscription, not to a particular websocket.
2. Loop over retries when disconnected, and give up if retries are exhausted.
3. Loop over polling `recv()` calls.

Here is a sketch that introduces one new method, `_run` for the outer loop.

```py
def _run(self):
    "Receive websocket messages, reconnecting when necessary."
    while not self._disconnect_event.is_set():
        self._connect()
        try:
            self._receive()
        except (websockets.exceptions.ConnectionClosedError, OSError):
            logger.debug("Disconnected! Will attempt to reconnect")
            continue  # reconnect

def _receive(self):
    "Receive and process websocket messages."
    while not self._disconnect_event.is_set():
        try:
            data = self._websocket.recv(timeout=RECEIVE_TIMEOUT)
        except (TimeoutError, anyio.EndOfStream):
            continue
        ...

def _connect(self):
    logger.debug("Disconnected! Attempting reconnection.")
    for attempt in stamina.retry_context(
            on=(
                websockets.exceptions.ConnectionClosedError,
                OSError,
            ),
            ...
    ):
        with attempt:
            ...
```

gwbischof: Isn't this the same as setting the _connect retries to unlimited?

danielballan: Based on my reading, I am appreciating that unlimited retries is basically never the right solution. If something is persistently down, eventually the client should give up, and manual recovery becomes a feature rather than a bug. For a passive service like the end-of-run consumer, the retries might be generous—more generous than a client running in an interactive session like a Jupyter notebook—but still finite.

> And if its able to connect the number of retries get reset

Yes. The feature of the two loops (well, three loops if you could the `recv` loop) is riding out an _unlimited_ number of dropped-connection events, but a _limited_ number of re-connection attempts per dropped-connection event.

A service running for a very long time may see an unlimited number of disconnection events, and that's fine. But it should not hammer a distressed service indefinitely before giving up.

tiled/client/stream.py
```diff
@@ -259,6 +266,8 @@ def __init__(
         self._disconnect_lock = threading.Lock()
         self._disconnect_event = threading.Event()
         self._thread = None
+        self._last_received_sequence = None  # Track last sequence for reconnection
+        self._connected = False  # Track connection state
```
danielballan: Under the general rule of "state is the root of all evil" I wonder if we can avoid adding this state. I think that the control flow ensures that `self._connect(...)` only gets called (1) when `self._run(...)` starts and (2) after a connection error is raised. So the `if self._connected: return` cut-out is protecting a codepath that, in fact, you can never hit.

tiled/_tests/test_subscription.py
```diff
@@ -24,6 +26,21 @@
...
+
+    stamina.set_active(True)
+    yield
+    stamina.set_active(False)
+
+
+@pytest.fixture(autouse=True)
+def fast_retries(monkeypatch):
+    """Set retry attempts to 2 for faster tests (down from default 10)."""
+    monkeypatch.setattr("tiled.client.stream.TILED_RETRY_ATTEMPTS", 2)
```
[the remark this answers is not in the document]
gwbischof: oh nice :)

PR #1250, opened 2025-12-08
Split apikeys scope into create and revoke

### LINKED ISSUE #1252: Change apikeys scope to support disposable API keys
Currently as a user there is one scope for api key operations: `apikeys`. This scopes allows both creating and revoking an API key.

However, one may wish to generate a "disposable" API key: for example, an API key to be used for an automated or short task. While you can set an expiry on the API key, you may want to instead (or additionally be able to) quickly revoke the API key from that same session when finished rather than waiting for the expiration time or by logging in elsewhere to revoke the key.

This is already possible, but not without giving that API key `apikeys` scope, which also enables that API key to mint new API keys. Instead, it should be possible for an API key to revoke itself but not be able to create new keys. This is akin to a "logout" but for API key auth.

nmaytan:
Closes #1252 

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/authn_database/migrations/versions/d829476bc173_split_apikeys_scope_into_create_and_.py
+++ b/tiled/authn_database/migrations/versions/d829476bc173_split_apikeys_scope_into_create_and_.py
@@ -18,8 +18,10 @@ depends_on = None
 
 
 ROLES = ["admin", "user"]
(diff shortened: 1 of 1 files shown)


danielballan (approved):

danielballan:
CI failure is an unrelated test that has recently shown itself to be flaky—may just need a longer timeout. Will debug/resolve in a separate PR. This can go ahead.

Issue #1180, opened 2025-10-07
Bug: registering a .zarr requires registering its parent folder


dylanmcreynolds:
Start with an empty `/data/tmp` folder.

1. Startup tiled:
```
tiled serve directory /data/tmp --public --verbose
```

2. Copy a example.zarr file into /data/tmp

3. Try and register the .zarr directory

```
 tiled register   --verbose   --api-key a_key   http://localhost:8000/api/v1/metadata   /data/tmp/example.zarr
```

Observe the errors:
```
  Overwriting '/'
  Walking '/data/bl832/example.zarr'
  Walking '/data/bl832/example.zarr/scale0'
  Walking '/data/bl832/example.zarr/scale0/image'
  Walking '/data/bl832/example.zarr/scale0/image/0'
  Walking '/data/bl832/example.zarr/scale0/image/0/0'
    SKIPPED: Could not resolve mimetype for '/data/bl832/example.zarr/scale0/image/0/0/0'
    SKIPPED: Could not resolve mimetype for '/data/bl832/example.zarr/scale0/image/0/0/1'
  Walking '/data/bl832/example.zarr/scale0/image/0/1'
    SKIPPED: Could not resolve mimetype for '/data/bl832/example.zarr/scale0/image/0/1/0'
    SKIPPED: Could not resolve mimetype for '/data/bl832/example.zarr/scale0/image/0/1/1'
```

4. Try and register the parent directory:

```
 tiled register   --verbose   --api-key a_key   http://localhost:8000/api/v1/metadata   /data/tmp
```

**Success**

This is an imperfect workaround, since you might have a new zarr in a large folder and not want to re-register everything in that folder.

PR #1254, opened 2025-12-12
Register Directories as Single Nodes

genematx:
This fixes a bug in the client-side `register` function that prevented one from registering standalone directories as single data nodes in Tiled (if allowed, e.g. zarr assets).

Closes: #1180

### Checklist
- [x] Add a Changelog entry


genematx:
> Test failures on Python versions for which only zarr 2.x is available:
> 
Fixed.

PR #1257, opened 2025-12-15
Fix typo when searching in catalog node

tpoliaw:


danielballan:
Three adapters have a `search` method: Zarr, HDF5, Mapping. But Zarr and HDF5 both raise `NotImplementedError` and Mapping cannot be registered within a catalog. So this code path is not presently meaningfully testable.

I'll merge this as a clear typo fix.

PR #1259, opened 2025-12-16
Improve AuthN DB migration script 

nmaytan:
This improves/fixes the migration script added in #1250 

### Checklist
- [x] Add a Changelog entry


danielballan:
This _can_ be done with pure SQL, and that would be more scalable than pulling batches into Python. But as we have only 396 API keys in the prod deployment (very likely the largest that exists) I think no further optimization is warranted.

PR #1262, opened 2025-12-19
Stream larger data

gwbischof:
Add a max_size param for subscriptions so that clients can specify that they want to receive larger data.
This also improves the client behavior when receiving a message that is too large.

I think we ignore the codecov failures because the Starlette TestClient.websocket_connect() doesn't support `max_size`.
I tested this on a local tiled and it has the behavior that I was going for.

Any thoughts about what the client behavior should be when receiving data that is too large?

Test against tiled-staging worked well.
```
uv run receiver.py
Listening for updates. Use Ctrl+C to stop....
Array created: 1768934662.683
Message exceeds max_size (1000 bytes). Subscription will disconnect permanently. Increase max_size in start() to receive large messages.
```

I opened an issue to fix the warnings from the docs build: https://github.com/bluesky/tiled/issues/1273

### DIFF
--- a/docs/source/tutorials/streaming.md
+++ b/docs/source/tutorials/streaming.md
@@ -179,12 +179,15 @@ serve` CLI they are `--cache-data-ttl` and `--cache-seq-ttl`.)
 ## Handling Large Data
 
 By default, WebSocket messages are limited to 1 MB. For larger updates,
-increase the `max_size` parameter when subscribing:
+increase the `max_size` parameter when starting the subscription:
 
 ```py
 # Accept updates as large as 1 GB
-sub = client.subscribe(max_size=1_000_000_000)
+sub = client.subscribe()
+sub.start_in_thread(max_size=1_000_000_000)
 ```
+
+The `max_size` parameter can be passed to either `start()` or `start_in_thread()`.
 
 ## Performance Optimization: Non-persistent Updates
 


tiled/client/stream.py
```diff
@@ -316,7 +322,28 @@ def _run(self, start: Optional[int] = None) -> None:
...
+            except (websockets.exceptions.ConnectionClosedError, OSError) as exc:
+                # Check if it's a "message too big" error
+                # The client sends close code 1009 when received message exceeds max_size
+                close_code = None
+                if isinstance(exc, websockets.exceptions.ConnectionClosedError):
+                    close_code = (
+                        exc.sent.code if hasattr(exc, "sent") and exc.sent else None
+                    )
+
+                if close_code == 1009:  # MESSAGE_TOO_BIG
```
gwbischof: I'm not sure if this is the way way to handle this, but it's at least better than silently reconnecting in a loop and trying to read the same message thats too large again.

danielballan: Indeed.

I'm not sure where we want to land on this ultimately. I'm inclined to be strict at first because we can always add affordances later. It's harder to take them away.

Logging an error describing what to do (as you have) and then _permanently disconnecting_ feels like a safe, strict initial implementation. We can grow from there.

gwbischof: How would the end-of-run consumers handle MESSAGE_TO_BIG if we do a permanent disconnect?

The process running the consumer stops, Then we notice the problem because the processing job didn't start, then someone searches the logs to find the error, updates the consumer code and then starts the service again?

Or maybe we just set the `max_size` to be really large so that we don't get an error?

I'm thinking that we should have a solid theory about how the end-of-run consumers would work with whatever design choice we make here, because I think the plan is to retire Kafka in the next cycle.

danielballan (approved):

PR #1265, opened 2026-01-08
Raise an Error when API key can not be created

genematx:
If the Tiled server is deployed in a single-user mode, it does not support creation of additional API keys. This PR raises an exception with a suitable message to indicate this.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -564,7 +564,9 @@ class Context:
         authn_links = self.server_info.authentication.links
         if (authn_links is None) or (not authn_links.apikey):
             raise RuntimeError(
-                "Server does not support API key creation for the current user."
+                "This Tiled server is deployed as a single-user server. "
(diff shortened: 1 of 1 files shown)


tiled/client/context.py
```diff
@@ -558,11 +558,18 @@ def create_api_key(self, scopes=None, expires_in=None, note=None, access_tags=No
         """
         if isinstance(expires_in, str):
             expires_in = parse_time_string(expires_in)
+
+        authn_links = self.server_info.authentication.links
+        if (authn_links is None) or (not authn_links.apikey):
+            raise RuntimeError(
+                "Server does not support API key creation for the current user."
```
danielballan: Is this too wordy?

> This Tiled server is deployed as a single-user server. It has a single, static API key. It does not support generating API keys (as a multi-user server does).

PR #1269, opened 2026-01-12
Add custom auth for tiled client

ZohebShaikh:
The tiled client currently supports non-interactive authentication exclusively through API keys. With this PR, users can authenticate using service accounts (i.e., obtain access tokens non-interactively for machine users) or take full control of the tiled-client authentication flow themselves.

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/client/constructors.py
+++ b/tiled/client/constructors.py
@@ -6,8 +6,6 @@ from urllib.parse import parse_qs, urlparse
 
 import httpx
 
-from tiled.client.auth import TiledAuth
-
 from ..utils import import_object, prepend_to_sys_path
 from .container import DEFAULT_STRUCTURE_CLIENT_DISPATCH, Container
 from .context import DEFAULT_TIMEOUT_PARAMS, UNSET, Context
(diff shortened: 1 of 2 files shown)


ZohebShaikh:
https://github.com/DiamondLightSource/blueapi/blob/token-exchange/src/blueapi/service/interface.py#L176-L237
Above is an example of how I will be using (WIP) it for service to service communication (blueapi to tiled) in non-interactive way using token exchange flow [docs](https://www.keycloak.org/securing-apps/token-exchange#_standard-token-exchange).

The problem is currently the only non-interactive way of doing this is by API-keys and API-keys are managed by tiled and will give us no information that who is trying to write to tiled(required for authZ). We got around this by providing direct access tokens that were generated by blueapi to tiled but then we faced another hurdle which is that tiled is not able to refresh this token as the tokens were not generated by tiled(using device flow which is interactive) so we made the access tokens lifespan long lived temporarily to just check tiled writing from blueapi. But having long lived access tokens is a security risk so the above process  to use token exchange flow is a better way.

There is a one to one replacement of API key in OAuth2 service accounts this will not have any information of the user. We are using this for our system-test in blueapi like this 
https://github.com/DiamondLightSource/blueapi/blob/main/tests/system_tests/test_blueapi_system.py#L97-L108

PR #1276, opened 2026-01-21
Add uri to LiveChildCreated

danielballan:
Enables:

```py
def on_new_array(update):
    print(f"URI of new array is {update.uri}")
```

### Checklist
- [ ] Add a Changelog entry


danielballan:
Power-cycled to re-run CI

PR #1283, opened 2026-01-27
Fix Node Deletion

genematx:
This applies changes to the `catalog.adapter.CatalogNodeAdapter` class along with modifications to the database that fix degraded performance of node deletion. With these changes, on large catalogs (e.g. NSLS2 prod catalog with >20M nodes), the time to delete a Bluesky run node drops from approx. 2.5 mins to several ms.

The following reasons for the slow deletion have been identified and fixed:
1. Unnecessary deletion triggers on the `nodes_closure` table configured in one of the past alembic migrations (but not in ORM) -- explicitly removed in the new migration script.
2. Suboptimal queries generated by sqlalchemy to exclude referenced Assets -- rewritten in pure SQL for Postgres, restructured for SQLite.
3. Missing indices that would have supported anti-joins for the above queries -- added.
4. Redundant indices in several tables that mirror primary keys -- removed, although this likely has minor if any effect on the performance.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,7 +5,7 @@
 		"dockerfile": "../Containerfile",
 		"target": "developer",
 		"args": {
-			"PYTHON_VERSION": "3.9"
+			"PYTHON_VERSION": "3.10"
 		}
 	},
 	"containerEnv": {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - "3.9"
           - "3.10"
           - "3.11"
           - "3.12"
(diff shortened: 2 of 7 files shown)


ZohebShaikh:
The performance improved looks really impressive.

I think tiled is maturating quite a bit are there plans to add regression tests for performance ? Might help to flag problems early on / discover problems.
There are 2 popular libraries [fastapi](https://github.com/pydantic/pydantic/blob/main/tests/benchmarks/test_fastapi_startup_simple.py) and [pydantic](https://github.com/pydantic/pydantic/blob/main/tests/benchmarks/test_fastapi_startup_simple.py) which use codespeed for performance regression. Example of what a [CI](https://github.com/pydantic/pydantic/runs/35660193667) run it does show quite a lot of details.

danielballan:
Good idea! In the scipy space, [asv](https://github.com/airspeed-velocity/asv) has been a popular choice for this. There are nascent [asv benchmarks in tiled](https://github.com/bluesky/tiled/tree/main/benchmarks) but they need some love.

I'm open to swtiching frameworks, though.

PR #1285, opened 2026-01-30
Require setting explicit scopes to use access-tag restricted apikeys

nmaytan:
When creating an API key which is restricted to a list of access tags, we previously only disallowed asking for the API key to have the designated "admin" scope. This is because those with "admin" scope bypass tag restrictions in `TagBasedAccessPolicy`. It doesn't make sense to request a key that has admin powers, but then restrict its access.

However, when an API key is created without specifying any scopes explicitly, the scopes section defaults to `inherit` - which means the key assumes all scopes that the user has.

This implicitly allowed an API key to be generated, with access tag restrictions, that ultimately bypassed said restritions due to inherited admin powers.

This behavior was expected when originally implemented, but we now recognize that this is too unexpected and confusing. Therefore, we no longer allow creating a tag-restricted API key with the `inherit` scope set on that API key. Instead, the specific scopes needed for that key should be requested.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Write the date in place of the "Unreleased" in the case a new version is release
   `FileSequenceAdapter`) are reshaped by default to match the shape declared in the
   associated structure, where possible.
 - Removed support for Python 3.9, which reached its end of life in October 2025.
+- When creating an access-tag restricted API key, the `inherit` scope is no longer valid to request. Instead, the specific scopes desired for the key should be requested.
 
 ## v0.2.3 (2025-12-17)
 


nmaytan:
No scopes supplied:
```
ClientError: 403: Requested scopes ['inherit'] contain scopes {'inherit'}, which cannot be combined with access tag restrictions. http://localhost:8000/api/v1/auth/apikey`
```

"Admin scope" supplied:
```
ClientError: 403: Requested scopes ['admin:apikeys'] contain scopes {'admin:apikeys'}, which cannot be combined with access tag restrictions. http://localhost:8000/api/v1/auth/apikey
```

"Inherit" scope explicitly entered:
```
ClientError: 403: Requested scopes ['inherit'] contain scopes {'inherit'}, which cannot be combined with access tag restrictions. http://localhost:8000/api/v1/auth/apikey
```

"Admin scope" supplied with "inherit" scope explicitly entered:
```
ClientError: 403: Requested scopes ['inherit', 'admin:apikeys'] contain scopes {'inherit', 'admin:apikeys'}, which cannot be combined with access tag restrictions. http://localhost:8000/api/v1/auth/apikey
```

Proper scopes supplied:
```
In [18]: t.context.create_api_key(scopes=["read:metadata", "read:data"], access_tags=["alice_tag"])
Out[18]:
{'first_eight': 'f7e1078d',
 'expiration_time': None,
 'note': None,
 'scopes': ['read:metadata', 'read:data'],
 'access_tags': ['alice_tag'],
 'latest_activity': None,
 'secret': '<redacted>'}
```

PR #1286, opened 2026-02-02
allow ALL_ACCESS queries

ZohebShaikh:
### Checklist
- [ ] Add a Changelog entry


danielballan:
Fair. The return type signature would be simpler if we insisted on just `NO_ACCESS | list[Query]` and updated the AccessPolicy to return `[]` here. Any feelings either way, @ZohebShaikh?

PR #1339, opened 2026-04-13
Serialization and Streaming of Table Updates

genematx:
This addresses a couple of minor issues with the WebSockets interface:

1. Ensures that `websocket.send_text` receives string (not bytes, as returned by `safe_json_dump`);
2. Extends streaming capabilities to `table-data` with the `json` envelope format.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/server/core.py
+++ b/tiled/server/core.py
@@ -142,13 +142,13 @@ def pagination_links(base_url, route, path_parts, offset, limit, length_hint):
             }
         )
     if offset + limit < length_hint:
-        links["next"] = (
-            f"{base_url}{route}/{path_str}?page[offset]={offset + limit}&page[limit]={limit}"
-        )
+        links[
+            "next"
+        ] = f"{base_url}{route}/{path_str}?page[offset]={offset + limit}&page[limit]={limit}"
     if offset > 0:
-        links["prev"] = (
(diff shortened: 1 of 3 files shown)


tiled/server/core.py
```diff
@@ -769,28 +769,39 @@ async def stream_json(
...
+                            structure.data_type.to_numpy_dtype(),
+                            metadata.get("shape"),
+                        )
                 metadata["payload"] = payload_decoded
+            # Convert non-serializable schema objects (e.g. pyarrow.Schema)
+            # to their string representation for JSON transport.
+            if "arrow_schema" in metadata:
+                metadata["arrow_schema"] = str(metadata["arrow_schema"])
             data = safe_json_dump(metadata)
-            await websocket.send_text(data)
```
danielballan, line 793: Some say that encoded JSON, being an interchange format, should be handled as bytes. This is why `orjson` and `safe_json_dump` return bytes. Should we use `send_binary`?

Doesn't matter, just an academic question.

danielballan (commented):
Nice, thanks. Is it easy enough to slap a test on that would have failed without this fix?

tiled/server/core.py
genematx: good point. I don't know... I suppose if a client wanted an encoded message, it would ask for msgpack, but then we also send encoded JSON in the usual http packets.

danielballan (approved):

PR #1343, opened 2026-04-15
Fix call to `ProxiedOIDCAuthenticator.decode_token`.

danielballan:
Follow-up on #1333. Note that `jose.jwt.decode` accepts an `access_token` (in addition the the `token` to be decoded) but only requires it in certain circumstances.

```
Docstring:
Verifies a JWT string's signature and validates reserved claims.

Args:
    token (str): A signed JWS to be verified.
    key (str or iterable): A key to attempt to verify the payload with.
        This can be simple string with an individual key (e.g. "a1234"),
        a tuple or list of keys (e.g. ("a1234...", "b3579"),
        a JSON string, (e.g. '["a1234", "b3579"]'),
        a dict with the 'keys' key that gives a tuple or list of keys (e.g {'keys': [...]} ) or
        a dict or JSON string for a JWK set as defined by RFC 7517 (e.g.
            {'keys': [{'kty': 'oct', 'k': 'YTEyMzQ'}, {'kty': 'oct', 'k':'YjM1Nzk'}]} or
            '{"keys": [{"kty":"oct","k":"YTEyMzQ"},{"kty":"oct","k":"YjM1Nzk"}]}'
        ) in which case the keys must be base64 url safe encoded (with optional padding).
    algorithms (str or list): Valid algorithms that should be used to verify the JWS.
    audience (str): The intended audience of the token.  If the "aud" claim is
        included in the claim set, then the audience must be included and must equal
        the provided claim.
    issuer (str or iterable): Acceptable value(s) for the issuer of the token.
        If the "iss" claim is included in the claim set, then the issuer must be
        given and the claim in the token must be among the acceptable values.
    subject (str): The subject of the token.  If the "sub" claim is
        included in the claim set, then the subject must be included and must equal
        the provided claim.
    access_token (str): An access token string. If the "at_hash" claim is included in the
        claim set, then the access_token must be included, and it must match
        the "at_hash" claim.
```

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/authenticators.py
+++ b/tiled/authenticators.py
@@ -207,7 +207,7 @@ properties:
         return httpx.get(self.jwks_uri).raise_for_status().json().get("keys", [])
 
     def decode_token(
-        self, id_token: str, access_token: Optional[str]
+        self, id_token: str, access_token: Optional[str] = None
     ) -> dict[str, Any]:
         return jwt.decode(
             id_token,
--- a/tiled/authenticators.py
+++ b/tiled/authenticators.py
@@ -206,7 +206,9 @@ properties:
     def keys(self) -> List[str]:
         return httpx.get(self.jwks_uri).raise_for_status().json().get("keys", [])
 
-    def decode_token(self, id_token: str, access_token: str) -> dict[str, Any]:
+    def decode_token(
+        self, id_token: str, access_token: Optional[str]
+    ) -> dict[str, Any]:
         return jwt.decode(
             id_token,
             key=self.keys(),
--- a/tests/test_authenticators.py
+++ b/tests/test_authenticators.py
@@ -140,11 +140,11 @@ def test_oidc_decoding(
 
     if not expired:
         # Decode does not currently care if issued_at_time > current time
-        assert authenticator.decode_token(encrypted_access_token, "placeholder") == access_token
+        assert authenticator.decode_token(encrypted_access_token) == access_token
 
     else:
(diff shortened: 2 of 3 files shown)


tiled/authenticators.py
```diff
@@ -206,7 +206,9 @@ def end_session_endpoint(self) -> str:
     def keys(self) -> List[str]:
         return httpx.get(self.jwks_uri).raise_for_status().json().get("keys", [])
 
-    def decode_token(self, id_token: str, access_token: str) -> dict[str, Any]:
+    def decode_token(
+        self, id_token: str, access_token: Optional[str]
+    ) -> dict[str, Any]:
```
ZohebShaikh, line 211:
```suggestion
    def decode_token(
        self, id_token: str, access_token: Optional[str] = None
    ) -> dict[str, Any]:
```
nit
Then you will not need to change the call above and in the [tests](https://github.com/bluesky/tiled/blob/c0c2be6d27c8769bda377639bcf60019ef027d9b/tests/test_authenticators.py#L143)

ZohebShaikh (approved):
Thank you Dan,
I have checked the changes in CI and it looks happy.

tiled/authenticators.py
danielballan: Great point, will do.

danielballan:
It took longer than I expected to circle back to this, but I incorporated your suggestion @ZohebShaikh, reverted the changes to the tests.

PR #1346, opened 2026-04-16
Fix: SimpleTiledServer race condition

sligara7:
This resolves the race condition where ThreadedServer.started becomes True when uvicorn opens the socket, but the app's startup_event() might still be running. The code now waits for a successful 200 response from the API before continuing.

Tested using: /tests/test_simple_server.py to verify that check for startup_event() completes does not cause issues.

### DIFF
--- a/tiled/server/simple.py
+++ b/tiled/server/simple.py
@@ -10,6 +10,7 @@ import time
 from typing import Optional, Union, cast
 from urllib.parse import quote_plus, urlparse
 
+import httpx
 import uvicorn
 
 from ..storage import SQLStorage, get_storage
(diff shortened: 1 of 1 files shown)


danielballan (commented):
Thanks! The latest batch of comments from copilot seem legit. Otherwise, this is good to go..

tiled/server/simple.py
```diff
@@ -140,11 +145,47 @@ def __init__(
...
+        # ThreadedServer.started is True as soon as uvicorn opens the
+        # socket, but FastAPI does not serve requests until the lifespan
+        # startup_event() completes. Poll /healthz to wait for that.
+        # Ensure the background server thread is shut down if anything
+        # in the readiness probe fails; otherwise it would leak.
+        try:
+            deadline = time.monotonic() + _STARTUP_TIMEOUT
+            with httpx.Client(trust_env=False) as client:
+                while True:
+                    if not self._server._thread.is_alive():
```
danielballan, line 163: Touching an internal attribute (`_thread`) from outside the object is not ideal. Making `_thread` into `thread` might be a better tradeoff here.

danielballan (changes requested):
One nit, and I notice CI failing.

danielballan (approved):
Oops, meant to "approve" after mentioning the nit-pick.

sligara7:
The failure in CI for Linux 3.13 seems to be something with: The failing job is due to a flaky/off-by-one expectation in tests/test_subscription.py::test_subscribe_after_first_update_from_middle, not an infrastructure/service startup failure.

danielballan (approved):

PR #1347, opened 2026-04-17
Adding missing frontslash in containerfile example config

burkeds:
See title.

    # Embedded tabular data storage
    - "duckdb:////storage/data.db"
    
    four forward slashes are needed after duckdb


danielballan (approved):

danielballan:
Thank you!

PR #1352, opened 2026-04-22
Accept Multiple Providers in TagBasedAccessPolicy -- ON HOLD

genematx:
Allow configuring multiple identity providers in `TagBasedAccessPolicy` by accepting either a single string or a list of strings for the provider parameter.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,9 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ### Added
 
-- Tests for the WebSocket endpoints that stream tabukar data.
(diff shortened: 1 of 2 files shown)


nmaytan:
I'm a bit concerned about this under the current implementation, as there is no protection against a case like this:

`personA `is `identity1 `according to `providerX`
`personB `is `identity2 `according to `providerX`

`personA `is `identity2 `according to `providerY`

If `identity2 `is used in AuthZ, this could give access meant for `personB `to `personA`. I wrote up some thoughts about what to do about this, I can share them.

This is why the access policy was locked to a single provider for now.

Issue #1315, opened 2026-03-25
Feature idea: Webhooks


danielballan:
Tiled uses Websockets to provide notifications to clients when datasets are created, updated, written to, or "closed". The clients must maintain active Websocket connections to the server in order to receive updates. If a client becomes disconnected, it is the client's responsibility to track their last update and either ask the server to "replay" recent updates that it missed (within limits) or use the REST API to load data from storage at rest. This "best effort" quality of service enables Tiled to service a potentially large number of clients, adapting its level of service (e.g., how much history is retained for replay) to fit within bounded RAM under varying load.

Webhooks might be a better fit for use cases such as kicking off workflow jobs when datasets are created or closed. Here, we want a stronger guarantee of delivery. The updates are less granular and therefore impose less burden on the server: we want a notification when datasets are created or closed, but not on every incremental write.

_Within_ a workflow job, the Websockets API or REST API may be used to access data, depending on the requirements of the workflow. But Webhooks may be a better fit than Websockets for kicking off the jobs themselves. Webhooks can implement retry to bridge over a brief outage in the consumer, and it can provide a history of success/failure to facilitate outage recovery. Webhooks do not require subscribers to hold active open connections to receive updates.

API sketch:

1. An authorized client can register a webhook that fires when a child of node `/x` is created. The POST body should include a URL that Tiled will be POST to. The server will store this, probably in a new PG table `webhooks`, which a foreign key to the `nodes` table.
   `POST /api/v1/webhook/target/x`

2. The client can review webhooks registered on x.
   `GET /api/v1/webhook/target/x`

3. Now say support a client creates a child of x.
   `POST /api/v1/metadata/x`

During the processing of request (3):
- JSON is stored in Redis and broadcast to any WS clients (as now)
- JSON is stored in PG and becomes available via REST API (as now)
- This event is logged in some kind of database of recent webhook events: probably a separate Redis db with a generous but finite TTL, or maybe PG.
- A starlette `BackgroundTask` is scheduled, which will send the webhook POST. This includes a reasonable but not unlimited number of retries. The outcome—success or failure, number of attemps—is added to the log mentioned above.

4. For debugging or outage recovery, client can review (recent) history of webhook invocations
   `GET /api/v1/webhook/history/{webhook_id}`

danielballan:
Additional context: The webhook tables should go in `tiled.catalog.orm`. We'll need migrations to add them to existing databases.

PR #1353, opened 2026-04-23
[#1315] webhooks

swelborn:
claude supervised PR for webhook support 

see #1315

not 100% confident in authz or secret storage

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/docs/source/user-guide/webhooks.md
+++ b/docs/source/user-guide/webhooks.md
@@ -1,0 +1,151 @@
+(webhooks)=
+
+# Webhooks
+
+```{warning}
+Webhooks are an **experimental feature**. The API and configuration format may
+change in future releases.
+```
+
+Webhooks let Tiled notify an external HTTP service whenever a catalog event
+occurs — for example, when a new entry is created or a stream is closed.  This
+is useful for triggering downstream pipelines, sending notifications, or
+integrating Tiled with other systems without polling.
+
+## Enabling webhooks
+
+Webhooks are **disabled by default**.  Add a `webhooks:` section to your
+server configuration file to turn them on:
+
+```yaml
+webhooks:
+  secret_keys:
+    - ${TILED_WEBHOOK_SECRET_KEY}
+```
+
+`secret_keys` is optional, but must be provided if you want to register
+webhooks with an HMAC signing secret (see [Verifying deliveries](#verifying-deliveries)).
+Generate a key with:
+
+```sh
+openssl rand -hex 32
+```
+
+Keys support rotation: the **first** key is used for encryption; **all** keys
+are tried during decryption.  Add the new key at the top, redeploy, then remove
+the old key and redeploy again once in-flight webhooks have drained.
+
+## Registering a webhook
+
+Webhooks are registered against a node path.  Any event on that node — or any
+of its descendants — will be delivered.
+
+```sh
+curl -X POST https://my-tiled-server/api/v1/webhooks/target/my/container \
+  -H "Authorization: Apikey $TILED_API_KEY" \
+  -H "Content-Type: application/json" \
+  -d '{
+        "url": "https://my-service.example.com/hooks/tiled",
+        "secret": "my-hmac-secret",
+        "events": ["container-child-created", "stream-closed"]
+      }'
+```
+
+| Field    | Required | Description |
+|----------|----------|-------------|
+| `url`    | yes      | HTTPS endpoint that will receive `POST` requests |
+| `secret` | no       | HMAC signing secret; requires `secret_keys` in server config |
+| `events` | no       | List of event types to deliver; omit to receive all events |
+
+Registering a webhook requires the `write:metadata` scope.
+
+The webhook URL **must use HTTPS**.
+
+## Event types
+
+| Type | When it fires |
+|------|---------------|
+| `container-child-created` | A new entry is created inside the watched container |
+| `container-child-metadata-updated` | Metadata on a child entry is updated |
+| `stream-closed` | A live-stream entry is closed |
+
+Each delivery is a JSON `POST` with a body like:
+
+```json
+{
+  "type": "container-child-created",
+  "timestamp": "2025-01-15T12:34:56.789Z",
+  "key": "scan_001",
+  "path": ["my", "container"],
+  "structure_family": "array",
+  "specs": [],
+  "metadata": {}
+}
+```
+
+(verifying-deliveries)=
+## Verifying deliveries
+
+When a `secret` is set, Tiled adds an `X-Tiled-Signature` header to every
+request:
+
+```
+X-Tiled-Signature: sha256=<hex-digest>
+```
+
+Verify it in your receiver:
+
+```python
+import hashlib, hmac
+
+def verify(body: bytes, secret: str, header: str) -> bool:
+    expected = "sha256=" + hmac.new(
+        secret.encode(), body, hashlib.sha256
+    ).hexdigest()
+    return hmac.compare_digest(expected, header)
+```
+
+Each delivery also carries an `X-Tiled-Event-ID` header — a unique hex string
+you can use to deduplicate retried deliveries.
+
+## Managing webhooks
+
+**List** webhooks registered on a node:
+
+```sh
+curl https://my-tiled-server/api/v1/webhooks/target/my/container \
+  -H "Authorization: Apikey $TILED_API_KEY"
+```
+
+**Delete** a webhook (use the `id` returned when you registered it):
+
+```sh
+curl -X DELETE https://my-tiled-server/api/v1/webhooks/42 \
+  -H "Authorization: Apikey $TILED_API_KEY"
+```
+
+**View delivery history** for a webhook:
+
+```sh
+curl "https://my-tiled-server/api/v1/webhooks/history/42?limit=20" \
+  -H "Authorization: Apikey $TILED_API_KEY"
+```
+
+History records are retained for 30 days and then pruned automatically.
+
+## Retry behaviour
+
+If the target URL returns a non-2xx response or is unreachable, Tiled retries
+up to **3 times** with exponential back-off (roughly 1 s → 5 s → 25 s, plus
+random jitter).  The final outcome (`success` or `failed`) is recorded in the
+delivery history.
+
+## Security notes
+
+* Webhook URLs must use HTTPS.
+* Tiled blocks registration of URLs that resolve to private, loopback, or
+  link-local addresses (SSRF protection).  For production, also route outbound
+  webhook requests through a network-level egress proxy (e.g.
+  [Smokescreen](https://github.com/stripe/smokescreen)) so that DNS-rebinding
+  attacks cannot bypass hostname-level checks.
+* HMAC signing secrets are encrypted at rest using the server's `secret_keys`.
--- a/docs/source/user-guide/webhooks.md
+++ b/docs/source/user-guide/webhooks.md
@@ -23,17 +23,11 @@ webhooks:
     - ${TILED_WEBHOOK_SECRET_KEY}
 ```
 
-`secret_keys` is optional, but must be provided if you want to register
-webhooks with an HMAC signing secret (see [Verifying deliveries](#verifying-deliveries)).
-Generate a key with:
+`secret_keys` is required.  Generate a key with:
 
 ```sh
 openssl rand -hex 32
 ```
-
-Keys support rotation: the **first** key is used for encryption; **all** keys
-are tried during decryption.  Add the new key at the top, redeploy, then remove
-the old key and redeploy again once in-flight webhooks have drained.
 
 ## Registering a webhook
 
@@ -55,9 +49,9 @@ curl -X POST https://my-tiled-server/api/v1/webhooks/target/my/container \
 |----------|----------|-------------|
 | `url`    | yes      | HTTPS endpoint that will receive `POST` requests |
 | `secret` | no       | HMAC signing secret; requires `secret_keys` in server config |
-| `events` | no       | List of event types to deliver; omit to receive all events |
+| `events` | no       | List of event types to deliver; set `"events": null` or omit to receive all events |
 
-Registering a webhook requires the `write:metadata` scope.
+Registering a webhook requires the `write:webhooks` scope (admin only).
 
 The webhook URL **must use HTTPS**.
 
@@ -106,7 +100,17 @@ def verify(body: bytes, secret: str, header: str) -> bool:
 ```
 
 Each delivery also carries an `X-Tiled-Event-ID` header — a unique hex string
-you can use to deduplicate retried deliveries.
+you can use to deduplicate retried deliveries, for example:
+
+```python
+seen = set()
+def handle(request):
+    event_id = request.headers["X-Tiled-Event-ID"]
+    if event_id in seen:
+        return  # duplicate delivery, ignore
+    seen.add(event_id)
+    ...
+```
 
 ## Managing webhooks
 
@@ -143,9 +147,13 @@ delivery history.
 ## Security notes
 
 * Webhook URLs must use HTTPS.
-* Tiled blocks registration of URLs that resolve to private, loopback, or
-  link-local addresses (SSRF protection).  For production, also route outbound
-  webhook requests through a network-level egress proxy (e.g.
-  [Smokescreen](https://github.com/stripe/smokescreen)) so that DNS-rebinding
-  attacks cannot bypass hostname-level checks.
+* Registering and listing webhooks requires the `write:webhooks` / `read:webhooks`
+  scope, which is granted to admin users only.  This is the primary access-control
+  layer preventing unauthorized webhook registration.
+* At registration time, Tiled additionally checks that the target URL does not
+  resolve to a private, loopback, or link-local address (SSRF protection).
+  Note that this hostname-level check can be bypassed by DNS-rebinding attacks;
+  for high-security deployments also route outbound webhook requests through a
+  network-level egress proxy (e.g.
+  [Smokescreen](https://github.com/stripe/smokescreen)).
 * HMAC signing secrets are encrypted at rest using the server's `secret_keys`.
--- a/tiled/catalog/orm.py
+++ b/tiled/catalog/orm.py
@@ -641,7 +641,6 @@ class WebhookDelivery(Timestamped, Base):
     # separate events table.
     event_type = Column(Unicode(128), nullable=False)
     # Full JSON payload that was (or would be) POSTed.
-    # TODO: keeping these in here may bloat database, we should consider expiry/pruning
     payload = Column(JSONVariant, nullable=False)
     # HTTP status code of the last attempt, NULL while pending.
     status_code = Column(Integer, nullable=True)
--- a/tiled/catalog/orm.py
+++ b/tiled/catalog/orm.py
@@ -630,7 +630,12 @@ class WebhookDelivery(Timestamped, Base):
         nullable=False,
         index=True,
     )
-    # Stable identifier for the logical event; retries share the same event_id.
+    # Stable identifier for the logical event sent in the webhook payload.
+    # Included in the POST body so that receiving endpoints can deduplicate
+    # deliveries: if an earlier attempt timed out and was retried, the remote
+    # may receive the same payload twice with the same event_id.  Retries
+    # within a single _deliver() call update this row in-place and share this
+    # event_id rather than creating a new delivery row.
     event_id = Column(Unicode(128), nullable=False, index=True)
     # Denormalised event info so history is queryable without joining to a
     # separate events table.
--- a/tiled/catalog/orm.py
+++ b/tiled/catalog/orm.py
@@ -642,9 +642,7 @@ class WebhookDelivery(Timestamped, Base):
     status_code = Column(Integer, nullable=True)
     # Total number of attempts made so far.
     attempts = Column(Integer, nullable=False, default=0, server_default=text("0"))
-    delivered_at = Column(
-        DateTime(timezone=True), nullable=True
-    )  # TODO/Question: why timezone=False in others?
+    delivered_at = Column(DateTime(timezone=False), nullable=True)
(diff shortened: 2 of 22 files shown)


genematx:
Hi @swelborn, thanks for putting this together! We will aim to review this asap. Would you mind if I submit my changes directly into your branch?

swelborn:
Nope, be my guest

danielballan:
This is really fun, @swelborn. Thanks for jumping in. I was just on call with Jana Thayer who said, "Isn't it awesome?"

---

Some quick high-level comments based on a skim:

I would like webhooks to get their own section in the config file, e.g.,

```yaml
catalog:
    ...
streaming_cache:
    ...
webhooks:
    ...
```

Suggestion: As with `streaming_cache`, the `webhooks` section should be optional and, if absent, webhook should be disabled. Maybe not indefinitely, but as a way of introducing this experimental feature carefully. Also, this config section should have its own `secret_keys` section. I believe the PR is currently reusing `secret_keys` from `authentication` (i.e., the keys used to sign JWTs for OAuth2 tokens). These are not always present, and so it would be better for webhooks to use their own secrets.

I see webhooks are run on a `BackgroundTask`, with some retries. If they fail, the outcome is stored in the database, and the client can potentially review re-run. That seems correct to me. For the retries, consider using [stamina](https://github.com/hynek/stamina) which includes good practices like back-off and jitter. We already use it in the Tiled client, so we may as well use it here rather than roll our own.

Again forgive the quick read: Did the word `/webhooks/` end up in the path? I think we should mount the router such that we get `/api/v1/webhooks/target/{path}`. Good to have a clear namespace for this.

swelborn:
:)

> I believe the PR is currently reusing secret_keys from authentication (i.e., the keys used to sign JWTs for OAuth2 tokens). These are not always present, and so it would be better for webhooks to use their own secrets.

yeah that is kind of what i figured, I didn't trust the use of those secret keys but i wanted to get your feedback first

> As with streaming_cache, the webhooks section should be optional and, if absent, webhook should be disabled. Maybe not indefinitely, but as a way of introducing this experimental feature carefully. 

agree. i do think it may be a little tricky since it alters the database, we would have to do something to avoid doing the migration. We use a separate router from the main router/zarr routers. So at the api layer we just need to add an `if`.

> consider using [stamina](https://github.com/hynek/stamina)

yes stamina is a nice lib. we are using it in `webooks.py`. 

> I see webhooks are run on a BackgroundTask, with some retries. 

It is not a starlette `BackgroundTask` as define in your spec in #1315, we just use asyncio task for it guarded by a semaphore (we could adjust this to use `anyio.CapacityLimiter`). I haven't used `BackgroundTask` before, but it seems like it is a bit less capable than straight up async task - from the [docs](https://www.starlette.dev/background/):

> The tasks are executed in order. In case one of the tasks raises an exception, the following tasks will not get the opportunity to be executed.

they run in order of the list, so if we have several webhook defined they will not run concurrently. Am I reading that right? 

> Again forgive the quick read: Did the word /webhooks/ end up in the path? I think we should mount the router such that we get /api/v1/webhooks/target/{path}. Good to have a clear namespace for this.

We have `/api/v1/webhook/`. we define it in the router instantiation at the top of the factory method. I was going to use `/api/v1/webhooks` (plural) which is for sure more "standard" rest, but I wanted to follow #1315 as closely as possible first.

danielballan:
Got it, thanks for your patience for my misreads; I'm aiming to get feedback out quickly amidst other things.

I think it would be fine to create the webhook tables always, unconditionally. The presence or absence of a `webhooks:` configuration section can activate the `/webhooks` router and associated machinery. (Similarly, the `/auth` endpoints are conditional on `authentication:` configuration.)

To be honest, I hadn't used `BackgroundTask` either. Perhaps you are right, this use case might be a little beyond what `BackgroundTask[s]` is designed for.

Good point about plural `webhooks` for idiomatic REST.

swelborn:
no prob at all

one thing i noticed - do we have any testing infra for migrations? what I have done in the past is simply spin up a [testcontainer](https://testcontainers.com/) and run all of the upgrades/downgrades just so we don't run into rollback problems.

this would be in another pr, would be a nice to have

genematx:
> do we have any testing infra for migrations?

no, we don't. and yes, it would be really nice to have!

genematx:
@swelborn it turned out to be more changes than I thought initially, so [here's a PR](https://github.com/swelborn/tiled/pull/1) :) Feel free to change anything or cherry-pick, of course. There probably will be more once a human can find some time to test it thoroughly, but I'll create a new PR then.

swelborn:
i rebase on main after the prs merged yesterday (hopefully did that right! :D)

swelborn:
i added some docs and one more fix that will hopefully make tests pass (still running atm)

@genematx can you double check that was the correct fix for the tests? and look into the claude-generated docs

swelborn:
okay, test are now passing except 3.13? no idea on this ...

swelborn:
this seems a little unrelated to the current PR. apparently problem is with GIL in 3.13 and the fix claude suggest is here:

```python

class Subscription(abc.ABC):
    """
    ...
    """

    def __init__(
        self,
        context: Context,
        segments: List[str] = None,
        executor: Optional[concurrent.futures.Executor] = None,
    ):
        ...
        self._executor = executor or concurrent.futures.ThreadPoolExecutor(
            max_workers=1 # CHANGED FROM 5
        )
        ...
```

Not sure if that is the right fix probably has some implications in <3.13. Just putting here... this is its root cause finding:

> Root cause: CallbackRegistry.process calls executor.submit(callback, update) using a ThreadPoolExecutor(max_workers=5). When sequential updates arrive rapidly (especially during replay of historical data), different worker threads pick up adjacent updates and execute them out of order. Python 3.13's more aggressive GIL release makes this race consistently reproducible.

genematx:
Thanks, @swelborn. I've had brief look; will read through the docs more carefully later today.

some tests (especially the ones related to WebSockets) do fail occasionally -- there might be a hidden race condition somewhere, although we've fixed the most obvious ones recently. Usually rerunning them helps.

danielballan:
That test has been intermittently flaky; I re-ran it and then saw your comment.

That diagnosis seems right! There are two paths we could take:

1. Keep the threadpool parallelism. Update the test to remove the assumption that updates are processed in order.
2. Reduce the threadpool worker count to 1 to guarantee that updates are processed in order.

Depending on the use case, I could see user's preferring either one. Perhaps in-order processing is a good default, and users can configure parallelism if they want to trade guaranteed ordering for parallel processing.

danielballan:
Addressed in a separate PR: https://github.com/bluesky/tiled/pull/1357

danielballan:
I think it would be OK to require `secret_keys`. Thoughts? Less conditionality, less to test/maintain. We're already strict about "HTTPS only."

---

I think this is a merge-able scope of work. Some suggestions for follow-up work in future PRs:

- Python client API for webhook management
- Add metrics to `/metrics`, such as the size of webhook "outbox" queue and maybe failure rate

docs/source/user-guide/webhooks.md
```diff
@@ -0,0 +1,151 @@
...
+
+`secret_keys` is optional, but must be provided if you want to register
+webhooks with an HMAC signing secret (see [Verifying deliveries](#verifying-deliveries)).
+Generate a key with:
+
+```sh
+openssl rand -hex 32
+```
+
+Keys support rotation: the **first** key is used for encryption; **all** keys
```
genematx: This rotation and redeployment procedure sounds a bit too complicated for production settings. I wonder if we should even mention it here in docs?

danielballan: For what it's worth, this is how the JWT signing keys work too. I don't think I invented this idea. Is there a more straightforward approach we should consider?

tiled/catalog/orm.py
```diff
@@ -571,3 +571,87 @@ class Revision(Timestamped, Base):
...
+    # deliveries: if an earlier attempt timed out and was retried, the remote
+    # may receive the same payload twice with the same event_id.  Retries
+    # within a single _deliver() call update this row in-place and share this
+    # event_id rather than creating a new delivery row.
+    event_id = Column(Unicode(128), nullable=False, index=True)
+    # Denormalised event info so history is queryable without joining to a
+    # separate events table.
+    event_type = Column(Unicode(128), nullable=False)
+    # Full JSON payload that was (or would be) POSTed.
+    # TODO: keeping these in here may bloat database, we should consider expiry/pruning
```
danielballan: Pruning is implemented. This comment can be dropped.

docs/source/user-guide/webhooks.md
```diff
@@ -0,0 +1,151 @@
...
+
+def verify(body: bytes, secret: str, header: str) -> bool:
+    expected = "sha256=" + hmac.new(
+        secret.encode(), body, hashlib.sha256
+    ).hexdigest()
+    return hmac.compare_digest(expected, header)
+```
+
+Each delivery also carries an `X-Tiled-Event-ID` header — a unique hex string
+you can use to deduplicate retried deliveries.
```
genematx:
```suggestion
you can use to deduplicate retried deliveries, for example:

```python
seen = set()
def handle(request):
    event_id = request.headers["X-Tiled-Event-ID"]
    if event_id in seen:
        return  # duplicate delivery, ignore
    seen.add(event_id)
    ...
```
```

File: docs/source/user-guide/webhooks.md [hunk 4]
```diff
@@ -0,0 +1,151 @@
...
+        "secret": "my-hmac-secret",
+        "events": ["container-child-created", "stream-closed"]
+      }'
+```
+
+| Field    | Required | Description |
+|----------|----------|-------------|
+| `url`    | yes      | HTTPS endpoint that will receive `POST` requests |
+| `secret` | no       | HMAC signing secret; requires `secret_keys` in server config |
+| `events` | no       | List of event types to deliver; omit to receive all events |
```
genematx:
```suggestion
| `events` | no       | List of event types to deliver; set `"events": null` or omit to receive all events |
```

tiled/server/webhooks.py
```diff
@@ -0,0 +1,452 @@
...
+    Carries the HTTP status code so it can be stored after all attempts.
+    """
+
+    def __init__(self, detail: str, status_code: int) -> None:
+        super().__init__(detail)
+        self.status_code = status_code
+        self.detail = detail
+
+
+# --- SSRF protection — blocklist ---
```
tacaswell, line 72: Is this actually needed?  webhooks should only be doing `post` so I'm not sure how it could be used to exfiltrate anything and are in general going to untrusted (from tiled's point of view) urls.

tiled/server/webhooks.py
```diff
@@ -0,0 +1,452 @@
...
+            self._client = None
+
+    async def dispatch(self, event: WebhookEvent, node_id: int) -> None:
+        if self._client is None:
+            return
+
+        event_payload = event.model_dump(mode="json")
+
+        async with self._session_factory() as db:
+            # Find webhooks registered on any ancestor of (or equal to) node_id.
```
danielballan, line 370: This has tricky implications for authorization. As we're in the middle of discussing simplifications to the access control system, I propose the following:

Define new scopes for webhooks, protect the endpoints in the webhook route with those scopes, and grant the scopes to the admin role only. This will satisfy our initial use cases, where a Tiled administrator and configuing pipelines to OpenMetadata or workflow launchers.

Later, we can look at getting fine-grained authorization right and make these scopes available to all Tiled users.

Scopes might be `read:webhooks` and `write:webhooks`? I have an open mind here.

danielballan (commented):
~~AuthZ~~ Edit: I commented on AuthZ in line.

tiled/server/webhooks.py
danielballan: Since Tiled is behind our firewall, an attacker could use webhooks to explore our internal network. Webhook success/failure state (and error messages about the failure) are accessible to the client. I could also imagine some unprotected web app that takes some action unconditionally on `POST`, which webhooks could be used to hit.

docs/source/user-guide/webhooks.md
genematx: ~~this makes sense for JWT, of course, but we don't rotate them manually in the server config. Also JWT token expire, these keys are meant to be stable.~~
ah sorry, you meant JWT _signing keys_, of course; I read it wrong. I honestly didn't know if (and how) we rotate them.

The general idea of key rotation is solid, but I think we are drawing too much attention to admin details in the user-facing docs (I just stumbled when I was reading -- it's not clear how one would tell when "in-flight webhooks have drained", if it's even possible). Maybe marking it as a "note" or a putting it in a foldable section could work.

swelborn:
> I think it would be OK to require `secret_keys`. Thoughts? Less conditionality, less to test/maintain. We're already strict about "HTTPS only."
> 
> I think this is a merge-able scope of work. Some suggestions for follow-up work in future PRs:
> 
>     * Python client API for webhook management
> 
>     * Add metrics to `/metrics`, such as the size of webhook "outbox" queue and maybe failure rate

yes let make it required and follow the same pattern as cache settings

tiled/server/webhooks.py
swelborn: thanks for responding @danielballan

swelborn:
Thanks for adding that enh issue @danielballan 

For this PR, I am second guessing the pruning mechanism here. I think we want to avoid metadata bloat, but i would imagine we would like to track history, at least for failed deliveries analagous to a deadletter queue 

thoughts?

danielballan:
At NSLS-II our `nodes` table has ~100M rows and counting, many holding a hefty chunk of JSON metadata. I think 30 days (or similar) of delivery history is a drop in the bucket.

tiled/server/webhooks.py
swelborn: @danielballan this sounds good

i have a clarifying Q about this - 

in "single user" mode it seems we only get user scopes if i read this correctly: https://github.com/bluesky/tiled/blob/22dc277553e4960a0f36ba0ff1e3bce56d6714bc/tiled/server/authentication.py#L352-L358

is that right? 

in that case single user mode I guess disallow use of webhooks if we added the `read:webhooks` and `write:webhooks` to admin only

fine by me, just want to check. this will also require changes to the tests which are set up for api key auth

docs/source/user-guide/webhooks.md
swelborn: fine by me to remove it from user-facing docs

swelborn:
updated auth db, etc. for scopes.

tiled/server/webhooks.py
danielballan: Good question. The variable name `USER_SCOPES` is confusing here; we had better name it `SINGLE_USER_SCOPES`.

In "single-user mode" you are either authenticated or you're not; there is no user/admin distinction. The `[SINGLE_]USER_SCOPES` should include the new `*:webhooks` scopes.

In a multi-user deployment---i.e., a deployment with `authentication.providers` configured---the scopes for the `user` and `admin` roles are defined here.

https://github.com/bluesky/tiled/blob/4c3d9b68f13f6866f4a25a4605ac0fdbb602db7e/tiled/authn_database/core.py#L32-L67

Note that there is precedent for this in the code base: we have the same situation with the `register` scope. As with the webhook scopes, we may in the future make `register` available to users (once we can be sure the authorization logic around it is airtight). But for now, the `user` role in a multi-user server does _not_ have `register`, while `USER_SCOPES` for a single-user server does.

swelborn:
added a dev mode here so that we can develop locally, i think it is necessary (otherwise it becomes really complicated and hacky for what you have to do locally to get it to work (believe my i tried)) 

This is a pretty standalone PR for a stub that receives webhook events:
https://github.com/carbonscott/tiled-catalog-broker/pull/54/

It does work : )

danielballan:
That makes sense: when I saw the HTTPS-only constraint the thought did pass through my mind, "It might be annoying to have to do self-signed certs for dev." This solution seems good.

@genematx Let's do one more review pass, kick the tires interactively to check for any unexpected impacts, and get this in.

genematx:
I wonder if we should encourage developers to use `SimpleTiledServer`, which could bypass the HTTPS/SSRF requirements, but still enforce them in production? Tbh, I'm not a big fan of `dev_mode` flags in configs, in general, but `SimpleTiledServer` is something that is intended specifically for this purpose and some reduced functionality is expected.
I've put my suggested changes [here](https://github.com/swelborn/tiled/pull/2) along with a simple "getting-started" script.

danielballan:
I like that this makes it harder to circumvent security for production deployments. Jupyter has run into "security issues" where scientists who know enough programming to start Jupyter but perhaps not enough to understand security implications disable Jupyter's security features.

https://blog.jupyter.org/please-dont-disable-authentication-in-jupyter-servers-dd197206e7f6

swelborn:
That makes sense to me!

swelborn:
Are there any features we lose out on by not bringing up the whole docker stack to test locally, though?

genematx:
>  any features we lose

that's what I was wondering too... The main simplifications are that `SimpleTiledServer` uses in-memory cache (instead of Redis) and SQLite (instead of Postgres) -- both are probably fine for dev and demos (especially for demos). As long as the full-stack use case is covered in CI, at least by simple smoke tests, I personally don't think it's worth adding the explicit `dev_mode` to the core.

swelborn:
agree let's avoid it. i will merge. If we ever _do_ need the full stack it is simple enough to apply a local dev_mode patch :)

danielballan:
Sounds good to me. And, if we need it, we could in the future add a more targeted/explicit setting, like `allow_http` or `webhook_url_validator`.

swelborn:
thank you for fixing that docs build, I was just about to do the same thing :D

swelborn:
One thing I notice from my local setup is that the simple server always runs on loopback.

Consider the following situation:
- we have a tiled registration CLI tool that we want to test locally
- we have a Simple Tiled instance in a container 
- we have a an api server for registering a webhook and then receiving webhook events

Situation could get more complicated than that obviously. Since the SimpleServer binds only to localhost and not all interfaces by default, we cannot reach it even if we expose ports. I can hack around this for now, but it would be nice if we had a `host` kwarg to that server, so that our only option isn't the loopback.

Does that make sense? I can open a PR.

swelborn:
In addition, it would be nice to have the option of stable webhook secret keys in the simple server rather than one that is generated during instantiation. is that in scope for this? don't want to add too many kwargs to the simple server ...

danielballan:
I'm sold on the problem; I'm not fully sure yet about `host` as the best solution.

`SimpleTiledServer` is primarily targeting a client _in process_---e.g., I want to spawn and use Tiled from within a Jupyter notebook. (I might also pull up the UI in my web browser or do analysis in another process, though.) It makes a couple choices based on the premise that it will be "loopback only." For example, it generates a short API key with less rigorous security than we get from `tiled serve catalog --temp`.

Once you're running Tiled in its own container, I think you've graduated beyond what `SimpleTiledServer` is targeting. The container's default behavior isn't _that_ different---they both use embedded databases, for example---but it's a little more robust. Is adding `allow_http` / `TILED_WEBHOOKS_ALLOW_HTTP` the lesser evil here? It solves the specific problem without opening up the risk of `SimpleTiledServer(host='0.0.0.0')` becoming a commonly copy/pasted solution (as happened to Jupyter).

I'm happy to go with whatever you two think is best here. Just voicing thoughts.

swelborn:
I agree with you. We may eventually end up just adding all the config into simple server if we do it this way...

I feel we need an escape hatch here, maybe not calling it `dev_mode` but `allow_http` and `webhook_url_validator`. Were you picturing `webhook_url_validator` as a reference to some class or method? or just bool? 

In other words should we do the same or similar thing we were doing before, but avoid calling it `dev_mode`? 

Let us know what you think @genematx

danielballan:
I had in mind picking one of these two to put under `webhooks:` config.
1. `allow_http: bool` as a more explicitly targeted option than "dev mode" (which could grow to mean a lot of things)
2. `webhook_url_validator` as a reference to a Python function

I lean (1) because it's easy to configure but I think (2) would be fine.

genematx:
yeah, I see, it's tricky. Needing to run the server in a _container_ is exactly what I was worried about. I agree, there's no reason to make SimpleServer more complex and opening the pathway for adding the entire config there.
I like the `allow_http: bool` option under `webhooks:` -- very targeted, indeed. I don't think we'd ever need to configure any other validator than the `default` or `noop`, so probably not much to select from (and the latter just returns `None`, so we can we even drop it altogether if we want to be minimalistic)

swelborn:
in the previous `dev_mode` we were doing two different things:

1. allowing http in the url that could be either outside or inside of your local network
2. url validation to ensure the requests won't be pointing to local hosts. this still could be https

right? so i figure we need both, otherwise `allow_http` is just `dev_mode`, which would have to do both of those things

genematx:
maybe another config parameter under `webhooks:` then? `allow_local_targets: bool` or `skip_ssrf_check: bool` -- false by default (for prod), flexibly configurable (explicit and orthogonal) for dev? what do you think, @swelborn ?

swelborn:
perfect, I will implement it

swelborn:
how does that look

swelborn:
updated this PR again to work with the latest: https://github.com/carbonscott/tiled-catalog-broker/pull/54

genematx:
Looking good to me! thanks, @swelborn

danielballan (approved):
Awesome to have this.

swelborn:
thanks for the fast reviews from y'all ! nice discussion, let's keep this relationship going

note I do not have permissions to hit the green merge button, so someone else will have to do that. you will get all the satisfaction of clicking 🟩, too >:)

swelborn:
https://github.com/carbonscott/tiled-catalog-broker/pull/57 

putting this here in case anyone (on your team or in the future) interested in type of thing we can do with this

it simply receives events, persists them in redis streams (which come along with tiled compose anyway), and workers start pushing if the metadata contains `amsc_public`.

that PR look huge but it is only because I temporarily moved the catalog models to ensure we are using pydantic

Issue #1356, opened 2026-04-24
tests take quite a while...


swelborn:
i think some app setup for tests could be module scoped but we need to see what side effects could be.

could be intentional, but it does take like 15 minutes to run the entire suite

danielballan:
Yeah, this has been on my mental list for a bit. Thanks for opening an issue.

I'd like to make the internals more modular. The StreamingCache and soon the Webhooks... are nicely self-contained. Not so much the older components, which were created while we were still discovering the project's scope and architecture.

I've been wondering whether to do that first, and then revamp the tests. Maybe a little of both: do the low-hanging fruit in the test suite, then improve modularity, then revisit tests again.

PR #1359, opened 2026-04-24
[#1356] add pytest-xdist

swelborn:
trying this out to see if tests run faster this way

### Checklist
- [ ] Add a Changelog entry


swelborn:
These run about twice as fast on Linux runners (2 cores?). Runs significantly faster locally depending on your workstation. Could help reduce your runner time until a more permanent test restructure is in. it does change some things that I'm not sure you want to adjust.

danielballan:
I am open to running the experiment of adding this. I have been wary of the added complexity, but I wouldn't overrule trying this out. It certainly does speed up the iteration cycle!

swelborn:
i have not used it before tbh (never had to worry about so many tests) but it works.

there are various options for which point you bring parallelism (by test, by module, etc.)...

up to you, just thought i would put it out there. it could be better to hold off and considering _during_ the test refactor you mentioned in #1356, so that tests are refactor in a way that is compatible with multiprocessing

danielballan:
> it could be better to hold off and considering during the test refactor you mentioned in https://github.com/bluesky/tiled/issues/1356, so that tests are refactor in a way that is compatible with multiprocessing

By default I'll go with that idea (and aim to prioritize that refactor in the coming weeks) unless someone speaks up to advocate for merging this sooner.

danielballan:
I've come around on this. It works, locally for me and on CI. The pytest-xdist project has been around for a good long while.

If we get it in now, we can be sure that the coming refactor does not break parallelism.

danielballan:
To be continued in https://github.com/bluesky/tiled/issues/1388.

PR #1360, opened 2026-04-27
Added EntraAuthenticator

vshekar:
### Checklist
- [x] Add a Changelog entry
- ~[ ] Add the ticket number which this PR closes to the comment section~


tiled/authenticators.py
```diff
@@ -306,6 +307,23 @@ def oauth2_schema(self) -> OAuth2:
...
+        claims = super().decode_token(id_token, access_token)
+
+        # sub generated by Entra is an opaque string, generating sub UUID for
+        # tiled based on "iss|sub" for uniqueness
+        original_sub = claims.get("sub")
+        issuer = claims.get("iss", "")
+        claims["sub"] = uuid.uuid5(uuid.NAMESPACE_URL, f"{issuer}|{original_sub}").hex
+        claims["entra_sub"] = original_sub
+
+        # Translate entra scopes to tiled
```
danielballan: The name `access_as_user` is arbitrary: I believe Claude suggested it when I was configuring OBO in Azure. Also, we may want multiple sets of such scopes.

This should be a piece of required Authenticator configuration, e.g.:

```yaml
authenticators:
  providers:
  - provider: tiled.authenicators:EntraAuthenticator
    args:
      scopes_map:
        # Map Entra scopes to Tiled scopes.
        access_as_user: ["read:metadata", "read: data"]
```

danielballan (changes requested):
Looks good! Just need configurability.

danielballan:
I think we can accept the lack of test coverage, at least for now. We'll be exercising this in prod, so it's unlikely to rot, and it's challenging to securely test this on GH CI.

danielballan (approved):

danielballan (approved):

PR #1362, opened 2026-04-29
Refactoring AwkwardAdapter

genematx:
Minor refactoring of `AwkwardAdapter` and associated storage mechanisms without changes to the functionality.

1. `AwkwardBufferAdapter` and `AwkwardAdapter` are moved into the same module, `adapters.awkward` with the intention tha tthe first one would be used as a storage-based adapter and the latter one -- as an in-memory adapter. `.read` method returns an instance of `awkward.Array` (instead of contents buffers) to match the signature of similar adapters.

3. `DirectoryContainer` is moved to `tiled.storage`, so it can be more cleanly reused by other mimetypes that require container storage (e.g. derivatives of awkward, such as ragged).

Related: #1104 

### Checklist
- [x] Add a Changelog entry


tiled/storage.py
cjboyle, line 28: I don't think this is strictly needed, but doesn't hurt.
```suggestion
__all__ = [
    "DirectoryContainer",
    "EmbeddedSQLStorage",
    "RemoteSQLStorage",
    "FileStorage",
    "SQLStorage",
    "ObjectStorage",
    "Storage",
    "get_storage",
    "parse_storage",
]
```

cjboyle: (the change is prepending `"DirectoryContainer",`, not sure why the diff isn't showing it correctly)

cjboyle (approved):
It seems I can approve this, so I will approve this.

tiled/storage.py
genematx: Thank you, Connor! Makes sense to me

cjboyle:
```
__all__ = [
    "EmbeddedSQLStorage",
    "RemoteSQLStorage",
    "FileStorage",
    "SQLStorage",
    "ObjectStorage",
    "Storage",
    "get_storage",
    "parse_storage",
__all__ = [
    "DirectoryContainer",
    "EmbeddedSQLStorage",
    "RemoteSQLStorage",
    "FileStorage",
    "SQLStorage",
    "ObjectStorage",
    "Storage",
    "get_storage",
    "parse_storage",
]
```

well that didn't work at all how I expected...

genematx: fixed!

danielballan (approved):
This makes sense.

tiled/adapters/awkward.py
```diff
@@ -65,47 +53,84 @@ def from_array(
...
+            for key in self._structure.awk_form.expected_from_buffers()
             if (form_keys is None)
             or any(key.startswith(form_key) for form_key in form_keys)
         ]
-        buffers = {}
-        for key in keys:
-            buffers[key] = self.container[key]
-        return buffers
 
-    def read(self) -> Dict[str, bytes]:
```
danielballan, line 90: This API change introduces an asymmetry between `read` and `write`. While `write` accepts labeled buffers, `read` now returns an `ak.Array`. This _might_ be surprising, but it might overall be the best choice. Just underlining for a close look.

tiled/structures/awkward.py
```diff
@@ -11,6 +11,10 @@ class AwkwardStructure(Structure):
     length: int
     form: dict
 
+    @property
+    def awk_form(self) -> awkward.forms.Form:
```
danielballan: Cosmetic question: awkward typically abbreviates itself `ak`. Could consider `awkward_form` or `ak_form`. I'm fine with `awk_form` though, perhaps a little easier for newcomers to decode?

genematx: I also have awk_form in ragged... (especially in the structure). `awkward_form` is probably too long although might be the lesser of two evils (`ak_form` would be more confusing, I think).

tiled/adapters/awkward.py
genematx: good point. I've made `read` to return the actual `ak.Array` to match the pattern in other adapters (e.g. numpy arrays array are returned as arrays, not bytestreams); without the structure, awkward buffers are useless. I would think that `write` should also accept `ak.Arrays` (even though some of the information, namely `Form`, would also be duplicated in the structure).

danielballan (approved):

Issue #618, opened 2023-12-12
Build out support for service accounts in Tiled


danielballan:
At NSLS-II, we have a use case for service accounts in Tiled. Service accounts have been built into the database from the start, but there is no HTTP API for _creating_ one.

TO DO:

- [ ] Add an API endpoint for creating a service account
- [ ] Add Python API for creating a service account: `context.admin.create_service_principal(...)`
- [ ] Add CLI for creating a service account `tiled admin service-principal create ...`
- [ ] Add Python API for creating an API key on behalf of another user `context.admin.create_api_key(...)`
- [ ] Add CLI for creating an API key on behalf of another user `tiled admin api_key create ...`

All of the above are administrative functions. The server should require such requests to have `admin` scope.

<hr />

Reading material for context:

We probably want `POST /api/v1/auth/principal` to create a new Principal. This is the associated `GET` route:

https://github.com/bluesky/tiled/blob/5984ad1d454985b8c6550e08a2173fbe445ca072/tiled/server/authentication.py#L826-L856

This is a where we create a Principal (user, not service) when someone logs in for the first time:

https://github.com/bluesky/tiled/blob/5984ad1d454985b8c6550e08a2173fbe445ca072/tiled/server/authentication.py#L411

The code for `create_user`:

https://github.com/bluesky/tiled/blob/5984ad1d454985b8c6550e08a2173fbe445ca072/tiled/authn_database/core.py#L93-L113

PR #1363, opened 2026-05-01
Implement admin commands for service principals + API keys

cjboyle:
### Checklist
- [x] Add a Changelog entry

I think this mostly implements the rest of #618, but doesn't specifically address the concerns raised in #619/#620. I'm happy to make more changes to close out both issues if needed.


danielballan:
Nice, thanks!

danielballan (approved):
I took this for a test drive against our production instance. Let's get it in.

Issue #319, opened 2022-09-12
Add notion of multichannel images to array structure


danielballan:
This was inspired by a demo by @prjemian.

## Proposal

Add a new key (`channels`?) to ArrayStructure indicated the array axis containing (color) channels such as `RGB`, `RBG`, `RGBA`, `CMYK`, etc. This is also useful in the context of multichannel imaging where the channels aren't "colors" in the traditional sense but different signals.

This could be used to by some exporters---in particular, of course, the image format exporters like PNG---to produce color images. It would give all clients more context for how to interpret the data.

The details of how to spell this remain to be worked out.

## Alternatives considered

* **Just leave this implicit.** That leaves clients displaying color images as 3-frame stacks of monochrome images. Seems bad.
* **Leave it up to the `metadata`.** Image formats like TIFF and JPEG have different ways of parametrizing and spelling this. Tiled should provide a consistent view on it.
* **Put it in `dims`.** ArrayStructure already carries a `dims` attribute, an (optional) list of strings mean to label each dimension of the array. Currently, the content of `dims` has no meaning to the server. If we made some `dims` like `"RGB"` have special meaning, we would be entering new territory. For both `dims` and `metadata`, I think there is value is reserving these are "user space", whose content will never change the behavior of the server.

danielballan:
More thoughts on why it's important to expose color channels as part of _structure_:

* In a sense the "topology" of a multichannel image array is different than if it's a stack of 3 images.
* Formats like JPEG apply lossy compression to the structure in a way that is fundamentally tied to the human visual system. It wouldn't make sense to expose just any `(3, N, M)` array as JPEG, and the server should refuse to do it.

PR #1367, opened 2026-05-04
Fix color image display in WebUI array viewer

genematx:
ArrayND component was treating the last two dimensions as (H, W) for all arrays, which broke color images shaped (H, W, 3) or (H, W, 4) — the color channel was incorrectly used as the width axis.

Introduce isColorImage() heuristic: if the last dimension is 3 or 4, treat the last three dims as the image plane (H, W, C) and build sliders only over the remaining stack dimensions. The stride-based downsampling slice is also updated to stride only H and W, leaving the color channel whole (e.g. ::3,::3,:).

Add tests for RGB (H,W,3), stacked RGB (N,H,W,3), and large color image striding.

image

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ### Fixed
 
+- Display of color images in the web UI array viewer.
 - JSON serialization of tables now correctly handles numpy scalar types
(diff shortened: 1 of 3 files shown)


danielballan:
I'd like to eventually make color channels explicit in the structure. But this is a good default / stop-gap. https://github.com/bluesky/tiled/issues/319

PR #1369, opened 2026-05-05
Added EntraInternalAuthenticator

vshekar:
Wrapper around OIDCAuthenticator that extracts username from Entra token

### Checklist
- [ ] Add a Changelog entry


danielballan (changes requested):
Just noting here our plan to _not_ merge this but step back and consider what we want to do here.

PR #1373, opened 2026-05-06
Avoid printing error message when waiting for external authenticator

genematx:
In device flow, when waiting for the external authenticator, the expected `authorization_pending` 400 error message has not been properly caught and was being displayed in CLI during stamina retires.

### DIFF
--- a/tests/test_device_flow.py
+++ b/tests/test_device_flow.py
@@ -173,7 +173,6 @@ def test_device_flow_success(
 
         LCWE-ROXW
 
-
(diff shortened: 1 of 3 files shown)


danielballan (approved):
Thanks!

danielballan (approved):
Tested against `tiled-staging.nsls2.bnl.gov`. This works and removes the extraneous retries/warnings.

danielballan (approved):

PR #1375, opened 2026-05-06
Fix typo in `auth-callback` path.

genematx:
Fix a typo in the `/auth-callback` path in WebUI.


danielballan (approved):

danielballan:
No point in waiting on Python unit tests for this Javascript-only change.

PR #1382, opened 2026-05-12
Enable Downstream OBO Flow

genematx:
Ensures that entra_access_token is stored in state dict and passed to downstream services that rely on OBO with Tiled (e.g. chat app).

### Checklist
- [ ] Add a Changelog entry


danielballan (approved):
Merge at will when CI passes

PR #1385, opened 2026-05-12
More OIDC follow-up

danielballan:
More follow up to #1381.

- In `main`, everyone is treated as a `Role.user`. This merges in `Role.admin` scopes when appropriate.
- Carry the access token in the Principal model (needed for DLS' authorization code and may be useful in general).
- Do not track sessions in Tiled's AuthN database when the sessions are generated externally.


danielballan:
Windows tests that failed are known flaky (CLI subprocess) and do not involve AuthN.

PR #1386, opened 2026-05-13
Fix using TILED_API_KEY to authenticate CLI

cjboyle:
w.r.t. the discussion in the dev call on Monday, it turns out the CLI already does support the use of `TILED_API_KEY`, but it was being ignored.

### Checklist
- [ ] Add a Changelog entry


danielballan (approved):

danielballan:
Ha! Thanks @cjboyle.

Issue #163, opened 2022-01-31
Support Moves and "Tombstone" endpoints.


danielballan:
I propose to add two special sentinel objects that can be placed in tiled trees:

* `Moved` which prompts the server issue a [`301 Moved Permanently`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) response that sends the client to a different location in the tree
* `Tombstone` which indicates that there used to be something here but that it has been removed, likely permanently, by issuing a [`410 Gone`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410) response.

### Use case

As we iterate on various Tiled deployments, we are likely to want to add levels of nesting and rename things from time to time. It would be best to avoid breaking user code during this phase of finding the right spellings for things. It would especially important to avoid invalidating their client-side caches, which might have cost some time (and bandwidth usage) to populate.

Note that this not the same as an _alias_ or a _soft-link_. There is still only one canonical location for the data, and client would be told explicitly to retry at the new new location, rather than transparently shown the data from a different location.

### Proposed implementation

The response to `/node/metadata/{path/to/moved/item}` should be a `301` with `Location: /node/metadata/{path/to/current/location}`. A client trying to refresh its cache or accessing a specific item by name will therefore be forwarded.

In `/node/search` items that are `Moved` should be omitted. That is, the former location of a node should not clutter the current namespace.

### Use case

The DOI standard has a concept of [tombstone pages](https://support.datacite.org/docs/tombstone-pages) placeholders for resources that have been removed. The reason for removal may be a retraction; or that no one is wiling to pay to maintain the resource anyone; or that it was accidentally lost. The goal is to communicate to the client: "You are probably in the right place, but what you are looking for is not available, likely permanently."

### Proposed implementation

In `/node/search` items that are `Tombstone` _should_ appear. They MAY carry the `metadata` of the original (or some portion of it) so as to be searchable.

The response to `/node/metadata/{path/to/moved/item}` should be a `410`. Perhaps the body should include an optional reason and an optional link to where more information can be found.

swelborn:
Had a slack convo with @danielballan about adding some state to a node to make it `Archived` - moved to tape or some colder storage.

This is a distinct case from `Tombstone` or `Moved` - we do not want to create a new node (as in `Moved`) or say that it no longer exists, possibly forever (as in `Tombstone`). We want to be able to keep the metadata information on the node, showing that it is in colder storage, and could be warmed up.

[...]

PR #1387, opened 2026-05-13
graph of links

dylanmcreynolds:
This implements a sql-based graph-like linking implementation. This is *EXPERIMENTAL*. Like, really quite experimental.

With #242, @danielballan implemented a linking mechanism to implement compatibility for bluesky references. 
We [discussed](https://github.com/bluesky/tiled/pull/242#issuecomment-1171363735) a more general purpose linking mechanism, which might even be able to produce JSON-LD and/or RDF. These specifications are widely used to provide semantic meaning to links between objects on the web. These technologies are often implemented in purpose-built "tripl store" databases, which are highly optimized for indexing and searching a triple, often described as subject, predicate object. Specialized query languages like SPARQL exist for querying triple stores. Triple stores are a specialization of graph database, with the `predicate` providing the semantic indicator of the link.

Another popular feature of triple stores to actually quad stores, with another dimension helping support groups of triples, imagine projects, for example

This PR is inspired by triplestores, but does not implement many of the features. It does this all in SQL, with lots of indexes. It adds a third database to the tiled family (in addition to catalog and auth). In that database are a couple of key t tables:
- `entities` which carry URIs and metadata blocks
- `links` which have columns for `subject`, `object`, and `predicate`

The PR exposes this new functionality using a GraphQL endpoint, rather than a RESTful endpoint. This is a departure for tiled, and something that should be considered and debated. I did it this way because it felt more natural for queries here. 

Much of this code comes from another project: [splash_links](https://github.com/als-computing/splash_links). It could easily stay as a separate service, but in discussions with @ahexemer and @danielballan, there was a strong desire to make it a core feature of Tiled. We imagine this helping with several key types of use cases:

- Applications that are storing processed data in tiled, and want to link that data to it's raw data...or processed to other processed.
- Information from multiple proposals or multiple instruments that naturally fit under a single "experiment", "project", or "campain"
- Links between nodes in tiled and resources external to tiled...or other tiled servers

### Checklist
- [x] Add a Changelog entry
- [x] Unit tests
- [x] Implement access_tags to support access control enforcement
- [x] User Documentation
- [x] Export links as JSON-LD
~~Support predicate sub-classing with another table (FUTURE)~~

Full disclosure, much of this PR was assited by an LLM (Claude). To the extent possible, all changes were reviewed before being accepted.


dylanmcreynolds (commented):
I have added:
- Access Controls
- Serverside import/export of JSON-LD
- Added a namespace table to store namespaces, which is useful for JSON-LD support
- An examples folder which spins up a tiled server and creates datasets and [RO Crate JSON-LD links](https://www.researchobject.org/ro-crate/)

tiled/catalog/migrations/versions/c31f6a1d7e20_add_graph_entities_and_links_tables.py
```diff
@@ -0,0 +1,84 @@
+"""Add graph entities and links tables
```
danielballan: This is migration script for the catalog database. Will the new tables `entities` and `links` be added to the catalog database or to a separate PG database?

My recollection of the plan was that the entities and links would live in their own database (and `initdb/01-create-databases.sh` creates one) with the idea that the database technology used might evolve later. Did we pivot to putting entities and links in the catalog database in order to keep referential integrity of foreign keys to nodes in tiled?

I can see it either way, but we should be consistent: either update `initdb/01-create-databases.sh` to drop the extra database or rework this to be separate from the catalog.

pixi.toml
```diff
@@ -74,6 +73,9 @@ sphinx_rtd_theme = "*"
 sphinxcontrib-mermaid = "*"
 types-cachetools = "*"
 
+[feature.dev.pypi-dependencies]
+pre-commit = ">=4.6.1, <5"
```
danielballan: Why moved from conda-forge to PyPI? Version 4.6.1 is [available](https://anaconda.org/channels/conda-forge/packages/pre-commit/overview).

pixi.toml
```diff
@@ -112,6 +114,9 @@ starlette = ">=0.48.0"
 uvicorn = "*"
 zarr = "*"
 
+[feature.server.pypi-dependencies]
+strawberry-graphql = {version = ">=0.315.3", extras = ["fastapi"]}
```
danielballan: In pixi, prefer conda pacakges where available and only fall back to PyPI wheels when necessary. Strawberry-graphql is [available](https://anaconda.org/channels/conda-forge/packages/strawberry-graphql/overview).

dylanmcreynolds (commented):
> This is migration script for the catalog database. Will the new tables `entities` and `links` be added to the catalog database or to a separate PG database?
> 
> My recollection of the plan was that the entities and links would live in their own database (and `initdb/01-create-databases.sh` creates one) with the idea that the database technology used might evolve later. Did we pivot to putting entities and links in the catalog database in order to keep referential integrity of foreign keys to nodes in tiled?
> 
> I can see it either way, but we should be consistent: either update `initdb/01-create-databases.sh` to drop the extra database or rework this to be separate from the catalog. 

I thought we pivoted so that we could get the fastest queries for joining `entities` and `nodes`. But we did go back and forth a bit.

dylanmcreynolds (commented):
Fixed the pixi.toml pypi to conda things.

pixi.toml
dylanmcreynolds: Fixed both pixi.toml things.

dylanmcreynolds:
Once of the unit tests is failing...is this a periodic thing?

```
-------------------------------------------------------------------------------
TOTAL                                             17671   3332    81%
Coverage XML written to file cov.xml
=========================== short test summary info ============================
FAILED tests/test_simple_server.py::test_readable_storage[False-True] - httpx.ReadTimeout: timed out
= 1 failed, 1969 passed, 18 skipped, 17 xfailed, 100 warnings in 1126.73s (0:18:46) =
Error: Process completed with exit code 1.
```

danielballan:
OK, agreed. It makes sense to prioritize referential integrity of the foreign keys and fast joins. Can you just remove that line added to `initdb/01-create-databases.sh` then, to avoid creating an unused database?

Yes, there seem to be a couple tests in `test_simple_server.py` that have recently become flaky. Needs investigation, can be ignored for the purposes of this PR.

tacaswell:
I have not read the code so this is a conceptual review.

------

Reflecting on both why analysis databroker never took off and what we need for HRD and QCT in the near future, I think that a first-class built-in way to track the provenance of the reduced and analyzed data is core to what tiled needs to do.

Given how core it is, I expect that if tiled does not support this then everyone will implement a half-baked 80% solution in metadata + client code.

dylanmcreynolds:
The json-ld import export still needs a lot of work. It currently doesn't have  a filtering or query mechanism. So, as it is right now, it's not safe to merge (you'd dump all graphs if you called it.) I'm not quite sure what the query mechanism should actually be. If we want to merge soon, I propose commenting the code out that is figured out.

docs/source/user-guide/graph-and-links.md
```diff
@@ -0,0 +1,326 @@
...
+
+Tiled can optionally serve a graph of entities and links alongside a
+catalog-backed tree, queryable through a GraphQL API. See
+{doc}`../explanations/graphs` for background on what this feature is and why
+it exists.
+
+This guide walks through starting a demo server with the graph enabled and
+exploring it interactively in the browser.
+
+## Enable the graph feature
```
genematx, line 13: I wonder if it would be helpful to be a bit more specific with terminology here; e.g. instead of just "graphs", maybe we could refer to this new feature as "graph of links" or "entity graph" or "triple store graph" or anything else more specific. In this case, if (when) we would like to add a `GraphStructure` for example, if would be easier to avoid any collisions.

docs/source/user-guide/graph-and-links.md
```diff
@@ -0,0 +1,326 @@
...
+The graph is available automatically whenever a server is serving a
+catalog-backed tree (see {doc}`example-server-config`)---there is no separate
+configuration flag. A ready-to-run demo lives in
+`example_configs/graphs/` in the Tiled source repository:
+
+```
+bash example_configs/graphs/run_demo.sh
+```
+
+This starts a server, seeds a small catalog of datasets, creates a handful of
```
genematx, line 24: perhaps, we could reuse the same demo example that already exists (i.e. `tiled serve demo`), expanding it with the relationship graph (and maybe adding a few nodes to make it more interesting), especially since there is no separate turn on/off switch for this feature, and it is always available?

genematx: this might require some rework of our demo example, but I think it would be nice to have a simple single way to run a comprehensive demo.

docs/source/user-guide/graph-and-links.md
```diff
@@ -0,0 +1,326 @@
...
+  }
+  ```
+
+  `path` is the list of key segments leading to the node (`["raw_dataset"]`
+  for a top-level entry, `["a", "b"]` for a nested one). It returns `null` if
+  no such node exists.
+
+- **`uri`** --- a free-form locator, stored and returned verbatim with no
+  lookup or validation. Follow this convention when setting it:
+  - If the entity points at data hosted by *this* server, set `uri` to the
```
genematx, line 245: Is this how one might link a _slice_ of an array -- i.e. provide both `nodeId` and the full uri with appropriate query parameters? Or should those query parameters go into properties on the Entity?

docs/source/user-guide/graph-and-links.md
```diff
@@ -0,0 +1,326 @@
...
+  }
+}
+```
+
+```json
+{
+  "input": {
+    "entityType": "dataset",
+    "name": "my_dataset",
+    "properties": { "schema:encodingFormat": "application/x-zarr" }
```
genematx: is encodingFormat a required property, or is it just an example? How does it relate to mimetypes of the underlying Tiled's data sources?

tiled/graph/schema.py
```diff
@@ -0,0 +1,594 @@
...
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# JSON scalar — pass arbitrary dicts / lists / primitives through GraphQL
+# ---------------------------------------------------------------------------
+
+JSON = StrawberryJSON
+
+
+# ---------------------------------------------------------------------------
```
genematx: A minor stylistic nitpick... From what I can tell, we tend not to use section breakers like these. Might be worth a decision on how we want to move forward (personally, I can see pros and cons: they could be informative, but they also take space and add noise). I've noticed, Claude does love them, especially in tests.

docs/source/user-guide/graph-and-links.md
genematx: although, I do like the short import statements, `from tiled.graph ...`

tiled/graph/store.py
```diff
@@ -0,0 +1,433 @@
+"""
+Storage layer for the splash-links entity graph service.
```
genematx: might be worth adding some details about what "splash links" are in the docs.

tiled/graph/__init__.py
genematx, line 1: should we rename this submodule to `tiled/graphql` maybe?

genematx:
Thank you, Dylan! This is super exciting!
I've added a few comments and made some changes [here](https://github.com/als-computing/tiled/pull/1) (mostly around the provisioning of the tables and fixing some minor bugs).
I wonder if we'd like to integrate it even closer with the catalog orm.

tiled/graph/schema.py
```diff
@@ -0,0 +1,594 @@
...
+    ) -> list[Entity]:
+        records = await _store(info).list_entities(
+            entity_type=entity_type, limit=limit, offset=offset
+        )
+        records = await _apply_policy_filters(info, records, "read:metadata")
+        namespaces = await _namespaces(info)
+        return [
+            _entity_from_record(r, namespaces)
+            for r in records
+            if await _is_allowed(info, r.access_blob, "read:metadata")
```
genematx: This is potentially a user-facing bug: the access filter is applied _after_ pagination, which may change the number of returned results.

docs/source/user-guide/graph-and-links.md
dylanmcreynolds: How about "semantic graph"?

docs/source/user-guide/graph-and-links.md
dylanmcreynolds: yeah, I like the idea. Maybe a future PR?

docs/source/user-guide/graph-and-links.md
genematx: I like it. Especially in docs!
For the module name, i thought maybe `tiled.graphql` could be an option, but I'm not sure -- don't quite understand yet how it is specific to only GraphQL only, or if it's more general.

docs/source/user-guide/graph-and-links.md
genematx: yeah, sounds good! I'm already thinking about a simple "getting-started"/"how-to script" - can be done together.

docs/source/user-guide/graph-and-links.md
danielballan: I like that `tiled.graph` addresses the key functionality without necessarily bolting us to (only) GraphQL as the interface.

tiled/catalog/migrations/versions/c31f6a1d7e20_add_graph_entities_and_links_tables.py
dylanmcreynolds: Same database. The extra database from `initdb/01-create-databases.sh` was cruft from when I was doing it as a separate database. It's been removed. Thanks for the catch.

docs/source/user-guide/graph-and-links.md
dylanmcreynolds: For simplicity, I've removed the property.

tiled/graph/__init__.py
dylanmcreynolds: I don't think so. In the next PR, I hope to add a restful endpoint to aid in importing/exporting graphs as json-ld.

tiled/graph/schema.py
dylanmcreynolds: nice catch, removed

tiled/graph/schema.py
dylanmcreynolds: You are right! And looking at this, that's also a reduntent check. I'll remove it, but it doens't address your comment, which still applies in the `_apply_policy_filters` call.

Fixing that is going to require some surgery.

dylanmcreynolds: I updated how the querying is done, better matching how it's done in the tiled catalog for notes.

tiled/graph/store.py
dylanmcreynolds: removing

danielballan:
CI failures are unrelated, known-flaky (somewhat recently...gotta deal with these), and just passed on the last run.

danielballan (approved):

dylanmcreynolds (commented):
Met with Dan and Eugne today to talk about the convern that `access_blob` of `entities` is completely separate from that of `nodes`. 

We came up with the idea that when the `entity` points to a `node`, that the access control rules for the node determine access control for the entity. So we want to do two things:
* add a database trigger to enforce that if entity has a node_id, that access_blob is NULL
* add logic to the access_control checks to use the logic of the node for the entity

dylanmcreynolds:
added a database trigger to enforce that if entity has a node_id, that access_blob is NULL
added logic to the access_control checks to use the logic of the node for the entity

tiled/catalog/migrations/versions/c31f6a1d7e20_add_graph_entities_and_links_tables.py
```diff
@@ -71,17 +74,76 @@ def upgrade():
         unique=False,
     )
 
-    op.create_table(
```
genematx: is there a reason this has been dropped from the migration script? I think we need this statement to mirror the orm.

tiled/catalog/migrations/versions/c31f6a1d7e20_add_graph_entities_and_links_tables.py
```diff
@@ -71,17 +74,76 @@ def upgrade():
...
+BEFORE INSERT OR UPDATE ON entities
+FOR EACH ROW
+WHEN (NEW.node_id IS NOT NULL AND NEW.access_blob IS NOT NULL)
+EXECUTE FUNCTION entities_reject_node_access_blob();"""
+            )
+        )
 
 
 def downgrade():
-    op.drop_table("namespaces")
```
genematx: same as above

tiled/graph/store.py
```diff
@@ -0,0 +1,473 @@
...
+        properties: Optional[dict] = None,
+        access_blob: Optional[dict] = None,
+    ) -> EntityRecord:
+        id_ = str(uuid.uuid4())
+        now = datetime.now(timezone.utc)
+        async with self._engine.begin() as conn:
+            await conn.execute(
+                insert(_entities).values(
+                    id=id_,
+                    node_id=node_id,
```
genematx, line 202: should we be checking if the principal who tries to create an Entity associated with a given node has access to this node at all? (maybe it's done somewhere else implicitly that I didn't notice)

tiled/graph/orm.py
```diff
@@ -0,0 +1,162 @@
...
+metadata = Base.metadata
+
+entities = Table(
+    "entities",
+    metadata,
+    Column("id", String, primary_key=True),
+    Column(
+        "node_id",
+        Integer,
+        ForeignKey("nodes.id", ondelete="SET NULL"),
```
genematx: should this be `ondelete="CASCADE"` instead? I.e. when we delete a Node, there's probably no reason to keep the entity, but I might be misunderstanding the intention behind it.

dylanmcreynolds: I sort of think that in that case the Entity should stay, as a sort of tombstone. But I'm not super strong in that opinion.

tiled/graph/store.py
dylanmcreynolds: Could we accomplish this by copying the `access_blob` from the node to the entity?

genematx: I don't think we need to _copy_ the access blob, just to check that the user has access to the node they are trying to use (i.e. can a user who can not see a node, but who guessed its uid, create an entity related to that node?). I think this is a very minor issue really; guessing the uid wouldn't be easy... Maybe to avoid complicating things, we can just leave it as is for now?

tiled/catalog/migrations/versions/c31f6a1d7e20_add_graph_entities_and_links_tables.py
dylanmcreynolds: fixed

tiled/catalog/migrations/versions/c31f6a1d7e20_add_graph_entities_and_links_tables.py
dylanmcreynolds: fixed

tiled/graph/orm.py
danielballan: I see potential ambiguity there. When `node_id` becomes the `NULL`, the Entity now looks like an _external_ entity. It could be ambiguous which entities are external and which are internal tombstones.

Could to switch to `ondelete="CASCADE"` in this first draft and discuss tombstones for a future revision? It might be best to approach tombstones for Entities and [tombstones for nodes themselves](https://github.com/bluesky/tiled/issues/163) holistically.

dylanmcreynolds: The only issue I have is that it's slightly inconsistent with entities that refer to external links, where items could be tombstoned without us even knowing. I'll switch to CASCADE for now but I think this is a bigger discussion.

PR #1396, opened 2026-05-27
Clean-Up Logging

genematx:
1. Decode percent-encoded characters in server and client logs for readability.
2. Suppress stamina-generated messages, e.g. `Scheduled retry in 0.28 seconds due to ConnectError('[Errno 61] Connection refused') (attempt 1)` by default; they can be shown with `show_logs()`.

```
In [9]: arr[:20, 10:50:2, -30:]
11:33:35.959 ->  GET 'http://localhost:8000/api/v1/array/full//e2b4bfb8-d41c-4a14-bf0a-48d0ca8a3ea4?expected_shape=20,20,30&slice=:20:1,10:49:2,270:300:1' 'host:localhost:8000' ...
```

```
[1b801a92d73b3177] 127.0.0.1:59031 (singleuser) - "GET /api/v1/array/full//e2b4bfb8-d41c-4a14-bf0a-48d0ca8a3ea4?expected_shape=20,20,30&slice=:20:1,10:49:2,270:300:1 HTTP/1.1" 200 OK
```

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/client/utils.py
+++ b/tiled/client/utils.py
@@ -1,4 +1,5 @@
 import builtins
+import logging
 import os
 import uuid
 from collections import defaultdict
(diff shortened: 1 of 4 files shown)


tiled/client/utils.py
```diff
@@ -144,25 +145,21 @@ def polling_retry_context(timeout: float):
...
-    Returned hook logs scheduled retries at *log_level*.
-
-    .. versionadded:: 23.2.0
+    Returned hook logs scheduled retries at *log_level* on the ``tiled.client``
+    logger.  This means retry messages are hidden by default and only shown
+    when the user calls :func:`tiled.client.logger.show_logs`.
     """
-    import logging
-
-    logger = logging.getLogger("stamina")
```
danielballan, line 158: Because this is configured globally, i.e.:

```python
LoggingOnRetryHook = stamina.instrumentation.RetryHookFactory(init_retry_logging)
stamina.instrumentation.set_on_retry_hooks([LoggingOnRetryHook])
```

...I liked the approach of using a `stamina` logger. Other libraries in the same Python process might be using stamina, and we should not put their retry logs into the `tiled.client` logger.

What are the trade-offs of using `show_logs()` and `hide_logs()` to adjust the log level of both `tiled.client` and `stamina`? Maybe what's proposed here is the best option after all, but I'm not sure.

genematx: good point. I'll refactor it to separate any third-party libraries retries from ours.

danielballan (approved):

PR #1397, opened 2026-05-27
Patch object used for generated CLI docs

danielballan:
### Checklist
- [ ] Add a Changelog entry


danielballan:
Of the pacakges involved, [typer just issued a new release](https://pypi.org/project/typer/#history) so I think that's where the new typer / sphinx-click incompatibility arose.

PR #1400, opened 2026-05-28
Improve Client Interactivity: Retries and Progress Bar

### LINKED ISSUE #1398: Client should respect `429 Too Many Requests`
I anticipate configuring Load Balancers in front of Tiled to throttle clients with `429 Too Many Requests`. The Tiled Python client should respond to a 429 response by retrying after the interval specified by the `429` response.

Stamina now facilitates this: https://github.com/hynek/stamina/pull/125

That is, a retry interval specified by the server can override the client-configured back-off interval.

---

I don't foresee any changes to the Tiled _server_ for this work. I expect that throttling is a job for Load Balancers, not application servers.

genematx:
## Summary

This PR improves tiled's client-side retry behavior, progress reporting, and connection management. It addresses several issues that affected interactive use in IPython/Jupyter.

---

### 1. Concurrency throttling

- Adds a `threading.Semaphore` (default: 16) to `Context` to cap concurrent HTTP connections during dask computes
- Exposes `max_connections` parameter on `Context`, `from_uri`, `from_app`, and `from_any_uri`

### 2. Rich progress bar for multi-chunk reads

For interactive sessions (`show_progress=True` and terminal detected), a transient Rich progress bar is shown during `.read()` calls on multi-chunk arrays, dataframes, and xarray datasets.

- `tracking_progress(context, total)`: context manager that sets `context._progress_state` to a `_ProgressState` instance during a tracked fetch; fetch methods advance the bar per completed request
- Applied to: `ArrayClient.read()`, `DataFrameClient.read()`, `DatasetClient.read()`, `CompositeClient.read()`
- No-op when: `total <= 1`, `show_progress=False`, non-interactive session, or an outer bar is already active (nesting prevention)
- `show_progress` defaults to `True`; opt out via `Context(show_progress=False)` or `TILED_SHOW_PROGRESS=0`
- Unpickled contexts (dask workers) always have `show_progress=False`

### 3. Retry indicator

A spinner appears when retries are in progress and disappears on success or failure — without any attempt count or error details (those remain at `DEBUG` in the logger).

**During a multi-chunk read (progress bar active):** A `⠋ Retrying…` spinner appears on the line below the progress bar, disappearing when the retry resolves.

**Without a progress bar** (e.g. `from_uri()` initial connection, single-chunk reads): A standalone transient spinner is shown.

Implementation:
- `_LoggingAttempt` now accepts an optional `context` argument; on each retry it calls `_signal_retry(context)`
- `retry_context(context=None)` passes `context` through to `_LoggingAttempt`; all `Context` methods and data-fetch methods pass `self` / `self.context`
- `_signal_retry_resolved(context)` is called in a `finally` block after every retry loop to ensure the indicator is always cleaned up — even when retries are exhausted or a non-retryable error propagates
- `_ProgressState.show_retrying()` / `hide_retrying()` manage the spinner-below-bar case
- `_StandaloneRetryIndicator` manages the standalone spinner; only operates from the main thread to avoid conflicts with IPython's display machinery

### 4. Ctrl-C cancels retries

Previously, `KeyboardInterrupt` during stamina's inter-attempt sleep was swallowed by tenacity's retry loop, requiring multiple Ctrl-C presses and leaving stale retry indicators.

- `retry_context()` now manually drives the stamina iterator via `next()` in a `while True` loop with an explicit `except KeyboardInterrupt` clause
- On Ctrl-C: calls `_signal_retry_resolved` to clean up any visible indicator, then re-raises immediately

### 5. HTTP 429 Too Many Requests support (closes #1398)

- `handle_error()` lets 429 pass through as a plain `HTTPStatusError` so stamina can intercept it
- `should_retry()` returns the `Retry-After` header value as a `float` (requires stamina ≥ 25.1.0), which stamina uses as the exact retry wait; falls back to default backoff if header is absent
- `Content-Type` header access uses `.get()` to avoid `KeyError` from proxy responses without that header

---

Screenshot 2026-05-29 at 11 09 59 AM
Screenshot 2026-05-29 at 11 11 08 AM

Screenshot 2026-05-29 at 2 31 35 PM

---

## Tests

32 tests in `tests/test_client.py` covering:

- Stamina logging isolation (3 tests)
- Semaphore concurrency throttling (2 tests)
- `tracking_progress` state, nesting, and no-op behavior (3 tests)
- `show_progress` env var and explicit override (2 tests)
- `_ProgressState` show/hide retrying, no NameError for Group import (2 tests)
- `_StandaloneRetryIndicator` show/reset lifecycle (1 test)
- `_signal_retry` routing to progress state vs standalone indicator (3 tests)
- `retry_context` signals retry indicator on retry (1 test)
- Ctrl-C cancels retries and cleans up indicator (2 tests)
- 429 retry: unit tests for `should_retry` and `handle_error`, integration test with mock server (7 tests)

---

## Checklist

- [x] Add a Changelog entry

Closes #1398

### DIFF
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -24,7 +24,6 @@ from .transport import Transport
 from .utils import (
     DEFAULT_TIMEOUT_PARAMS,
     MSGPACK_MIME_TYPE,
-    _signal_retry_resolved,
     handle_error,
     polling_retry_context,
     retry_context,
@@ -318,20 +317,17 @@ class Context:
         # (2) Let the server set the CSRF cookie.
         # No authentication has been set up yet, so these requests will be unauthenticated.
         # https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#double-submit-cookie
-        try:
-            for attempt in retry_context(self):
-                with attempt:
-                    server_info = handle_error(
-                        self.http_client.get(
-                            self.api_uri,
-                            headers={
-                                "Accept": MSGPACK_MIME_TYPE,
-                                "Cache-Control": "no-cache, no-store",
-                            },
-                        )
-                    ).json()
-        finally:
-            _signal_retry_resolved(self)
+        for attempt in retry_context(self):
+            with attempt:
+                server_info = handle_error(
+                    self.http_client.get(
+                        self.api_uri,
+                        headers={
+                            "Accept": MSGPACK_MIME_TYPE,
+                            "Cache-Control": "no-cache, no-store",
+                        },
+                    )
+                ).json()
         self.server_info: About = TypeAdapter(About).validate_python(server_info)
         self.api_key = api_key  # property setter sets Authorization header
         self.admin = Admin(self)  # accessor for admin-related requests
--- a/tiled/client/context.py
+++ b/tiled/client/context.py
@@ -2,6 +2,7 @@ import getpass
 import os
 import re
 import sys
+import threading
 import time
 import urllib.parse
 import warnings
@@ -30,6 +31,10 @@ from .utils import (
 
 USER_AGENT = f"python-tiled/{tiled_version}"
 API_KEY_AUTH_HEADER_PATTERN = re.compile(r"^Apikey (\w+)$")
+# Default cap on simultaneous outgoing HTTP connections and in-flight
+# data-fetch requests.  Keeps spike loads on the server bounded while
+# still allowing reasonable parallelism from dask's threaded scheduler.
+MAX_CONCURRENT_CONNECTIONS = 10
 
 
 def raise_if_cannot_prompt():
@@ -173,6 +178,7 @@ class Context:
         verify=True,
         app=None,
         raise_server_exceptions=True,
+        max_connections=MAX_CONCURRENT_CONNECTIONS,
     ):
         # The uri is expected to reach the root API route.
         uri = httpx.URL(uri)
@@ -224,9 +230,13 @@ class Context:
             timeout = httpx.Timeout(**DEFAULT_TIMEOUT_PARAMS)
         if cache is UNSET:
             cache = None
+        limits = httpx.Limits(
+            max_connections=max_connections,
+            max_keepalive_connections=max_connections,
+        )
         if app is None:
             client = httpx.Client(
-                transport=Transport(cache=cache),
+                transport=Transport(cache=cache, limits=limits),
                 verify=verify,
                 timeout=timeout,
                 follow_redirects=True,
@@ -257,8 +267,6 @@ class Context:
             # We are abusing it slightly to enable interactive use of the
             # TestClient.
 
-            import threading
-
             # The threading module has its own (internal) atexit
             # mechanism that runs at thread shutdown, prior to the atexit
             # mechanism that runs at interpreter shutdown.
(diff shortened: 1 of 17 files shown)


tiled/client/utils.py
```diff
@@ -484,3 +535,201 @@ def slices_to_dask_chunks(slice_dict, shape):
     )
 
     return dask_chunks
+
+
+def _is_interactive():
+    """Return True when running in an interactive Python session (REPL, IPython, Jupyter)."""
```
danielballan, line 540: To avoid `builtins.__dict__` consider:

```python
import importlib.util
if importlib.util.find_spec("IPython"):
    # IPython is installed.
    from IPython import get_ipython
    if get_ipython():
        # This Python process is an IPython process.
        return True
```

tiled/client/context.py
```diff
@@ -291,13 +292,32 @@ def __init__(
         # This is intentionally separate from the HTTP connection pool limit so
         # that it can be tuned independently; by default it mirrors the pool size.
         self._concurrent_request_semaphore = threading.Semaphore(max_connections)
+        # Slot used by tracking_progress(). Set to a (Progress, task_id) pair
+        # while a tracked .compute() is running; None at all other times.
+        # Assignment of a Python reference is atomic, so dask worker threads
+        # can safely read this without a lock.
+        self._progress_state = None
```
danielballan, line 302: As a general guideline, internal attributes named `_*` should only be touched by methods on the class they belong to. This is touched in many places by client classes, e.g. `self.context._progress_state`.

Options:
1. Ignore the problem. It's Python; we can do what we want.
2. Make it public.
3. Keep it internal, but add public methods that enable constrained public access to this internal state, such as:

```python
class Context:
    @property
    def progress_state(self):
        # public read-only access
        return self._progress_state

    @contextmanager
    def tracking_progress(self, total):
        ...
```

where your `tracking_progress` code gets moved from a function that takes `context` as its first argument to a method on `context`. This better communicates that `tracking_progress` may manipulate the Context's internal state (`_progress_state`).

As you can tell I prefer (3) but I could accept any of the above.

tiled/client/context.py
```diff
@@ -291,13 +292,32 @@ def __init__(
...
+        if show_progress is None:
+            show_progress = os.getenv(
+                "TILED_SHOW_PROGRESS", "1"
+            ).strip().lower() not in (
+                "0",
+                "false",
+                "no",
+            )
+        self.show_progress = show_progress
+        self._retry_indicator = (
```
danielballan: Similar to `_progress_state`, this is a supposedly internal variable that is touched by code external to the class. In this case I might suggest exposing public methods like `set_retry_indicator` and `clear_retry_indicator`.

danielballan (commented):
Awesome! Some thoughts in line.

tiled/client/context.py
genematx: yes! I agree; I've noticed this too. I like the suggested refactoring, will implement.

tiled/client/utils.py
genematx: Done!

tiled/client/context.py
genematx: Done.

tiled/client/context.py
genematx: Done

danielballan (approved):
I took it for a test drive. Awesome.

Thanks for address the comments; this looks solid.

PR #1402, opened 2026-06-01
Ensure Row Ordering in SQLAdapter

genematx:
This ensures the order of rows read from SQLAdapter is deterministic. It can be controlled by the new `order_by_args` parameter (list of `{"column": ..., "direction": ...}`). Additionally, setting `primary_key` to a list of column names would enforces the uniqueness of their values (this does not affect already created tables to ensure back-compatibility).

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/adapters/sql.py
+++ b/tiled/adapters/sql.py
@@ -73,6 +73,30 @@ class _OrderByArgs(TypedDict):
     direction: Literal["asc", "desc"]
 
 
+def _is_orderable_type(arrow_type: pyarrow.DataType) -> bool:
+    """Return True if the Arrow type can be used in a SQL ORDER BY clause.
+
+    Nested types (list, struct, map) cannot be sorted in SQL.
+    """
+    return not (
+        pyarrow.types.is_list(arrow_type)
+        or pyarrow.types.is_large_list(arrow_type)
+        or pyarrow.types.is_fixed_size_list(arrow_type)
+        or pyarrow.types.is_struct(arrow_type)
+        or pyarrow.types.is_map(arrow_type)
+    )
+
+
+def _is_primary_key_type(arrow_type: pyarrow.DataType) -> bool:
+    """Return True if the Arrow type is suitable for use as a primary key column.
+
+    In addition to the non-orderable nested types, floating-point columns are
+    rejected: NaN != NaN semantics and precision issues make float primary keys
+    almost always a mistake.
+    """
+    return _is_orderable_type(arrow_type) and not pyarrow.types.is_floating(arrow_type)
+
+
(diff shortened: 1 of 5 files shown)


tiled/adapters/sql.py
```diff
@@ -82,6 +82,8 @@ def __init__(
         table_name: str,
         dataset_id: int,
         *,
+        order_by_column: Optional[str] = None,
```
danielballan: I think that in the most common case a single column (e.g., `time`) will be sufficient to uniquely order the dataset. But there could be future cases where we want a series of order by clauses. Consider building that into the schema from the start by taking a list of `(name, direction)` paris like `[("time", 1)]`. This happens to match the client sort API (which in turn is inspired by the pymongo API).

danielballan (commented):
One structural question.

tiled/adapters/sql.py
genematx: yeah, I agree. I fact, this was my initial approach, to make it more general. but then I thought maybe it would complicate the parameters too much (passing a single string is easy) for very little benefit. I'm not sure... Maybe we can accept both.

genematx: perhaps we can expand this in the future, if needed and if the ordering proves useful?

danielballan: I'm torn on this one. It's fair point: extending it later with back-compat (i.e., support for both) isn't that bad.

But supporting ordering without a specified direction (ascenending/descending) feels to me like a significant gap. And if we need `{"column": "X", "direction": "ascending"}` from the start, it's not such a big jump from there to wrapping in a list `[{"column": "X", "direction": "ascending"}]`.

By the way, while the `[("X", 1)]` formulation is fine for Python (e.g., in pymongo) I think an explicit object formulation as above is better practical in JSON / REST. Does that appeal you any better than the `("X", 1)` thing?

genematx: oh yeah, definitely, i was going to use the explicit formulation for the purpose of specifying the parameters (which makes me even more torn on it -- the parameters dictionary becomes a bit too complicated)...

danielballan: I could get on board with accepting two formulations:

1. Explicit formulation accepts sort order and multiple columns: `[{"column": "time", "direction": "ascending"}]`
2. Shorthand is just a single string, with ascending order assumed: `"time"`

genematx: I slept on this, and I came to the same conclusion :) Moreover, I think we need to expand the `unique_columns` (or similar) to be a set/list, not necessarily containing those from the order_by (we need `unique_*` to ensure writing consistency, to resolve race conditions when appending data).

danielballan (approved):

Issue #1266, opened 2026-01-09
Do not retry on `UnsupportedProtocol`


danielballan:
A typo in (or omission of) `http[s]` in the URL sends the client into a pointless retry loop.

```
from_uri("oops://example.com")
Scheduled retry in 1.0 seconds due to UnsupportedProtocol("Request URL has an unsupported protocol 'oops://'.") (attempt 1)
Scheduled retry in 0.23 seconds due to UnsupportedProtocol("Request URL has an unsupported protocol 'oops://'.") (attempt 2)
Scheduled retry in 1.0 seconds due to UnsupportedProtocol("Request URL has an unsupported protocol 'oops://'.") (attempt 3)
```

I did put _some_ care into targeting which exceptions get retried and which do not, but it's wroth a review of this approach. It may be that `UnsupportedProtocol` is not the only exception we're retrying on that we shouldn't.

PR #1406, opened 2026-06-09
Fix no retry unsupported protocol

gnzng:
The client retries on every `httpx.HTTPError`, including ones where the exact same request will fail the same way every time. The clearest case is a typo in the URL scheme — `from_uri("htps://...")`  as mentioned in #1266.

This makes `should_retry` fail fast on two deterministic errors:

- `httpx.UnsupportedProtocol` if the URL scheme isn't `http`/`https` (e.g. a typod `htps://`). Retrying re-sends the same bad URL.
- `httpx.LocalProtocolError`** if the client built an invalid request
  (e.g. an illegal header value containing a carriage return). The request is
  rejected before it leaves the client

## Change
`tiled/client/utils.py` — `should_retry` returns `False` for the two errors above,
before the catch-all `isinstance(exception, httpx.HTTPError)`.

## Testing
Two focused unit tests in `tests/test_client.py` (`test_should_not_retry_unsupported_protocol`, `test_should_not_retry_local_protocol_error`) assert `should_retry(...) is False` for each.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ Write the date in place of the "Unreleased" in the case a new version is release
   for any tiled client request (connection failures, 5xx errors, 429 rate limits).
   The spinner is animated in Jupyter notebooks as well as TTY terminals.
 - Respect the `Retry-After` header on HTTP 429 (Too Many Requests) responses.
+- Fail fast instead of retrying on deterministic client request errors that a
+  retry cannot fix: an unsupported URL scheme and an invalid request such as an
+  illegal header value.
 
 
 ## v0.2.11 (2026-05-27)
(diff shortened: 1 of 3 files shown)


danielballan:
Thanks for the contribution, @gnzng! Someone in my group will take this for a spin, just to kick the tires, and then we'll get this in.

sjulian4:
Hi Damian! I took a look at your PR and it shows significant improvements in the time it takes to reach exceptions for httpx.UnsupportedProtocol and httpx.LocalProtocolError with typos in the URL scheme (like “htpp” or “htp” for example) and invalid requests (such as with an improper API key in the header) respectively. I think that this is a great improvement for Tiled!

sjulian4 (approved):

gnzng:
great, thank you both @danielballan and @sjulian4 !

Issue #1226, opened 2025-11-14
update_metadata accepts string, but then breaks node


nmaytan:
If you call `update_metadata` from a client, it will accept a simple string argument (as it is serializable), but then causes server 500's when doing anything with that node. 

The metadata should be type checked before being accepted as an update.

client:
```
In [5]: l['C']
Out[5]: <ArrayClient shape=(10, 10) chunks=((10,), (10,)) dtype=float64>

In [6]: l['C'].update_metadata(metadata="test")

In [7]: l['C']
Scheduled retry in 0.54 seconds due to HTTPStatusError("Server error '500 Internal Server Error' for url 'http://localhost:8000/api/v1/metadata/C'\nFor more information, server admin can search server logs for correlation ID 576705178793e1d6.") (attempt 1)
```

server:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for NodeAttributes
metadata
  Input should be a valid dictionary [type=dict_type, input_value='test', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/dict_type
```

nmaytan:
Closed in #1408

PR #1408, opened 2026-06-15
Ensuring that requesting to update the metadata with values that do not serialize as a JSON object raises an exception in the client

sjulian4:
This PR is intended to fix the issue where trying to update the metadata with update_metadata() with a parameter that is not of the right type (will serialize as a JSON object) allows for the modification but then causes server errors when trying to access. To fix this issue, this PR checks to see if the metadata does not match the collections.abc.Mapping type when it contains a value with calling update_metadata(metadata=value).

This issue was initially brought up in [#1226](https://github.com/bluesky/tiled/issues/1226).

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Write the date in place of the "Unreleased" in the case a new version is release
   The spinner is animated in Jupyter notebooks as well as TTY terminals.
 - Respect the `Retry-After` header on HTTP 429 (Too Many Requests) responses.
 
+### Fixed
+
+- Ensuring that the metadata parameter value entered when calling update_metadata is of the proper type (will serialize as a JSON object) before altering the metadata.
+
 
 ## v0.2.11 (2026-05-27)
 
--- a/tests/test_writing.py
+++ b/tests/test_writing.py
@@ -5,6 +5,7 @@ Persistent stores are being developed externally to the tiled package.
 """
 
 import base64
+import collections
 import math
 import os
 import threading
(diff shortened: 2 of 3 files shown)


danielballan:
Thanks for this fix! Indeed, it looks like we accept any JSON-serializable type, where we should insist on only things that serialize as a JSON _object_.

In Python, I've been trained to be suspicious of `isinstance` checks because they can overly constrain the user. We don't need something that is literally an instance of `dict`; we need something that implements a dict-like interface (and will therefore serialize as a JSON object).

Here's a real example of a dict-like object in the Python standard library that is not a `dict` and would fail this check, but actually works fine:

```python
In [1]: from tiled.client import simple

In [2]: import collections

In [3]: c = simple()
Tiled version 0.2.10
http://127.0.0.1:43145/api/v1?api_key=3a3359cdbc50ed74

In [4]: d1 = {'a': 1}

In [5]: d2 = {'b': 2}

In [6]: d = collections.ChainMap(d1, d2)

In [7]: d
Out[7]: ChainMap({'a': 1}, {'b': 2})

In [8]: isinstance(d, dict)
Out[8]: False

In [9]: x = c.write_array([1,2,3], key='x')

In [10]: x.update_metadata(metadata=d)

In [11]: x
Out[11]: <ArrayClient shape=(3,) chunks=((3,)) dtype=int64>

In [12]: x.metadata
Out[12]: {'a': 1, 'b': 2}

In [13]: c['x'].metadata
Out[13]: {'a': 1, 'b': 2}

In [15]: import collections.abc
```

One path is to check against the abstract base case (abc) for "mappings" which is broader than `dict`.

```python
In [16]: isinstance(d, collections.abc.Mapping)
Out[16]: True
```

danielballan:
Sorry, I should have mentioned in my previous comment: Would you please add a test to verify the new behavior for a range of invalid inputs? Look for existing tests that exercise `update_metadata` using `git grep update_metadata`, and check that the expected type of error is raised using `pytest.raises`.

sjulian4:
Hi! Thank you for looking it over! I added the pytests, let me know if there is anything else I should change.

danielballan (approved):

nmaytan (approved):

PR #1409, opened 2026-06-15
revisions_count: implement pagination for metadata revisions

### LINKED ISSUE #1389: Revisions 'next' pagination link is never generated
## Bug

The \`next\` link in the revisions endpoint response is never populated, so it is impossible to paginate through revisions via the API links.

## Root cause

In \`tiled/server/core.py\`, \`construct_revisions_response\` computes \`count = len(data)\` — the number of items on the **current page** — and passes that as \`length_hint\` to \`pagination_links\`:

\`\`\`python
revisions = await entry.revisions(page.offset, page.limit)
data = [...]
count = len(data)  # ← page size, not total count
links = pagination_links(base_url, route, path_parts, page, None, count)
\`\`\`

\`pagination_links\` generates a \`next\` link only when \`page.offset + page.limit < length_hint\`. Since \`length_hint\` equals the current page size (at most \`page.limit\`), this condition is never true, so \`next\` is always \`null\`.

## Fix

Pass the total revision count as \`length_hint\` instead of the page size. This requires either a separate \`COUNT\` query or changing \`revisions()\` to fetch \`limit + 1\` rows so the caller can detect overflow, matching the pattern used in \`keys_range\`/\`items_range\`.

gnzng:
This PR closes #1389 (@genematx).

The revisions endpoint used the current page size as the count, so `meta.count` was just `len(data)` and the `next` pagination link was never generated.

Add `CatalogNodeAdapter.revisions_count()` (a COUNT query) and use it as the length hint in `construct_revisions_response`. 

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,8 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ### Fixed
 
-- The metadata revisions endpoint now reports the total revision count, so 
-all revisions can be paged through via the API.
+- The metadata revisions endpoint now reports the total revision count, so
+  all revisions can be paged through via the API.
 
 
 ## v0.2.11 (2026-05-27)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ Write the date in place of the "Unreleased" in the case a new version is release
   retry cannot fix: an unsupported URL scheme and an invalid request such as an
   illegal header value.
 
+### Fixed
+
+- The metadata revisions endpoint now reports the total revision count, so 
+all revisions can be paged through via the API.
+
 
 ## v0.2.11 (2026-05-27)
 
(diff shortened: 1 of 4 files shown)


danielballan (approved):
Thanks @gnzng! Looks good. I'll let @genematx get a look before this goes in. (Might be next week.)

genematx:
Thanks, @gnzng!
Should we add `ORDER BY` in the revisions query (e.g. by `revision_number` or `id`) to ensure that SQL ordering is deterministic before constructing the pages with offset/limit?
Also, would it make sense to have a test that checks that the pages have indeed been constructed (using a mocked small `page_size` or similar monkeypatching), like we do in `test_pagination`?

genematx:
Just pushed a minor performance improvement to reuse the same DB session when querying the revisions with count + a back-compat check. Otherwise, LGTM! Thank you, @gnzng!

genematx (approved):

Issue #1417, opened 2026-07-01
Client "whoami" fails for SimpleTiledServer


nmaytan:
Reported by @jessica-moylan 

Because `SimpleTiledServer` does not configure authentication providers, the client does not receive any authentication links to keep in `server_info` (as the server does not have any). Wiithout these links, `client.context.whoami()` fails violently.

In other places inside Tiled where `whoami` is called, we [check for these links](https://github.com/bluesky/tiled/blob/6e8ee5029b3dc69e3c9c8127920b1aea0c591ece/tiled/client/context.py#L359) before calling `whoami`. We should do the same in `Context.whoami` so that it can print a graceful message ([func](https://github.com/bluesky/tiled/blob/6e8ee5029b3dc69e3c9c8127920b1aea0c591ece/tiled/client/context.py#L1134)).

nmaytan:
Closed in #1418

PR #1418, opened 2026-07-06
Adding a check for authentication links in Context.whoami and a message if they are not present

sjulian4:
Adding a check in the Context.whoami function to ensure that authentication links are present before attempting to access them. This is to prevent the issue of trying to use this function with a SimpleTiledServer as presented in issue [#1417](https://github.com/bluesky/tiled/issues/1417). A graceful message informing the user of the missing authentication is printed, avoiding a NoneType error.

### Checklist
- [x] Add a Changelog entry


tiled/client/context.py
```diff
@@ -1133,14 +1133,17 @@ def force_auth_refresh(self):
...
+            for attempt in retry_context(self):
+                with attempt:
+                    return handle_error(
+                        self.http_client.get(
+                            self.server_info.authentication.links.whoami,
+                            headers={"Accept": MSGPACK_MIME_TYPE},
+                        )
+                    ).json()
+        else:
+            print("Authentication providers were not configured on the server.")
```
danielballan: Software tools, and _libraries_ especially, generally shouldn't print user messages to the stdout. Some alternatives to consider:

1. Just return `None`, silently.
2. Return `None` and issue a warning using `warnings.warn(...)`. (Warnings are sent to the _stderr_ by default.)
3. Return `None` and log at DEBUG, INFO, or WARNING level. All of these go to _stderr_, and the WARNING option is effectively equivalent to what `warnings.warn` does.

danielballan (changes requested):
Thanks for this fix.

I think returning `None` (which you do implicitly here) makes sense. :+1:

I'd like to revisit the choice to print a user message to the standard out; see comments in line.

tiled/client/context.py
sjulian4: Hi, thank you for the feedback! I just updated to use warnings.warn

danielballan (approved):
The test failures are flaky/unrelated.

PR #1423, opened 2026-07-08
Move node access_blob column into association table

nmaytan:
This is WIP, an an intermediate step towards moving towards a streamlined AccessTags table.

This adds an `AccessBlob` dataclass and uses it throughout, rather than just a dict with specific keys.

This is the first in a series of stacked PRs.

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tiled/utils.py
+++ b/tiled/utils.py
@@ -511,6 +511,12 @@ def import_object(colon_separated_string, accept_live_object=True):
         )
 
 
+def access_blob_matches(left, right) -> bool:
+    if left is None or right is None:
+        return left is right
+    return (left.username, left.tags) == (right.username, right.tags)
+
+
 def modules_available(*module_names):
     for module_name in module_names:
         if not importlib.util.find_spec(module_name):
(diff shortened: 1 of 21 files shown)


tiled/utils.py
```diff
@@ -476,6 +476,12 @@ def import_object(colon_separated_string, accept_live_object=True):
         )
 
 
+def access_blob_matches(left, right) -> bool:
+    if left is None or right is None:
+        return left is right
```
genematx, line 516: There might be some inconsistency between `access_blob == None` and `tags=[], username=None` (i.e. empty tags, no username); c.f. `_access_blob_to_orm(None)` returns `AccessBlob` with `tags=[]`. I need to think more about this, but it definitely raised some red flags.

nmaytan: I understand the concern. It should not be possible for `access_blob` to be `None` with these changes, but the case is handled (in multiple places) anyway. That is partly an artifact of working through the change from the column to the table.

nmaytan:
Merged in as part of #1514

Issue #1431, opened 2026-07-10
Update webhook documentation


JunAishima:
While trying to replicate the code in the webhook documentation, a few issues were seen - namely that these code fragments must be run in an IPython notebook, integration with Bluesky was not demonstrated (as this is how the webhooks will be used in practice), and that information on how to use the webhook to signal an application was not developed.

PR #1432, opened 2026-07-10
Update webhook docs

JunAishima:
### Checklist
- [X] Add a Changelog entry

### DIFF
--- a/docs/source/getting-started/webhooks.md
+++ b/docs/source/getting-started/webhooks.md
@@ -217,7 +217,7 @@ server.close()
 Configure Bluesky and run a scan!
 
 ```{code-cell} ipython3
-from bluesky.callbacks.tiled_writer import TiledWriter
+from bluesky-tiled-plugins.writing.tiled_writer import TiledWriter
 from bluesky import RunEngine
 from tiled.client import from_uri
 
--- a/docs/source/getting-started/webhooks.md
+++ b/docs/source/getting-started/webhooks.md
@@ -73,7 +73,7 @@ print(f"Receiver listening on http://127.0.0.1:{receiver_port}/hook")
 Note that the `received` variable will be accessible from the IPython session,
 which will enable easier debugging.
 
-The structure of the json object is described below.
+The structure of the json object is described in [](Write_Deliveries)
 
 ## Start a Tiled server with webhooks enabled
 
@@ -123,7 +123,7 @@ webhook = resp.json()
 webhook_id = webhook["id"]
 print(f"Webhook registered (id={webhook_id})")
 ```
-
+(Write_deliveries)=
 ## Write data and watch the deliveries arrive
 
 Every `write_array` call creates a new catalog entry, which triggers a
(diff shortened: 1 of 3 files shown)


JunAishima:
https://github.com/bluesky/tiled/issues/1431

docs/source/getting-started/webhooks.md
```diff
@@ -31,11 +31,16 @@ This makes webhooks a good fit for:
 - integrating Tiled with systems that cannot maintain a long-lived connection
 
 This tutorial demonstrates the full webhook lifecycle end-to-end in a single
-Python session, with no external services and no configuration files required.
+IPython session, with no external services and no configuration files required.
+Using IPython is necessary to instantiate and persist the classes and services.
```
danielballan, line 35: I remember you mentioning this, but I question it. What "persistence" is IPython providing exactly? Can you provide a minimal example of what fails if you use a plain `python` interpreter?

If this doesn't working with a vanilla `python` interpreter, I think I would rather update the example than require IPython to run it.

docs/source/getting-started/webhooks.md
```diff
@@ -31,11 +31,16 @@ This makes webhooks a good fit for:
...
+
+The server and receiver can be made more modular by using the port and API key information
```
danielballan, line 37: A guiding principle of diataxis tutorials is: don't present the user with choices. They don't have enough context to make choices yet. Just prescribe. I don't think we're at the point yet where we need to be reconfiguring...let's stick to a single process.

docs/source/getting-started/webhooks.md
```diff
@@ -65,6 +70,11 @@ threading.Thread(target=receiver.serve_forever, daemon=True).start()
 print(f"Receiver listening on http://127.0.0.1:{receiver_port}/hook")
 ```
 
+Note that the `received` variable will be accessible from the IPython session,
```
danielballan, line 73: I don't think this is helping. _All_ variables are accessible from whatever Python interpreter you are using...

docs/source/getting-started/webhooks.md
```diff
@@ -100,7 +111,11 @@ resp = httpx.post(
         "Authorization": f"Apikey {server.api_key}",
         "Content-Type": "application/json",
     },
-    content=json.dumps({"url": f"http://127.0.0.1:{receiver_port}/hook"}),
```
danielballan, line 103: I find the simpler version (without explicit `None`s) more appropriate style for a tutorial.

docs/source/getting-started/webhooks.md
```diff
@@ -212,7 +229,21 @@ receiver.shutdown()
 server.close()
 ```
 
-## See also
+# Extending toward real-world situations
```
danielballan, line 232: Recapping a Slack conversation from a week or so ago: We studiously keep "bluesky" separate here to make clear that tiled is useful on generic data, not just bluesky data. Let's wait to add this as part of a larger section of "bluesky integration." We don't have space in this document to contextualize "RunEngine," "stop document," and so on for Tiled users.

danielballan (changes requested):
Thanks for taking a look at this.

As I look over my line review here, I think we brought different expectations to this document. I want it to guide the user down a prescribed path to a minimal working example. I want to avoid presenting them with choices or applied examples that might pull in a lot of additional context ("RunEngine").

I'm also wondering there is some confusion around whether it works with vanilla `python` or requires IPython. I think this will work fine with vanilla `python` as is. I'm inclined to close the PR unless we can confirm an issue there.

PR #1444, opened 2026-07-20
Fix device code auth 

burkeds:
When I try to use OIDCAuthenticator to authenticate a device (the Tiled python client) with Keycloak I get what looks like a malformed URI:

api/v1/auth/provider/keycloak_oidc/token

This is a broken URI brings me to a page that says "method not allowed" or similar. Per the docs, the redirect URI should end in `code` or `device_code`.

I think I found the issue in `tiled.client.context:device_code_grant` 

```python
            else:
                # The device code will be given to Tiled's own implementation
                # of device code flow. Tiled will do authorization code from with the
                # external providers. (This handles cases that ORCID and Globus that do
                # not support device code flow.)
                oauth2_spec = False
                verification_response = http_client.post(auth_endpoint)
                handle_error(verification_response)
                verification = verification_response.json()
                token_endpoint = verification["authorization_uri"]
                verification_uri = verification["verification_uri"]
```

Here I think the authorization_uri and verification_uri are mistakenly swapped. When I swap the token_endpoint and verification_endpoint as in the PR, the device flow is able to contact Keycloak and initiate device auth.

verification_uri and authorization_uri (tiled and 3rd party auth endpoints respectively) are reversed, breaking the device code workflow. Results in no contact with the auth provider and an HTML yield to the downstream client which breaks its JSON decoder.

### DIFF
--- a/tests/test_device_flow.py
+++ b/tests/test_device_flow.py
@@ -13,7 +13,7 @@ from tiled.adapters.mapping import MapAdapter
 from tiled.client import Context
 from tiled.client.auth import TiledAuth
 from tiled.client.constructors import from_context
-from tiled.client.context import prompt_for_credentials
+from tiled.client.context import device_code_grant, prompt_for_credentials
 from tiled.server.app import build_app_from_config
 
 tree = MapAdapter({})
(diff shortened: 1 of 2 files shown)


genematx:
Thanks @burkeds for fixing this. I've added [a couple of regression tests](https://github.com/burkeds/tiled/pull/1) (with Clade); please check if they make sense to you too. Otherwise -- looks good to me.

genematx (approved):

burkeds:
Looks good to me! Thanks for adding the tests.

Issue #1380, opened 2026-05-10
Make Webhooks SSRF configurable


danielballan:
Sometimes, Webhooks need to be delivered to _selected_ internal services. We need to add configuration to Tiled that allows exceptions to the block list.

https://github.com/bluesky/tiled/blob/eeb92646206b7b1ed54c3641851603de2584ea15/tiled/server/webhooks.py#L91-L105

The exceptions should be given as host names, not IPs, because there may be a mixture of allowed and disallowed servers behind the same proxy.

Additionally, we should make the block list extensible to add certain site-specific internal IP ranges.

danielballan:
Webhooks configuration goes into the config file like

```yaml
webhooks:
  ...
```

and the options are defined here:

https://github.com/bluesky/tiled/blob/691b42a79851ef818d54994b82d6be5ca94bc56b/tiled/config.py#L241-L267

PR #1445, opened 2026-07-20
Webhook configurability

JunAishima:
### Checklist
- [x] Add a Changelog entry


JunAishima:
closes #1380

tiled/server/simple.py
```diff
@@ -72,6 +72,12 @@ class SimpleTiledServer:
         checks), making it easy to test against a local receiver.
         A ``webhook_secret_key`` is auto-generated and exposed as an
         attribute.  Default is False.
+    blocked_networks: Optional[list[str]]
```
danielballan: The `SimpleTiledServer` is hard-coded to accept local traffic only. Once you're customizing this configuration, you have graduated beyond "simple". Let's drop this config options here and test against a real server.

danielballan:
For logging config, consider updating this too.

https://github.com/bluesky/tiled/blob/1853c2a1c58792815ef331af0c5d9cc9191773da/tiled/server/logging_config.py#L70-L77

tiled/server/simple.py
JunAishima: fixed in https://github.com/bluesky/tiled/pull/1445/commits/a6aef4d9de014ced953b9681aa1b37d7c3281647

JunAishima:
> For logging config, consider updating this too.
> 
> https://github.com/bluesky/tiled/blob/1853c2a1c58792815ef331af0c5d9cc9191773da/tiled/server/logging_config.py#L70-L77

fixed in https://github.com/bluesky/tiled/pull/1445/commits/ce4ba2782c94d4bee2d86bc697c847c5e1d3efdc

tiled/config.py
```diff
@@ -260,6 +266,8 @@ class WebhooksConfig(BaseSettings):
     """
 
     secret_keys: list[str] = []
+    blocked_networks: Optional[list[str]] = []
```
danielballan: I think you had the right the first time, i.e., _without_ the `Optional`. Without the `Optional`, you can still have a default value and make the value optional.

```python
In [4]: class A(BaseModel):
   ...:     stuff: list = []
   ...:

In [5]: a = A()
```

Additional `Optional` makes `None` a valid value.

```python
In [4]: class A(BaseModel):
   ...:     stuff: list = []
   ...:

In [5]: a = A()

In [6]: a = A(stuff=None)  # not allowed
...
ValidationError: 1 validation error for A
stuff
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.13/v/list_type

In [7]: class B(BaseModel):
   ...:     stuff: Optional[list] = []
   ...:

In [8]: b = B(stuff=None)  # allowed
```

I don't think we want or need to accept `None` here; empty list is right.

JunAishima: fixed in https://github.com/bluesky/tiled/pull/1445/commits/51f643ff4e76a07ea0eaea46ae967b786ffdc110

danielballan:
This [comment](https://github.com/bluesky/tiled/pull/1445#issuecomment-5046889887) wasn't addressed. The file isn't changed at all by this PR. (Maybe a change got lost somehow?)

danielballan:
As discussed on Zoom, the comment about logging still needs to be addressed. The hard-coded defaults in `tiled.server.logging_config` should get the same update as the example logging config YAML file.

danielballan (approved):
Tested locally on BNL infra. Works as advertised.

PR #1453, opened 2026-07-28
FORK: publish LCLS image to cr.stanford.edu from CI

swelborn:


swelborn:
oops! meant for pushing to our container registry, targeted bluesky:main instead of my fork. sorry for the noise.

PR #1460, opened 2026-08-04
Redis HA

gwbischof:
Updates to support Redis high availability (Sentinel) for the streaming cache.

This PR only changes Tiled behavior when a Redis HA configuration is set in the Tiled config file; the standalone `uri` path is unchanged.

## Configuration

The `streaming_cache` section gains a Sentinel-based alternative to a single `uri` because the Sentinal-based Redis client does not support a `uri`.

- `sentinels` — a list of `"host:port"` Sentinel addresses.
- `service_name` — the monitored primary's name (e.g. `mymaster`).
- `password` — password for the Redis data nodes.
- `ssl` — enable TLS for the Sentinel path (Sentinel discovery **and** the data-node connections the client follows failover across). The standalone path expresses TLS via the `rediss://` scheme instead.
- `health_check_interval` — seconds between health-check PINGs on idle connections, so redis-py detects a stalled connection after a failover and reconnects instead of blocking on the dead primary.

When `sentinels` is set the datastore uses `Sentinel(...).master_for(service_name)`, which auto-detects failover and reconnects to the current primary.

## Notes

- `redis.asyncio.from_url()` and `redis.asyncio.Sentinel.master_for()` return the exact same class (`redis.asyncio.client.Redis`), and every method Tiled calls (`incr`, `pipeline`, `hset`, `expire`, `publish`, `get`, `hmget`, `pubsub`) is literally the same function. The only difference is the connection pool: `SentinelConnectionPool` vs a plain `ConnectionPool`.
- `Sentinel(...).master_for(service_name)` auto-detects failover and reconnects to the new primary (and doesn't support a URI connection string): https://redis.readthedocs.io/en/stable/connections.html#id1

## Handling a dropped subscription

In redis-py's asyncio client, the server's pinned pub/sub connection drops on **both** a hard crash of the primary **and** a graceful `sentinel failover`. In the graceful case this is because redis-py's Sentinel connection pool detects on its next health check that the node it is connected to was demoted from primary, and disconnects to re-resolve the new primary. Either way the drop surfaces as a `ConnectionError` that terminates `pubsub.listen()`.

Rather than resubscribing in place server-side — which would duplicate Tiled's existing client reconnect + sequence-replay — the server unblocks its send loop and closes the websocket abnormally (`1012`). The client's existing reconnect loop resumes from `_last_received_sequence` (tiled-client state) and the server replays from the new primary's current state. Tiled server doesn't keep track of `_last_received_sequence` this adds to the argument that the tiled-client is where the reconnect logic belongs.

Without this, the send loop blocks forever on an empty buffer, the socket stays open, and the subscriber silently stops receiving events permanently; with it, the subscriber reconnects and re-syncs. This keeps the diff minimal and leaves the non-HA path unchanged.

## Delivery semantics

Live streaming is **best-effort** — the durable record is the SQL catalog, which is committed before the streaming publish. According to the Redis docs, Redis Pub/Sub is at-most-once:

> Redis' Pub/Sub exhibits at-most-once message delivery semantics. As the name suggests, it means that a message will be delivered once if at all. Once the message is sent by the Redis server, there's no chance of it being sent again. If the subscriber is unable to handle the message (for example, due to an error or a network disconnect) the message is forever lost. (https://redis.io/docs/latest/develop/pubsub/, "Delivery semantics")

An **ordinary connection drop is lossless**: the client reconnects and resumes from `_last_received_sequence + 1`, and the server replays anything after that (skipping sequences `<= last_sent`), so a normal reconnect neither loses nor duplicates updates.

The **residual risk is a failover to a replica that had not yet caught up**. Redis replication is asynchronous and there is no per-write replication confirmation, so:

- a write the old primary acknowledged but had not replicated is **lost**; and
- because the promoted replica's `INCR` sequence counter is behind, it can re-issue already-used sequence numbers for new data, **re-delivering** them (the client does not filter incoming messages by its own last-received).

Server-side `min-replicas-to-write` / `min-replicas-max-lag` on the Redis primary narrow this window (reject writes when replicas are absent or lagging) but cannot guarantee a given write was replicated. That is the guard this PR relies on. A client-issued Redis `WAIT` write-concern would bound the window further; it is **deferred** — see below for why, and as a reference if we add it later.

## Deferred: client-issued `WAIT` write-concern (kept for reference)

An earlier revision of this PR added a Redis `WAIT` write-concern on streaming publishes, controlled by two `streaming_cache` settings (`wait_num_replicas`, `wait_timeout`): after each publish, block until `wait_num_replicas` replicas acknowledge the write, and on a shortfall (fewer acks than requested within `wait_timeout`) raise so the client's write fails — the same path any Redis write error already takes (the publish runs after the Postgres commit, so a failure propagates to the client). This section records the rationale so we understand the trade-offs.

### Why `WAIT` requires a non-transactional pipeline

Redis `WAIT` is a **per-connection** guarantee: it blocks the calling connection until writes previously issued *on that same connection* are acknowledged by the requested number of replicas (https://redis.io/docs/latest/commands/wait/). Two consequences follow.

First, `WAIT` must run on the **same connection** as the writes. The datastore uses a single shared, pool-backed async client (both the standalone `from_url` path and the Sentinel `master_for` path draw from one pool shared across concurrent requests). Running `pipeline.execute()` and then a separate `client.wait(...)` does not guarantee both land on the same connection — under concurrency the pool can hand `WAIT` a *different* connection than the one that did the writes, making the check meaningless. (Serial tests pass only because the pool is LIFO and returns the just-released connection — luck, not a guarantee.) Appending `WAIT` to the **same pipeline** forces redis-py to acquire one connection, run the writes and `WAIT` on it, then release it.

Second, `WAIT` **does not block inside a `MULTI`/`EXEC` transaction pipeline** — there it returns immediately, reporting the replica count at enqueue time instead of waiting for acknowledgements. redis-py's default `pipeline()` is transactional, so a `WAIT` appended to it would return instantly and guarantee nothing. To make `WAIT` block, the publish pipeline must be **non-transactional** when the write-concern is active (`self.client.pipeline(transaction=not self.wait_num_replicas)`), reading the trailing result from `results[-1]`.

Dropping `MULTI`/`EXEC` atomicity (transaction=false) on that path may not be safe, it would change the current proven write behavior and would require a decent amount of investigation to prove that we can drop the atomicity of the write pipeline.

**Why it is client-driven and not a server setting.** Redis replication is asynchronous by design and there is no `redis.conf` directive that makes an individual write block until it is replicated; the docs state that *"synchronous replication of certain data can be requested by the clients using the `WAIT` command"* (https://redis.io/docs/latest/operate/oss_and_stack/management/replication/). The one server-side knob, `min-replicas-to-write` / `min-replicas-max-lag`, is explicitly *best-effort* — it gates writes on how many replicas are connected and within a lag bound, but *"it is not possible to ensure the replica actually received a given write."* It therefore complements `WAIT` but cannot replace the per-write acknowledgement check, which must be issued and verified by the client. (Both MongoDB and Kafka fold the replication check into the write itself and enforce it server-side, rather than exposing it as a separate command the client has to issue on the right connection.)

## Integration testing

Because this path can't be exercised by the in-process `TestClient`, it was validated manually against a live Redis Sentinel replica set in podman (3 Redis nodes + 3 Sentinels, Tiled running in-network, pointed at the Sentinels via structured config), in both plaintext and full-TLS mode.

The test reproduces a Redis node failure:

1. A subscriber attaches to a node and records every sequence number it receives.
2. A background writer appends a new frame every 0.4s, continuously, before/during/after the outage.
3. After a steady-state phase, the primary Redis node the server's pub/sub is pinned to is crashed (`podman stop`), and Sentinel elects a new primary.
4. The writer keeps appending straight through the failover and reconnect window.

We verify that:

- the received sequence is **contiguous with no gaps** (for the caught-up subscriber);
- the stream **kept growing** after the crash;
- the server logs a **non-zero `start=` reconnect handshake** for that node — the unambiguous fingerprint of drop → reconnect → resume (the initial connect is always `start=0`);
- the **array read back** at the end equals the full range that was written (end-to-end data integrity).

**Result:** with the primary killed mid-stream, Sentinel promoted a replica and the client reconnected. The caught-up subscriber's stream stayed contiguous across the failover (it had already received everything the old primary acked); the reconnect resynced it to the new primary rather than hanging. A subsequent graceful `sentinel failover` likewise dropped the pinned connection, triggered a resume, and resynced. Without this change the same scenario hangs the socket open and the subscriber silently stops receiving events.

This is exactly the behavior we want: a primary crash or failover is transparent to subscribers instead of a silent permanent stall.

### DIFF
--- a/tiled/server/app.py
+++ b/tiled/server/app.py
@@ -176,7 +176,7 @@ def build_app(
 
     if scalable:
         streaming_cache = server_settings.get("streaming_cache", None)
-        if streaming_cache and streaming_cache["uri"].startswith("memory"):
+        if streaming_cache and (streaming_cache.get("uri") or "").startswith("memory"):
             raise UnscalableConfig(
                 dedent(
                     """
--- a/tiled/server/streaming.py
+++ b/tiled/server/streaming.py
@@ -11,7 +11,6 @@ import orjson
 from fastapi import WebSocketDisconnect
 from redis import asyncio as redis
 from redis.asyncio.sentinel import Sentinel
-from redis.exceptions import ConnectionError as RedisConnectionError
 
 from ..ndslice import NDSlice
 from ..utils import safe_json_dump as _safe_json_dump
@@ -194,6 +193,12 @@ class PubSub:
         return gen()
 
 
+# Sentinel pushed onto the live-event buffer when the live subscription drops
+# (e.g. a Redis failover). It signals the handler to close the socket abnormally
+# so the client reconnects and replays any sequences missed during the outage.
+_LIVE_SUBSCRIPTION_LOST = object()
+
+
 def _make_ws_handler_common(
     *,
     websocket,
@@ -302,6 +307,9 @@ def _make_ws_handler_common(
                 logger.exception(
                     f"Live subscription error for node {node_id}: {e}",
                 )
+                # Signal the handler to close the socket abnormally so the client
+                # reconnects and replays any sequences missed during the outage.
+                await stream_buffer.put(_LIVE_SUBSCRIPTION_LOST)
             finally:
                 if live_cleanup is not None:
                     await live_cleanup()
@@ -319,6 +327,13 @@ def _make_ws_handler_common(
         try:
             while not end_stream.is_set():
                 live_seq = await stream_buffer.get()
+
+                if live_seq is _LIVE_SUBSCRIPTION_LOST:
+                    # Close abnormally (1012) so the client's reconnect loop
+                    # resumes from its last received sequence and replays any
+                    # events missed while the subscription was down.
+                    await websocket.close(code=1012, reason="Live subscription lost")
+                    return
 
                 # Skip duplicates or already replayed messages
                 if live_seq <= last_sent:
@@ -522,28 +537,12 @@ class RedisStreamingDatastore(StreamingDatastore):
             await pubsub.subscribe(f"notify:{node_id}")
 
             async def live_iter():
-                nonlocal pubsub
-                while True:
-                    try:
-                        async for message in pubsub.listen():
-                            if message.get("type") == "message":
-                                try:
-                                    yield int(message["data"])
-                                except Exception as e:
-                                    logger.exception(f"Error parsing live message: {e}")
-                    except RedisConnectionError:
-                        # A Sentinel failover closes the pub/sub link; re-create
-                        # it (re-resolving the current primary) and resubscribe.
-                        logger.warning(
-                            "Streaming pub/sub connection lost; resubscribing."
-                        )
+                async for message in pubsub.listen():
+                    if message.get("type") == "message":
                         try:
-                            await pubsub.aclose()
-                        except Exception:
-                            pass
-                        await asyncio.sleep(0.5)
-                        pubsub = self.client.pubsub()
-                        await pubsub.subscribe(f"notify:{node_id}")
+                            yield int(message["data"])
+                        except Exception as e:
+                            logger.exception(f"Error parsing live message: {e}")
 
             async def cleanup():
                 await pubsub.unsubscribe(f"notify:{node_id}")
--- a/tiled/server/streaming.py
+++ b/tiled/server/streaming.py
@@ -32,7 +32,9 @@ def _build_redis_client(settings: Dict[str, Any]) -> redis.Redis:
     kwargs = dict(
         socket_timeout=settings["socket_timeout"],
         socket_connect_timeout=settings["socket_connect_timeout"],
-        health_check_interval=settings["health_check_interval"],
+        # Default mirrors StreamingCacheConfig; tolerate a partial cache_config
+        # (e.g. from a direct from_uri call) as the memory datastore does.
+        health_check_interval=settings.get("health_check_interval", 30),
     )
     sentinels = settings.get("sentinels")
     if sentinels:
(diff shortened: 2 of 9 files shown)


tiled/server/streaming.py
```diff
@@ -502,12 +522,28 @@ async def live_sequence_source():
...
+                        )
                         try:
-                            yield int(message["data"])
-                        except Exception as e:
-                            logger.exception(f"Error parsing live message: {e}")
+                            await pubsub.aclose()
+                        except Exception:
+                            pass
+                        await asyncio.sleep(0.5)
+                        pubsub = self.client.pubsub()
```
dylanmcreynolds: sleeping for .5 seconds and reconnecting will only work once since this `pubsub()` is not caught. Do we want to catch this to and go back to the loop too keep trying?

gwbischof: Thanks for looking at my PR :D
I just pulled out all of this code, we don't actually need the extra reconnect loop in the server code.

dylanmcreynolds (approved):

tiled/server/app.py
```diff
@@ -176,7 +176,7 @@ def build_app(
 
     if scalable:
         streaming_cache = server_settings.get("streaming_cache", None)
-        if streaming_cache and streaming_cache["uri"].startswith("memory"):
+        if streaming_cache and (streaming_cache.get("uri") or "").startswith("memory"):
```
dylanmcreynolds, line 179: What's the need for the extra ` or ""`?  It seems harmless if config isn't as you'd expect, but I'm curious what edge case this covers.

gwbischof: The reason this is here is because `uri` became optional with this PR, and the reason that it's optional is because `uri` isn't supported by the HA version of the Redis client. So there are two different sets of arguments required for the 2 different types of Redis clients.

danielballan (approved):
Great work!

Issue #1459, opened 2026-08-03
Column/identifier validation errors (e.g. name > 63 chars) surface as an opaque 500 instead of a descriptive 4xx


Bilchreis:
## Summary

When creating a table-backed node whose schema contains a column name that fails `is_safe_identifier`'s validation (e.g. exceeds the 63-character SQL identifier limit in `tiled/adapters/sql.py`), the resulting `ValueError` isn't caught anywhere in the request path. It propagates up through the ASGI stack as an unhandled exception, and the client only ever sees:

```
Client error '500 Internal Server Error' for url '...'
For more information, server admin can search server logs for correlation ID ...
```

The actual cause, `ValueError: Invalid SQL identifier "...": max character number is 63` , only exists in the server's `error.log` file on disk. In cases where a user has no access to the error logs, they have no chance of finding out what went wrong even though the fix would be ad simple as shortening the signal name. 

## Compounding effect

Because 500s are treated as retryable by the client, the identical request gets retried automatically. But the catalog row for the node appears to be inserted before `init_storage()` (where the identifier validation happens) runs, so the retry doesn't reproduce the 500, it hits a `409 Conflict` on the now-partially-created node instead. The 409 is *more* misleading than the original 500: it reads like a naming/concurrency conflict, with no trace left of the actual validation failure that caused it.

## Suggested fix

- Catch schema/identifier validation errors (`ValueError` from `is_safe_identifier`, and similar) in `create_node`/`init_storage`, and return them as a `4xx` (422) response with the actual message , not a bare 500.

## Minimal repro

```python
from bluesky import RunEngine
from bluesky.plans import count
from ophyd_async.core import init_devices
from bluesky_tiled_plugins import TiledWriter
from tiled.server import SimpleTiledServer
from tiled.client import from_uri


# Initialize the Tiled server and client
save_path = "/data"
tiled_server = SimpleTiledServer(readable_storage=[save_path])
tiled_client = from_uri(tiled_server.uri)



RE = RunEngine({})
tw = TiledWriter(tiled_client)
RE.subscribe(tw)


# Minimal, ophyd-async-only reproduction of the "max character number is 63"
# SQL identifier bug -- no secop_ophyd involved, to prove it's generic to any
# sufficiently long/nested ophyd-async Device, not specific to SECoP.
from ophyd_async.core import StandardReadable, soft_signal_rw, init_devices


class VerboseDemoDevice(StandardReadable):
    def __init__(self, name: str = ""):
[...]

PR #1462, opened 2026-08-05
Raising a 422 exception when column name is too long instead of a 500 exception

sjulian4:
This PR addresses Issue #1459. It raises an HTTPException with status code 422 when the column name is greater than 63 characters instead of 500 and 409 exceptions.

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/tests/adapters/test_sql.py
+++ b/tests/adapters/test_sql.py
@@ -20,6 +20,7 @@ from tiled.storage import (
 from tiled.structures.core import StructureFamily
 from tiled.structures.data_source import DataSource, Management
 from tiled.structures.table import TableStructure
+from tiled.utils import UnsafeIdentifier
 
 names = ["f0", "f1", "f2", "f3"]
 data0 = [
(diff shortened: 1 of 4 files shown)


danielballan:
Thanks for taking this on. We make an effort to keep HTTP-related code separated from the core application code. We are not perfect about this, but we're pretty good about it, and I'd like to keep moving in that direction. Notice,  for example, only one import of `fastapi` and no imports of `HTTPException` within the whole `tiled.adapters` sub-package.

```sh
❯ pwd
/home/dallan/Repos/bnl/tiled/tiled/adapters
❯ git grep fastapi
mapping.py:    from fastapi import APIRouter
❯ git grep HTTPException
❯ 
```

To maintain a clean separation, we define certain custom exception classes in `tiled.utils` and raise _those_ within the adapter code. Then, we use a mechanism is FastAPI for registering exceptions that should be handled a particular way. For example, `UnsupportedQueryType` becomes a `400` with a custom error message. The same approach could be used in this case.

https://github.com/bluesky/tiled/blob/f1e9efcaaf0f6849062a3f4d84b0610b0c73d8bc/tiled/server/app.py#L351-L375

sjulian4:
Hi, thank you for the feedback! I pushed up some code updates to better align with that.

danielballan (approved):
Works for me! It would be great if someone else could test drive this just to be sure. Then, good to merge from my side.

sjulian4:
Thank you for taking a look!

After discussing with @nmaytan, we thought it may be better to check that the name of the identifier is within the size constraint based on the amount of bytes rather than the number of characters in the string. This is because the smallest limit (across PostgreSQL, SQLite, and DuckDB) for the maximum identifier length is 63 bytes rather than 63 characters (this comes from [PostgreSQL](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)). Some characters, such as accented letters or emojis, are more than 1 byte. The proposed changes for converting to a comparison based on byte length are made in the most recent commit.

danielballan:
Tested PR branch locally, works as advertized.

PR #1465, opened 2026-08-11
Implement Lazy Loading of HDF5 Assets -- ON HOLD

genematx:
## Summary

Follow-up on #1463.

#1463 made many-file HDF5 reads reuse a cached Dask graph, but the **first** read of a dataset in a fresh process still opened *every* constituent file once to discover specs before the cache could help. This PR extends the lazy asset-resolution path (introduced for file sequences in #1463) to multi-file HDF5 array datasets, so the first read opens only the native chunks its slice actually touches.

## Why HDF5 needed its own path

A file sequence stacks **uniform** frames — one file adds exactly one row on a new leading axis — so the catalog's `1 file = 1 row` geometry recovers file boundaries from the structure alone. HDF5 files instead **concatenate** along the array's leading axis, and each file may contribute a *different* number of rows, so the boundaries can not be read off `structure.chunks`.

## What changed

- **`catalog/adapter.py`** — `_get_lazy_adapter` is now adapter-agnostic. It no longer gates on a `chunks` property; it derives the file count from a single aggregate `data_uris` asset query (which also performs the existing `num` contiguity check), then asks the adapter class `file_indices_for_slice(structure, n_files, slice, properties=..., parameters=...)` which files a slice touches. Only the adapter interprets `properties`. It also forwards the data source `parameters` (as the eager path does) so adapters that need them (e.g. HDF5's `dataset`) are configured. Previously the lazy path was reachable only when the data source recorded a `chunks` property (the old file-count source); dropping that requirement lets stacks without one use it too.
- **`adapters/hdf5.py`** — `HDF5ArrayAdapter` opts in via `supports_lazy_assets` and gains `_file_layout`, which resolves each file's extent along the leading axis in priority order:
  1. an optional per-asset **`extents`** property — the authoritative per-file row counts (`sum(extents) == shape[0]`); supports genuinely **non-uniform** files,
  2. else `structure.chunks[0]` when it has exactly one chunk per file,
  3. else the **grid** geometry (a uniform stack whose leading axis was reshaped across several structure dimensions),
  4. else `None` → the catalog falls back to the full build.

  `_assemble_partial` builds the file-stacked Dask array from the known structure without opening any file for specs. It mirrors the eager `lazy_load_hdf5_array`: the flat (pre-reshape) concatenation is tiled by **native HDF5 chunks** (axis 0 by `chunks[0]`, also split at file boundaries; trailing axes by `chunks[1:]`), so the native chunk is the minimum read unit. Touched chunks resolve their URI to a path lazily *inside* the read task; untouched files carry a placeholder block that Dask culls once the read reshapes and slices the array (and that raises loudly if ever computed, so a geometry mismatch can never serve wrong data). A reshaped grid with no recorded native chunking is the one case that still serves a whole file per grid cell (the file is then the minimal read unit).
- **`adapters/utils.py`** — the file-alignment geometry shared by the sequence and HDF5 adapters is extracted into `grid_shape_for_files` (a pure function of shape + file count, no I/O).

## On the optional `extents` property

`extents` is optional and fully adapter-contained; the catalog stays ignorant of it. The dataset benchmarked below resolves *without* it — one leading chunk per file (path 2). It is needed only for genuinely non-uniform *flat* files (differing per-file row counts that are not one-chunk-per-file). A writer can populate `properties["extents"]` to unlock the lazy path there; that consolidator change is a planned follow-up (not in this PR).

## Benchmark

Synthetic xspress3-like dataset: `N` files, each `(11, 8, 4096)` `uint32` with native chunk `(1, 8, 4096)` (one frame per chunk), concatenated to `(N, 11, 8, 4096)`. The metric is the `h5open` call count (deterministic) plus median wall time over 5 reps. **Before** is the merge-base (eager HDF5, post-#1468 revert); **after** is this PR. Script at the bottom.

**`h5open` calls (before → after):**

| scenario | N=3 | N=25 | N=100 | N=300 |
|---|---|---|---|---|
| single frame `[1,5]` | 5 → **1** | 27 → **1** | 102 → **1** | 302 → **1** |
| one file, all frames `[1]` | 15 → **11** | 37 → **11** | 112 → **11** | 312 → **11** |
| two files `[0:2]` | 26 → **22** | 48 → **22** | 123 → **22** | 323 → **22** |
| one frame from every file `[:,5]` | 7 → **3** | 51 → **25** | 201 → **100** | 601 → **300** |
| whole dataset | 37 → **33** | 301 → **275** | 1201 → **1100** | 3601 → **3300** |

Before, every scenario pays an O(N) specs-reading pass across all files. After, opens equal the number of native chunks the slice actually overlaps — **independent of `N`** for targeted reads (`[1,5]`→1, `[1]`→11 frames of one file, `[0:2]`→22). Reads that genuinely span every file (`[:,5]`, whole) stay O(N) but shed the redundant specs pass.

**median wall time, ms (before → after):**

| scenario | N=3 | N=25 | N=100 | N=300 |
|---|---|---|---|---|
| single frame `[1,5]` | 4.3 → 2.6 | 8.1 → 3.1 | 21.2 → 5.6 | 56.3 → **11.3** |
| one file, all frames `[1]` | 8.0 → 6.6 | 14.5 → 7.1 | 25.8 → 9.7 | 60.5 → **16.5** |
| two files `[0:2]` | 11.9 → 10.4 | 16.2 → 11.1 | 29.4 → 14.6 | 64.3 → **19.9** |
| one frame from every file `[:,5]` | 4.9 → 3.6 | 17.8 → 13.6 | 64.6 → 48.9 | 224.2 → **191.6** |
| whole dataset | 16.0 → 15.3 | 97.6 → 100.7 | 455.5 → 428.4 | 1287.7 → 1333.7 |

Whole-dataset reads are within noise (both read every chunk); targeted reads improve ~5× at N=300 and the gap widens with `N`.

## Tests

- `pytest tests/test_catalog.py tests/test_hdf5.py tests/test_tiff.py tests/test_slicer.py` → **264 passed, 29 skipped**.
- New unit tests in `test_catalog.py` cover the grid path (opens only touched files), multi-chunk files, the non-file-aligned fallback, the `extents` property (parameterized, incl. non-uniform), the single-file decline, and the `slice`/`squeeze` transform decline.
- New tests in `test_hdf5.py` verify native-chunk reads from multi-file URIs and that only the touched files are opened and then closed.
- pre-commit (flake8 / isort / black / mypy) all pass.

## Checklist
- [x] Add a Changelog entry

<details>
<summary>Benchmark script</summary>

```python
"""Ad-hoc benchmark for multi-file (xspress3-like) lazy HDF5 reads.

Runs against whichever tiled is importable (editable install -> current git
branch). Measures h5open() call counts and wall time for representative slice
patterns as a function of the number of assets (files).
"""
import asyncio
import statistics
import tempfile
import time
from dataclasses import asdict
from pathlib import Path
from unittest.mock import patch

import numpy

import tiled
import tiled.adapters.hdf5
from tiled.catalog import in_memory
from tiled.ndslice import NDSlice
from tiled.server.schemas import Asset, DataSource
from tiled.structures.array import ArrayStructure, BuiltinDtype
from tiled.utils import ensure_uri

# xspress3-like geometry: each file holds (F, H, W) uint32, one frame per chunk.
F, H, W = 11, 8, 4096
DTYPE = numpy.dtype("uint32")
DATASET = "/entry/data/data"


def make_files(tmpdir: Path, n_files: int):
    import h5py

    rng = numpy.random.default_rng(0)
    data_uris, filepaths, files = [], [], []
    for i in range(n_files):
        arr = rng.integers(0, 2**16, size=(F, H, W), dtype=DTYPE)
        fp = tmpdir / f"file{i:05}.h5"
        with h5py.File(fp, "w") as f:
            f.create_dataset(DATASET, data=arr, chunks=(1, H, W))
        files.append(arr)
        filepaths.append(fp)
        data_uris.append(ensure_uri(str(fp)))
    return data_uris, filepaths, files


async def build_node(tmpdir: Path, n_files: int):
    data_uris, filepaths, files = make_files(tmpdir, n_files)
    struct = ArrayStructure(
        shape=(n_files, F, H, W),
        chunks=((1,) * n_files, (F,), (H,), (W,)),
        data_type=BuiltinDtype.from_numpy_dtype(DTYPE),
    )
    pre_chunks = [[1] * (n_files * F), [H], [W]]
    adapter = in_memory(readable_storage=[str(tmpdir)])
    await adapter.startup()
    key = f"ds{n_files}"
    await adapter.create_node(
        key=key,
        structure_family="array",
        metadata={},
        data_sources=[
            DataSource(
                structure_family="array",
                mimetype="application/x-hdf5",
                structure=asdict(struct),
                parameters={"dataset": DATASET},
                properties={"chunks": pre_chunks},
                management="external",
                assets=[
                    Asset(parameter="data_uris", num=i, data_uri=u, is_directory=False)
                    for i, u in enumerate(data_uris)
                ],
            )
        ],
    )
    x = await adapter.lookup_adapter([key])
    full = numpy.stack(files)  # (N, F, H, W)
    return adapter, x, full


async def measure(x, slice_input, reps=5):
    sl = NDSlice(slice_input)
    with patch("tiled.adapters.hdf5.h5open", wraps=tiled.adapters.hdf5.h5open) as m:
        result = await x.read(slice=sl)
    opens = len(m.call_args_list)
    shape = tuple(getattr(result, "shape", ()))
    times = []
    for _ in range(reps):
        t0 = time.perf_counter()
        await x.read(slice=sl)
        times.append((time.perf_counter() - t0) * 1e3)
    return opens, statistics.median(times), shape, result


async def main():
    Ns = [3, 25, 100, 300]
    scenarios = [
        ("whole", (slice(None),)),
        ("single frame [1,5]", (1, 5)),
        ("one time full [1]", (1,)),
        ("2 files [0:2]", (slice(0, 2),)),
        ("bin across ALL files [:,5]", (slice(None), 5)),
    ]
    print(f"\n=== per-file (F,H,W)=({F},{H},{W}) {DTYPE} ===")
    header = f"{'scenario':<28}{'N':>5}{'opens':>8}{'time_ms':>10}  result_shape"
    for n in Ns:
        with tempfile.TemporaryDirectory() as td:
            adapter, x, full = await build_node(Path(td), n)
            print(header)
            for label, sl in scenarios:
                opens, ms, shape, result = await measure(x, sl)
                print(f"{label:<28}{n:>5}{opens:>8}{ms:>10.2f}  {shape}")
            print("-" * len(header))


if __name__ == "__main__":
    asyncio.run(main())
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,9 @@ Write the date in place of the "Unreleased" in the case a new version is release
   optional per-asset `extents` property records each file's length along that
   axis, so a read opens only the files its slice touches; the adapter also
   infers the layout from the structure chunks or grid shape when the property
-  is absent. (#1465)
+  is absent. Datasets served through a `slice`/`squeeze` adapter transform fall
+  back to a full build, since the transform reshapes each file and breaks the
+  mapping from the served axis to whole files. (#1465)
 - Add a new feature that stores a graph of links into the catalog database. Adds strawberry
   as a dependency. Import/search/export of graph links is accomplished through graphql.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,9 @@ Write the date in place of the "Unreleased" in the case a new version is release
   read fetches. The budget is now a modest 16 MiB, keeping small reads cheap
   while full reads still coalesce sensibly. Override with
   `TILED_HDF5_READ_CHUNK_BYTES`. (#1465)
+- Route single-file HDF5 array datasets through the eager (non-lazy) adapter.
+  The lazy path culls no assets when there is only one file and would otherwise
+  turn the whole file into a single coalesced read block. (#1465)
 - Fix the `raw_export` download progress bar, which showed a wrong total (e.g.
   `1,257,333,024/100 bytes`) and did not advance during the transfer. The bar
   now seeds each task's total from the known asset size, and raw-asset downloads
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,9 +46,9 @@ Write the date in place of the "Unreleased" in the case a new version is release
 ### Fixed
 
 - Route single-file HDF5 array datasets through the eager (non-lazy) adapter.
-  The lazy path culls no assets when there is only one file and would read the
-  whole file as one task even for a small slice; the eager adapter honors the
-  file's native HDF5 chunking and reads only the touched chunks. (#1465)
+  With only one file the lazy path can cull nothing, so it offers no benefit
+  over the eager adapter, which already honors the file's native HDF5 chunking
+  and reads only the touched chunks. (#1465)
 - Fix the `raw_export` download progress bar, which showed a wrong total (e.g.
   `1,257,333,024/100 bytes`) and did not advance during the transfer. The bar
   now seeds each task's total from the known asset size, and raw-asset downloads
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,12 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ### Fixed
 
+- Bound the bytes a single HDF5 read task pulls from disk. Coalescing tiny
+  native chunks previously used a 1 GiB budget, so reading one frame of a large
+  dataset fetched up to a gigabyte because a block is also the minimum a partial
+  read fetches. The budget is now a modest 16 MiB, keeping small reads cheap
+  while full reads still coalesce sensibly. Override with
+  `TILED_HDF5_READ_CHUNK_BYTES`. (#1465)
 - Fix the `raw_export` download progress bar, which showed a wrong total (e.g.
   `1,257,333,024/100 bytes`) and did not advance during the transfer. The bar
   now seeds each task's total from the known asset size, and raw-asset downloads
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,8 +29,9 @@ Write the date in place of the "Unreleased" in the case a new version is release
   axis, so a read opens only the files its slice touches; the adapter also
   infers the layout from the structure chunks or grid shape when the property
   is absent. Datasets served through a `slice`/`squeeze` adapter transform fall
-  back to a full build, since the transform reshapes each file and breaks the
-  mapping from the served axis to whole files. (#1465)
+  back to a full build, since the lazy read reshapes each whole file into its
+  slab of the served structure and that transform can make the served per-file
+  shape differ from the raw file. (#1465)
 - Add a new feature that stores a graph of links into the catalog database. Adds strawberry
   as a dependency. Import/search/export of graph links is accomplished through graphql.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,17 +21,18 @@ Write the date in place of the "Unreleased" in the case a new version is release
   `TILED_SEQUENCE_IO_WORKERS` and `TILED_SEQUENCE_READ_BATCH_BYTES` environment
   variables. (#1463)
 - Faster reads of HDF5 datasets spanning many files: the lazy Dask graph is
-  cached through the resource cache, per-file specs are read in parallel, and
-  tiny native chunks are coalesced into whole-file read tasks. (#1463)
+  cached through the resource cache and per-file specs are read in parallel.
+  (#1463)
 - Lazy asset resolution for multi-file HDF5 array datasets, extending the lazy
   path above to datasets that concatenate files along the leading axis. An
   optional per-asset `extents` property records each file's length along that
   axis, so a read opens only the files its slice touches; the adapter also
   infers the layout from the structure chunks or grid shape when the property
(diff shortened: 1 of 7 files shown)


danielballan:
> extents is optional and fully adapter-contained; the catalog stays ignorant of it. 
Does this indicate that `extents` belongs in `parameters`? I am not sure whether we took a clear position on:

A) Items in `properties` have semantic meaning to the catalog/server.
B) Items in `properties` have semantic meaning that generalizes across Adapters.


If (A) then it sounds like `extents` should be a parameter. If (B) then it could stay a property.

genematx:
Interesting. I always thought of `properties` as something _not_required_ for reading the data (e.g. unlike `dataset`), but which might help (e.g. to serve the portions aligned with chunks).
To me, it looks like `extents` belong in `properties` -- they just help to disambiguate (native) `chunks`.

danielballan:
I'll summarize here, for Future Us, and @genematx and I discussed wanting stronger evidence that this is effective before we merge it. It's a large addition. We agree that something roughly shaped like this is needed, but we hesitate to adopt it just yet.

PR #1471, opened 2026-08-17
feat: add check_authn_scopes flag

ZohebShaikh:
### Checklist
- [X] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ Write the date in place of the "Unreleased" in the case a new version is release
   variables. (#1463)
 - Add a new feature that stores a graph of links into the catalog database. Adds strawberry
   as a dependency. Import/search/export of graph links is accomplished through graphql.
+- Add `check_authn_scopes` option to `ProxiedOIDCAuthenticator` (default
+  `True`). Set to `False` to skip Tiled's own scope-based authorization
+  checks, for deployments where authorization is enforced by an external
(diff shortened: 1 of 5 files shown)


danielballan:
Tiled protects each of its routes with a list of required scopes. This is a [standard pattern in FastAPI](https://fastapi.tiangolo.com/advanced/security/oauth2-scopes). The scopes on each endpoint are rendered in the OpenAPI schema.

However, these Tiled application scopes in the OpenAPI schema do not necessarily match the scopes from the OIDC provider (Entra for NSLS-II, Keycloak for DLS). Generally, applications like Tiled need to _translate_ the scopes from OIDC token into their application scopes via a configured mapping.

For example, at NSLS-II, Entra gives us a list of scopes in the JWT's `scp` claim. We map these onto Tiled application scopes like `"read:metadata read:data"`. (As an aside, Tiled's _implementation_ of this mapping can and will shortly be improved, but the logic is sound.)

If I'm reading this PR right, it sidesteps any application scope check. The scopes specified in OpenAPI are just ignored, via an early `return` that bypasses the scope check. It would be up to some intermediate proxy to control which endpoints a request is allowed to reach. There are a few ways that an external proxy could work with the current scheme. We could discuss those in a future call.

danielballan:
Closing, per discussion.

We'll follow up with:
- https://github.com/bluesky/tiled/issues/1476
- https://github.com/bluesky/tiled/issues/1477

PR #1473, opened 2026-08-19
Fix scope extraction precedence

danielballan:
Entra gives us a list of scopes in the JWT's `scp` claim. We map these onto Tiled application scopes like "read:metadata read:data".

We do this by inserting a `"scopes"` key next to the `"scp"` key during the token decoding step.

https://github.com/bluesky/tiled/blob/f4eb100884fc2a338d91bc8224266897b5eb0125/tiled/authenticators.py#L413-L431

But for that to work, we need to give `"scopes"` higher precedence than `"scp"`. That is what this PR fixes.

In a follow-up PR, I think we should refactor this more aggressively. It might be better to avoid combining this "translation" with the token-decoding step. Rather, this scope mapping could perhaps more cleanly be done further downstream.

I consider this small PR a hotfix.

- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 ## Unreleased
 
+### Fixed
+
+- Fix precedence of scope check for `ProxiedOIDCAuthenticator`.
+
+## v0.2.15 (2026-08-11)
+
 ### Added
 
 - Support clustered Redis for high availability.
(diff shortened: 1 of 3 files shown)


danielballan:
This is deployed on `tiled-demo.nsls2.bnl.gov`.

On `v0.2.15` / `main` you can see that Tiled application scopes and Entra OIDC scopes are being mixed up, comparing apples to oranges.

```
ClientError: 401: Not enough permissions. Requires scopes ['read:metadata']. Request had scopes ['access_as_user'] https://tiled-demo.nsls2.bnl.gov/api/v1/search/?page%5Boffset%5D=0&fields=
```

On this PR branch, it works correctly. The Entra scopes are "mapped" into Tiled authentication scopes via the `scopes_map` configuration in the Tiled config file.

genematx (approved):

PR #1482, opened 2026-08-21
📤️ Add a native `xdi` exporter

Kezzsim:
This adds support for exporting `Xray Data Interchange` files as defined by the spectroscopy community: https://github.com/XraySpectroscopy/XAS-Data-Interchange

Registers a serializer that works only on containers that have been explicitly labeled with `[Spec('xdi')]`, supporting versioning for an upcoming revision that will support diffraction gradient based beamlines.


tests/test_xdi_serialization.py
```diff
@@ -0,0 +1,334 @@
...
+    )
+
+
+# ---------------------------------------------------------------------------
+# Tests: plain table (no Spec, no XDI metadata) can still be exported
+# ---------------------------------------------------------------------------
+
+
+def test_plain_table_exports_as_xdi(client):
+    """Any table can be exported as XDI; a plain table produces a minimal valid file."""
```
danielballan: Is this intended? I could be fine either way, but I think I remember a discussion this which concluded, "The XDI exporter should only be offered to users when the dataset has been marked with the `xdi` spec."

Again, fine if we want to reverse that, but I want to be sure we're intentional about it.

tiled/serialization/xdi.py
```diff
@@ -0,0 +1,141 @@
...
+    # --- Comments block ---
+    comments = metadata.get("comments", "")
+    if comments:
+        for line in comments.splitlines():
+            stripped = line.strip()
+            if stripped and not stripped.startswith("#"):
+                output.write(f"# {line}\n")
+            else:
+                output.write(f"{line}\n")
+    output.write("# generated by tiled\n")
```
danielballan: Is it worth specifying the version of tiled here?

tiled/serialization/__init__.py
```diff
@@ -22,6 +22,11 @@ def register_builtin_serializers():
         from ..serialization import table as _table  # noqa: F401
 
         del _table
+
+    if modules_available("pandas", "dask.dataframe"):
```
danielballan: I believe `dask.dataframe` is included here because you are following existing patterns.

The goal of this line is to avoid `ImportErrors` by checking whether a given serializer's specific dependencies are available before trying to import and register that serializer. Since `dask.dataframe` does not appear to be imported in `xdi.py`, it can be omitted here.

```suggestion
    if modules_available("pandas"):
```

danielballan (commented):
Awesome! A couple questions.

tests/test_xdi_serialization.py
danielballan: I notice in the PR description too:

> Registers a serializer that works only on containers that have been explicitly labeled with [Spec('xdi')], supporting versioning for an upcoming revision that will support diffraction gradient based beamlines.

That conflicts with this test.

genematx:
I wonder if this should go to bluesky-tiled-plugins or tiled-site-config and be registered as an additional exporter, not the core tiled. Just a thought...

danielballan:
I had the same thought, and I'm open to either outcome. It's niche compared to the other formats we support in core, but one point in its favor is, it does not introduce any new dependencies.

I think the fact that it's registered against a specific _spec_ might better push it out of core, though. And especially if we register something on BlueskyRuns---that would point to bluesky-tiled-plugins.

tiled/serialization/xdi.py
Kezzsim: Good idea, I'll add that!

tests/test_xdi_serialization.py
Kezzsim: I can add that restriction to this test, this stems from a sanity check I was doing while developing

Kezzsim: This test has been replaced with `test_plain_table_cannot_export_as_xdi` which confirms that `Spec('xdi')` is required

danielballan:
I think we _might_ move this out of core someday, depending on how the next steps play out, but I'm inclined to merge this into core for now to notch the win.

@Kezzsim There is a docs page that uses XDI as a case study. The page has always stuck out as a bit awkward, and I think the content as now covered better by other guides. Would you delete it as part of this PR? https://blueskyproject.io/tiled/explanations/specialized-formats.html

danielballan:
I tested this interactively.

````python

In [2]: from tiled.client import simple

In [3]: c = simple()
Tiled version 0.2.15b4.dev32+ga502a7d6b
http://127.0.0.1:32925/api/v1?api_key=8a458099dc19a284

In [4]: XDI_METADATA = {
   ...:     "xdi_version": "1.0",
   ...:     "extra_version": "GSE/1.0",
   ...:     "fields": {
   ...:         "Element": {"symbol": "Cu", "edge": "K"},
   ...:         "Mono": {"d_spacing": "3.13553", "name": "Si 111"},
   ...:         "Beamline": {"name": "13ID", "collimation": "none", "focusing": "yes"},
   ...:         "Facility": {
   ...:             "name": "APS",
   ...:             "energy": "7.00 GeV",
   ...:             "xray_source": "APS Undulator A",
   ...:         },
   ...:         "Scan": {"start_time": "2001-06-26T22:27:31", "edge_energy": "8980.0"},
   ...:         "Detector": {"I0": "10cm N2", "I1": "10cm N2"},
   ...:         "Sample": {"name": "Cu", "prep": "Cu metal foil"},
   ...:         "Column": {"1": "energy eV", "2": "i0", "3": "itrans", "4": "mutrans"},
   ...:         # extension namespace
   ...:         "GSE": {"EXTRA": "config 1"},
   ...:     },
   ...:     "comments": "# Cu foil Room Temperature\n# measured at beamline 13-ID\n",
   ...: }

In [6]: import pandas

In [7]: XDI_DF = pandas.DataFrame(
   ...:     {
   ...:         "energy": [8779.0, 8789.0, 8799.0],
   ...:         "i0": [149013.7, 144864.7, 132978.7],
   ...:         "itrans": [550643.089065, 531876.119084, 489591.10592],
   ...:         "mutrans": [-1.3070486, -1.3006104, -1.3033816],
   ...:     }
   ...: )

In [8]: t = c.write_table(XDI_DF, metadata=XDI_METADATA)

In [9]: t
Out[9]: <DataFrameClient ['energy', 'i0', 'itrans', 'mutrans']>

In [10]: t.read()
Out[10]:
   energy        i0         itrans   mutrans
0  8779.0  149013.7  550643.089065 -1.307049
1  8789.0  144864.7  531876.119084 -1.300610
2  8799.0  132978.7  489591.105920 -1.303382

In [11]: t.metadata
Out[11]:
{'comments': '# Cu foil Room Temperature\n# measured at beamline 13-ID\n',
 'extra_version': 'GSE/1.0',
 'fields': {'Beamline': {'collimation': 'none',
                         'focusing': 'yes',
                         'name': '13ID'},
            'Column': {'1': 'energy eV',
                       '2': 'i0',
                       '3': 'itrans',
                       '4': 'mutrans'},
            'Detector': {'I0': '10cm N2', 'I1': '10cm N2'},
            'Element': {'edge': 'K', 'symbol': 'Cu'},
            'Facility': {'energy': '7.00 GeV',
                         'name': 'APS',
                         'xray_source': 'APS Undulator A'},
            'GSE': {'EXTRA': 'config 1'},
            'Mono': {'d_spacing': '3.13553', 'name': 'Si 111'},
            'Sample': {'name': 'Cu', 'prep': 'Cu metal foil'},
            'Scan': {'edge_energy': '8980.0',
                     'start_time': '2001-06-26T22:27:31'}},
 'xdi_version': '1.0'}


In [14]: t = c.write_table(XDI_DF, metadata=XDI_METADATA, specs=['xdi'])

In [15]: t.export('test.xdi')

In [16]: !cat test.xdi
# XDI/1.0 GSE/1.0
# Facility.name: APS
# Facility.energy: 7.00 GeV
# Facility.xray_source: APS Undulator A
# Beamline.name: 13ID
# Beamline.collimation: none
# Beamline.focusing: yes
# Mono.d_spacing: 3.13553
# Mono.name: Si 111
# Detector.I0: 10cm N2
# Detector.I1: 10cm N2
# Sample.name: Cu
# Sample.prep: Cu metal foil
# Scan.start_time: 2001-06-26T22:27:31
# Scan.edge_energy: 8980.0
# Element.symbol: Cu
# Element.edge: K
# Column.1: energy eV
# Column.2: i0
# Column.3: itrans
# Column.4: mutrans
# GSE.EXTRA: config 1
# /////////////
# Cu foil Room Temperature
# measured at beamline 13-ID
# generated by tiled 0.2.15b4.dev32+ga502a7d6b
# -------------
# energy i0 itrans mutrans
8779.0 149013.7 550643.089065 -1.3070486
8789.0 144864.7 531876.119084 -1.3006104
8799.0 132978.7 489591.10592 -1.3033816
```

danielballan (approved):

PR #1486, opened 2026-08-24
Add support for access token in tiled client ws

ZohebShaikh:
I was trying to do stacked PRs but I think it does not work from forked branches in UI.

This PR has 3 commits which explain the three things I'm doing in the PR.

This is the first part of the stack https://github.com/bluesky/tiled/pull/1485


danielballan:
Thanks for this. #1485 is merged. I have two questions.

First, is there a benefit to introducing "first message" auth? For me, it would be simpler to pass the credential in the `Authentication` header unconditionally, regardless of whether credential is a JWT or an API key.

For background, which you may already know, the option to pass the credential in the first WS message is necessary _for web browsers_ because they cannot set arbitrary headers in a WS request. But as far as I know there is no reason that the Python client cannot use an `Authorization` header in every case. (I found [this explanation of WS auth options](https://websockets.readthedocs.io/en/stable/topics/authentication.html) helpful when designing this.)

Second, is there a way to make this work without this kind of conditionality?

```python
if self._http_client and isinstance(self._http_client.auth, TiledAuth):
```

I think it can be hard to follow the implications of that logic, and I'm hoping it might not be necessary. Would it work to do this?

```python
def get_auth_header(context):
    if not context.authenticated:
        result = None
    elif context.api_key:
        # Client is configured to use API key auth.
        result = f"Apikey {self.context.api_key}"
    else:
        # Client is has a JWT access token.
        access_token = context.http_client.auth.sync_get_token(
                "access_token", reload_from_disk=True
        )
        result = f"Bearer {access_token}"
    return result

headers = {}
if auth_header := get_auth_header(self.context):
    headers["Authorization"] = auth_header
```

There is one wrinkle need to handle, which applies to both this PR as written and my suggestion above: I believe the WS endpoint has no way to trigger a refresh because `401` is not a valid response from a WS endpoint. Options:

1. If the connection fails, refresh the token before retrying.
2. Parse the JWT and check that it is not about to expire. (This requires parse this JWT client-side, which to date we have not done. It would add a dependency.)
3. Proactively refresh before connecting. (This might result in hammering refresh in the event of many connections. I think we can rule this option out.)

danielballan:
About that unauthenicated branch (`context.authenticated is None`) I need to confirm later whether we even support that. Maybe it should just raise.

I recall thinking that, _even on public servers_, we should require clients to authenticate WS connections, given the load that a long-lived connection places on the server. In the event of abuse, we want to tie that to a username, not just an IP address.

Issue #1310, opened 2026-03-16
Redis is not Unsubscribed when the Client Disconnects


genematx:
### The Problem

When receiving updates via WebSockets, the server process subscribes to Redis (on behalf of the client); e.g. in Redis logs this can be observed as:

```
1773444651.297893 [0 192.168.65.1:57132] "SUBSCRIBE" "notify:1"
1773444651.299970 [0 192.168.65.1:22509] "CLIENT" "SETINFO" "LIB-NAME" "redis-py"
1773444651.299979 [0 192.168.65.1:22509] "CLIENT" "SETINFO" "LIB-VER" "7.0.1"
1773444651.300702 [0 192.168.65.1:22509] "GET" "sequence:1"
...
```

In the end of the client session, when the client disconnects, an unsubscribe even should be posted. This is the seemingly the case with the `TTLCacheDatastore` / `_TestClientWebsocketWrapper` used in tests.

```
...
1773444651.549106 [0 192.168.65.1:28832] "HMGET" "data:1:7" "payload" "metadata"
1773444653.663609 [0 192.168.65.1:57132] "UNSUBSCRIBE" "notify:1"
```

However, with the usual (production) `RedisStreamingDatastore` / `_RegularWebsocketWrapper`, the `UNSUBSCRIBE` event doesn't happen after the client is disconnected. Only after the server is turned off, Redis receives all `UNSUBSCRIBE` events at once.

### A Way to Reproduce

1. Start Redis and Postgres (possibly can use SQLite too) in containers, e.g. use the following docker commands:

```bash
docker run -p 5432:5432 -e POSTGRES_PASSWORD=secret -d docker.io/postgres:16
docker run --name tiled-redis -p 6379:6379 -d redis
```

2. Start a local Tiled server with the following config file:
```config.yml
authentication:
  single_user_api_key: "secret"
  allow_anonymous_access: true
trees:
  - path: /
    tree: catalog
    args:
      uri: postgresql://tiled:secret@localhost:5432/catalog
      init_if_not_exists: true
streaming_cache:
  uri: "redis://localhost:6379"
```

3. Exec into the Redis container:
```bash
docker exec -it tiled-redis redis-cli
```
and start monitoring the log:
```
127.0.0.1:6379> monitor
OK
```

4. Run the following client-side script:
```
client = from_uri("http://localhost:8000", api_key="secret")['test/ws_test']

for i in range(5):
    sub = x.subscribe()
    sub.start_in_thread()
[...]

PR #1495, opened 2026-08-27
Add additional asyncio task that watches for disconnect events when streaming. Otherwise, a server with a subscription would hang on close.

jwlodek:
Without this change, creating a subscription on a server would cause it to hang indefinitely on `close()`, even if the subscription is closed first client side.

Caught in a pytest unit test where a subscription was created on a `SimpleTiledServer`, which caused the
test to hang forever on fixture cleanup.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/server/streaming.py
+++ b/tiled/server/streaming.py
@@ -369,14 +369,15 @@ def _make_ws_handler_common(
         disconnect_task = asyncio.create_task(watch_for_disconnect())
 
         last_sent = 0
-        if sequence is not None:
-            # If a sequence number is passed, replay old data
-            current_seq = last_sent = int(await current_sequence_getter())
-            logger.debug("Replaying old data...")
-            for s in range(sequence, current_seq + 1):
-                await stream_data(s)
         # Finally stream all buffered data into the websocket
         try:
+            if sequence is not None:
+                # If a sequence number is passed, replay old data
+                current_seq = last_sent = int(await current_sequence_getter())
+                logger.debug("Replaying old data...")
+                for s in range(sequence, current_seq + 1):
+                    await stream_data(s)
+
             while not end_stream.is_set():
                 live_seq = await stream_buffer.get()
 
--- a/tiled/server/streaming.py
+++ b/tiled/server/streaming.py
@@ -360,8 +360,6 @@ def _make_ws_handler_common(
                     if message["type"] == "websocket.disconnect":
                         disconnect_code = message.get("code")
                         break
-            except WebSocketDisconnect as exc:
-                disconnect_code = exc.code
             except RuntimeError:
                 # Raised if the socket was already closed by the send path.
                 pass
--- a/tiled/server/streaming.py
+++ b/tiled/server/streaming.py
@@ -407,7 +407,14 @@ def _make_ws_handler_common(
         finally:
             live_task.cancel()
             disconnect_task.cancel()
-            await asyncio.gather(live_task, disconnect_task, return_exceptions=True)
+            try:
+                await asyncio.gather(live_task, disconnect_task, return_exceptions=True)
+            except asyncio.CancelledError:
+                # In case a second cancel arrives while cleaning up, ignore it.
+                # The tasks are already cancelled and we don't want to raise an exception here.
+                # This was observed in tests using Starlette's TestClient, which cancels
+                # all tasks on exit.
+                pass
 
     return handler
 
--- a/tiled/server/streaming.py
+++ b/tiled/server/streaming.py
@@ -217,6 +217,12 @@ class PubSub:
 # so the client reconnects and replays any sequences missed during the outage.
 _LIVE_SUBSCRIPTION_LOST = object()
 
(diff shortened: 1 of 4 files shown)


jwlodek:
Minimal repro:

```Python
import time

from tiled.server import SimpleTiledServer

server = SimpleTiledServer()
client = from_uri(server.uri)

# Create subscription and wait for it to connect.
node = client.write_array([1, 2, 3], key="x")
subscription = node.subscribe().start_in_thread()
time.sleep(1.0) 

t0 = time.monotonic()
server.close() # Before the fix, this line hangs forever.
print(f"server.close() returned in {time.monotonic() - t0:.2f}s")
```

Note that this also happens even if we first disconnect the subscription via `subscription.disconnect()` before calling `server.close()`.

genematx:
Thank you, @jwlodek ! This might be related: #1310

jwlodek:
Yes seems like that could be related

danielballan (approved):

tiled/server/streaming.py
```diff
@@ -335,6 +341,35 @@ async def buffer_live_events():
...
+
+            The send loop below blocks on an idle buffer and only sends; it never
+            reads. A client disconnect or a server shutdown arrives as a
+            websocket.disconnect receive event, so without draining ``receive()``
+            here the handler would never notice the socket is gone and would
+            keep the ASGI task (and thus graceful shutdown) alive forever.
+            """
+            nonlocal disconnect_code
+            try:
+                while True:
```
gwbischof, line 358: Is there any way to detect that the client disconnected without adding this Loop?

tiled/server/streaming.py
```diff
@@ -335,6 +341,35 @@ async def buffer_live_events():
...
+                    message = await websocket.receive()
+                    if message["type"] == "websocket.disconnect":
+                        disconnect_code = message.get("code")
+                        break
+            except WebSocketDisconnect as exc:
```
genematx: Claude says Starlet doesn't raise `WebsocketDisconnect`, it just returns `({"type": "websocket.disconnect", "code": ...}`. Might be a dead code.

tiled/server/streaming.py
```diff
@@ -347,6 +382,12 @@ async def buffer_live_events():
             while not end_stream.is_set():
                 live_seq = await stream_buffer.get()
 
+                if live_seq is _PEER_DISCONNECTED:
```
genematx, line 384: This is pre-existent, but if the client disconnect _during_ the retry loop and the `WebSocketDisconnect` is raised on line 389, and finally in line 409 is never reached (resulting in leaking tasks).

danielballan:
Thanks! Conceptually good, and the implementation looks idiomatic to me.

I'm doing some reading around why CI fails on Python 3.10-3.12. Looks like some subtlety on how internal and external cancelations interact. The cleanup needs a closer look.

tiled/server/streaming.py
jwlodek: Do you want me to address this in this PR, or leave to a follow up?

tiled/server/streaming.py
jwlodek: Thanks, will remove.

tiled/server/streaming.py
jwlodek: I couldn't think of a way to do this, since ultimately you have two blocking calls - `websocket.recieve()` and `stream_buffer.get()`. I think they have to be on two independent loops, unless I'm missing something? The loop is also needed because we only want to break on a disconnect event, so we need to be able to continue if non-disconnects are received.

tiled/server/streaming.py
genematx: I am guilty of putting too much into my PRs, so I personally don't mind if it's a quick fix in this one, while we are already here (Dan might disagree). Otherwise, if you don't mind just making an issue, that would be great!

tiled/server/streaming.py
gwbischof: Got it, that makes sense. I'm not totally sure, but I think the only message the client will send to the server is the disconnect, so the loop will only ever run once. So I'm on board with the loop because it shouldn't be looping lots of times, and we would probably need this if we let the client send other messages to the server. (For example, to request multiple data streams over the websocket or to downsample, etc.)

tiled/server/streaming.py
jwlodek: Moved the replay loop into the try/except.

jwlodek:
@danielballan it seems on the older python versions one of the starlette testing context managers would cancel all tasks on exit, which was recancelling our already cancelled tasks. I added a try/except around it and that seems to have resolved it

danielballan:
Ah, that makes sense. I had fallen down a rabbit hole looking for changes to asyncio itself.

genematx (approved):

jwlodek:
Any objections with clicking the merge button?

danielballan:
Aiming to merge this today. Will resolve the CHANGELOG merge conflicts once one other PR is in.

PR #1497, opened 2026-08-27
Address regression in API key auth for UI

danielballan:
Launching Tiled like this, just for example,

```
tiled serve catalog --temp
```

prints a URL in the terminal with an `?api_key=...` query parameter. Formerly, clicking this link and opening the React UI in the browser "just worked." It worked because the server "moves" the API key from the query parameter to an `HttpOnly` cookie, which authenticates subsequent requests from the React UI.

https://github.com/bluesky/tiled/blob/24f9bb30c118c7e0262324316de257ee1f7b640f/tiled/server/authentication.py#L331-L344

The addition of the login redirect broke this, by redirecting users to the login screen even when a cookie was in place with the API key. This PR fixes that by testing an authenticated request to (`whoami`) to detect the presence of the cookie. It also updates the Log In / Log Out UI to reflect this, and clears the cookie on logout.

Finally, it incorporates the frontend in CI. I think it will totally omitted from the CI up to this point.

~~I have not yet tested this interactively.~~

AI disclosure: Opus 4.8 with detailed planning

### Checklist
- [ ] Add a Changelog entry

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,7 +109,7 @@ jobs:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   web_frontend_checks:
-    name: WebUI Type Check and Tests
+    name: WebUI Type Check, Tests, and Build
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -133,6 +133,9 @@ jobs:
 
     - name: Test
       run: npm run test:run
+
+    - name: Build
+      run: npm run build
 
   windows_checks:
     name: Test on Windows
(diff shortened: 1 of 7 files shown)


genematx (dismissed):
Just tightened up the cookie-session fallback, some tests and CI: https://github.com/danielballan/tiled/pull/11

danielballan:
Thanks, @genematx. I merged your PR into this PR branch.

I tested this locally. The login button stays set at "log in." It should probably say "log out."
image

genematx:
Is this in single-user mode?

genematx:
here's a [possible fix](https://github.com/danielballan/tiled/pull/12); tested locally in single- and multi-user modes. In the first case, no Log In button is displayed at all, in the second, case, after logging in, the username is displayed (as before). Please double-check that it works in your setup.

danielballan:
It works on my setup. Thanks for running with that.

Issue #1496, opened 2026-08-27
On Uniqueness and Naming of Graph Entities Attributes


genematx:
A few thoughts that occurred to me while I've been working on implementing a client interface for the graph of links.

In the `entities` table should we impose a uniqueness constraint on `(node_id, entity_type, name)`? In the absence of the DB-imposed constraint, we would have to check if an entity already exists every time we would be trying to create it -- degrading performance and introducing a possibility for a race condition. More specifically, this would allow us to (easily) create unambiguous canonical entities for Tiled nodes, e.g. by setting `entity_type` to one of `{'array', 'table', 'container'}` and using node key as the name. The entity uniqueness (or lack thereof) is a more general design consideration though. (The user can always create more entities with difeerent types/names, of course.)

Furthermore, should we consider renaming "entity_type" to simply "kind"? It's much shorter, and it's already understood from the table name ("entities") what object the type relates to. Then, on the client side, the API becomes pretty user-friendly, e.g. `array_client.create_entity(name=..., kind=...) -> EntityHandler`, but that's a separate story.

Wondering what your thoughts on this are, @dylanmcreynolds, @danielballan.

PR #1499, opened 2026-08-27
Enforce Uniqueness of Certain Graph Attributes

genematx:
Two related changes to the experimental graph entities table:

**Rename `entity_type` → `kind`.** Shorter, and the `entities` table name already conveys what the type refers to. The rename covers the database column, the GraphQL API (`entityType` → `kind`), the store, example config, tests, and docs. A migration renames the column via a portable `ALTER TABLE ... RENAME COLUMN` (applied on both PostgreSQL and SQLite) plus the supporting index; the historical migration that first created the column is left untouched.

**Enforce entity uniqueness per node.** A unique index on `(node_id, kind, name)` guarantees at most one entity for a given node, kind, and name — avoiding a check-then-insert round trip and the associated race condition, and enabling canonical per-node entities (e.g. `kind` in `{array, table, container}` keyed by node key). Node-bound entities are deduplicated; free-standing (external) entities have `node_id` NULL and are left unconstrained, since NULLs are distinct in a unique index. The store raises `EntityConflictError` on a conflicting insert/update, which the `createEntity`/`updateEntity` mutations surface as an `ENTITY_EXISTS` error.

Issue: #1496

### Checklist
- [x] Add a Changelog entry


dylanmcreynolds:
I intentionally designed node to entity as zero to many. Enttiies have a `properties` member tha tI envision as being unique the triple that they're in.

genematx:
Yes, the relationship will still be zero-to-many (on the FK-level). I am afraid having several entities with the same _type_ and _name_ related to the same _node_ might become confusing. E.g. if I want to create an entity of 'type' called 'name' bound to a given node, I would have always check first if such an entity already exists (and potentially also check all their properties too). This sounds excessive to me, and there's a potential for race condition (although I'm looking at this mostly from the SQL point of view -- maybe it's a standard practice in GraphQL).
If a user needs multiple entities bound to the same node, it's always easy to use different names (not just properties).

dylanmcreynolds (commented):
I really don't think we want to enforce uniqueness of one entity per node. Different graphs will be built relating to the same node, and each might have it's own set of `properties` and `access_blob`.

This PR prompted some discussion about uniqueness. How about saving and exposing a `uuid` field? Then clients can know exactly which entity they have.

genematx:
Yes, I see how different graphs can be reusing the same node, and that users may want to keep the same name for them.

I am afraid we might quickly run into a problem when it would be hard to tell which exact entity I am looking for out of a few already existing with the same type and name (the uniqueness of `properties` can not guaranteed either, so even they can be the same, and `access_blob` is inherited from the Nodes table anyway, so everyone would see the same set of entities). Of course, this means that every time a user is about to create a new entity, they should first check if it already exists -- this is something that the DB engine should be doing, ideally.

I personally think that relying on uuids is a good idea, but unfortunately they are not human-friendly, and I expect significant push back from the users (as I am currently seeing with regard to Bluesky uuids).

(Note that for external entities, that would have `node_id=None`, one can still create multiple ones with same name/type).

dylanmcreynolds:
> I personally think that relying on uuids is a good idea, but unfortunately they are not human-friendly, and I expect significant push back from the users (as I am currently seeing with regard to Bluesky uuids).

We would still keep the `name` field for display, with the UUID in the graphql api for applications to use.

genematx:
I thought uuid was there, in the schema, already. it definitely shouldn't go anywhere.

PR #1504, opened 2026-09-01
Fix the config parser for adapters_by_mimetype

dylanmcreynolds:
### Checklist
- [ ] Add a Changelog entry

The config parser defines a catalog's `adapters_by_mimetype` as a list of entries. It should be a dict with the mimetype as a key.


danielballan:
Thanks! Since I do not have permission to push to PR branches from `als-computing`, I opened a new PR.

Superseded by https://github.com/bluesky/tiled/pull/1506.

Issue #1461, opened 2026-08-05
Investigation into challenges reading HDF5 dataset while it is being extended


danielballan:
Following up on a report from @abbiemery, I made an attempt at a small reproducer:

Create extensible HDF5 dataset.

```python
import h5py, numpy
file = h5py.File("test.h5", "w", libver="latest")
file.swmr_mode = True
file.create_dataset("data", data=numpy.ones((2, 3)), maxshape=(None, 3))
```

Index it with Tiled and start a server.

```sh
TILED_HDF5_SWMR_DEFAULT=1 tiled serve directory . --api-key=secret
```

Access it via Tiled; extend (resize) the dataset; access it again.

```python
from tiled.client import from_uri
c = from_uri('http://localhost:8000', api_key='secret')
c['test']['data'][:]  # works

ds.resize((4, 3))
ds[:] = numpy.ones((4, 3))

c['test']['data'][:]  # slices out (2, 3) unbothered by new data
```

I suspect the problem you see at  DLS _is_ a real bug, but that I'm missing the essential issue. Can you show more context about what goes wrong?

jacobfilik:
So I think what happens is force_reshape in the ArrayAdapter fails because the shape of the dataset is potentially different from the shape stored in the metadata, so you can't force it to be the same.

I think we would expect that the dataset in the file will always be the same shape, or larger in any dimension than tiled expects, but never smaller (messages sent when data is flushed, gpfs, posix stuff...).

I think a solution could be to not fail the force_reshape if all dimensions are >= what tiled expects, just return the arr (which I assume is most likely a lazy representation, but might not be), but then in the read method, if no slice is specified, build the slice so it only returns data self._structure.shape (even though there may be more data on disk). Either that or slice the array to _structure.shape in force_reshape (although previously that function was a metadata only operation - but maybe it still would be for numpy and dask?)

Although it might be useful to "return all the data from this dataset" from a method, even if its larger than what tiled expects, i don't think it is worth considering now.

danielballan:
Thanks, @jacobfilik, and sorry for the delayed follow up. Many of us have been on work travel recently. That explanation sounds plausible.

We'll dig in.

genematx:
Thank for raising this issue and apologies for the delayed response. I was trying to make sense of it and reproduce the error (with Claude's help). Here's what I think a comprehensive example. Please have a look; does it accurately reflect what you're seeing? Please note the docstring explaining the differences in registering container vs array.

[...]

Issue #1314, opened 2026-03-23
Shape mismatches when reading new frames while streaming


canismarko:
I am often getting HTTP 500 responses when I try to read a frame of an array after being notified over websockets that a new frame is available.

I have a javascript app that listens to a websocket on `/api/v1/stream/single/{path}`. When a new frame is received by Tiled, the client gets a websocket message with `{type: "array-ref"}`. 

The client then does 2 sequential GET requests over HTTP: 1) GET the zarr.json data to get the updated shape (using the zarrita package), then 2) GET the data for the new frame and add it to the live plots.

If the data are coming from an internal array, everything is fine, but if it's an external array I often get a 500 error during the HTTP request, and the Tiled log (full log below) shows that the array on disk is bigger than Tiled was expecting: `ValueError: Can not reshape (119, 13, 4096) array data to (122, 13, 4096)`. The first number (`119`) matches what the shape is expected to be based on the zarr.json file (or `/api/v1/metadata/{path}` if I read that instead).

I've tried this with reading array data from `/api/v1/array/full/{path}` instead of  `/zarr/v3/{path}/c/{block}`, with similar results. I did not try `/api/v1/array/block/{path}`.

This behavior is not 100%: sometimes I get lucky and the client is able to retrieve the new frame. If I'm running using simulated hardware following [the ophyd-async docs](https://blueskyproject.io/ophyd-async/main/tutorials/using-devices.html). then the HTTP requests succeed more often than they fail. If I am using real hardware, then the failure is more common than a success. If I'm reading from a different Tiled server than the detectors are writing to, then the failure is still common but the difference in shape seems to be bigger.

This seems like a race condition, but one I don't totally understand. If I try and read everything after the scan is finished, then it all works.

SWMR is turned on for the detector writer. All the devices that write these data are ophyd-async detectors.

Here is an example of the Tiled logs that are written while this is happening:

```python
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/home/beams0/S25STAFF/src/cortex/backend/.pixi/envs/default/lib/python3.13/site-packages/uvicorn/protocols/http/httptools_impl.py", line 416, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        self.scope, self.receive, self.send
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
[...]

PR #1505, opened 2026-09-01
Fix Array Shape Desynchronization

genematx:
## Summary

Reading an array can fail when the shape recorded in the catalog structure has diverged from the shape of the data actually present in storage. This happens while an array is being extended out-of-band (e.g. streaming appends), because the write path is non-atomic and observable out of sync: the file and the catalog `structure` row are updated separately, and a reader can observe them in either order.

Previously both directions surfaced as an opaque `500`. This PR handles them explicitly.

## Behavior

- **Storage ahead of structure** (data file has grown along the leading axis, trailing dimensions unchanged): Tiled serves exactly the shape the structure advertises, silently ignoring the extra frames, instead of failing. (#1461)
- **Structure ahead of storage** (the advertised shape cannot be served without fabricating data): Tiled distinguishes a transient lag from a permanent mismatch.
  - If the mismatch is confined to the leading axis **and** a backing asset was modified recently (a streaming append is likely still in progress), the read returns a retryable **`503 Service Unavailable`** with a `Retry-After` header. The Tiled client retries these automatically, so a live read succeeds once the writer catches up.
  - If the backing asset has not been modified for a while (the data appears to be at rest and will not catch up), or the mismatch is otherwise unresolvable (e.g. trailing dimensions disagree), the read returns **`422 Unprocessable Content`**. (#1314)

The freshness window is configurable via `TILED_ARRAY_NOT_READY_WINDOW` (default `10` seconds) and the retry hint via `TILED_ARRAY_RETRY_AFTER` (default `1` second).

Note: the Zarr adapter that writes Tiled-native data was never affected (it trims with a stencil rather than reshaping); the affected adapters are those that register external assets and serve through `force_reshape` (HDF5, TIFF, npy, JPEG, file sequences).

## Implementation

- `tiled/adapters/utils.py`: introduce a `ShapeMismatchError` hierarchy — `DataNotReadyError` (transient) and `IncompatibleShapeError` (permanent) — and rework `force_reshape` to trim a leading-axis overshoot, preserve genuine equal-size reshapes, and otherwise raise the appropriate typed error.
- `tiled/storage.py`: add `mtime_from_uri`, mirroring `size_from_uri`, returning an asset's last-modified time for both `file://` and object-store URIs.
- `tiled/catalog/adapter.py`: on `DataNotReadyError`, consult asset freshness via `mtime_from_uri` and downgrade to a permanent error when the data is at rest.
- `tiled/server/router.py`: map `DataNotReadyError` to `503` + `Retry-After` and other shape mismatches to `422`.

## Tests

- `force_reshape` unit tests (trim, genuine reshape, transient vs permanent typed errors).
- HTTP-level tests asserting `503` for structure-ahead and `422` for incompatible structures.
- Catalog-level freshness tests: a recently-modified backing file is treated as transient; a stale file is treated as permanent.

Related Issues: #1314, #1461

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/adapters/utils.py
+++ b/tiled/adapters/utils.py
@@ -79,6 +79,28 @@ def init_adapter_from_catalog(
     return adapter_cls(structure=data_source.structure, **kwargs)
 
 
+class ShapeMismatchError(ValueError):
+    """Array data cannot be served at the shape the structure advertises."""
+
+
+class DataNotReadyError(ShapeMismatchError):
+    """Transient mismatch: the structure is ahead of the data.
+
+    The advertised shape exceeds the available data along the leading axis
+    only, with all trailing dimensions matching. This typically means frames
+    are still being written and the data will catch up, so the read may be
+    retried.
+    """
+
+
+class IncompatibleShapeError(ShapeMismatchError):
+    """Permanent mismatch: the data cannot be reshaped to the advertised shape.
+
+    Retrying will not help: the trailing dimensions disagree, or the total
+    sizes are otherwise incompatible.
+    """
+
+
 def force_reshape(
     arr: Union[np.array, dask.array.Array], desired_shape: Tuple[int, ...]
 ) -> Union[np.array, dask.array.Array]:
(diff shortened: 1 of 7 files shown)


tiled/adapters/utils.py
```diff
@@ -98,14 +120,44 @@ def force_reshape(
...
-    if arr.shape == tuple(desired_shape):
+    desired_shape = tuple(desired_shape)
+
+    if arr.shape == desired_shape:
         return arr  # Nothing to do here
 
+    same_rank = len(arr.shape) == len(desired_shape)
+    trailing_match = same_rank and arr.shape[1:] == desired_shape[1:]
+
+    # The data in storage has grown along the leading axis beyond what Tiled's
```
genematx, line 131: I don't like that we implicitly make the left-most dimension special. It is a valid assumption in Bluesky, but possibly Tiled should be more general. On the other hand, it is best effort strategy for a very common special case with a sensible fallback (error if can't reshape).

danielballan:
The 503 gave me pause because I associate that with, "The backend is down!" But, reading the [description on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503), this is an appropriate use, and the error message does make it clear that this is a narrow failure, not an outage.

I thought of [425 Too Early](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/425), but it means something specific and is not well supported by browsers.

tiled/adapters/utils.py
danielballan: Per discussion, this seems to be safe for numpy-style `reshape`. Any future endpoints that support more complex reshaping (like a snake scan) will need to engage with this---perhaps with a `strict` flag or structural info inline.

danielballan:
Discussed with @jacobfilik, @abbiemery, @ZohebShaikh this morning. We believe this will address the issue that Jacob has seen.

His use case is "storage ahead of structure," the data on disk being _larger_ than the slice requested/expected.

PR #1506, opened 2026-09-02
Fix the config parser for adapters_by_mimetype

danielballan:
This is #1504, rebased and with a changelog entry added.

### Checklist
- [x] Add a Changelog entry


danielballan:
#1504 branched off of a slightly old commit and relied on a stale example that had been removed from `main`. I updated to use an extant example.

PR #1509, opened 2026-09-04
Applied desync'd array strategy from array endpoint to zarr endpoints.

canismarko:
This PR applies the same strategy as #1505, but for the Zarr endpoints.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@ Write the date in place of the "Unreleased" in the case a new version is release
 
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- Extend to zarr routes the previous fix for reads of array data whose
+  on-disk shape has diverged from the shape recorded in the catalog
+  structure, which can happen while an array is being extended
+  (e.g. streaming appends).
+
 ## v0.2.18 (2026-09-02)
 
 ### Fixed
--- a/tests/test_zarr.py
+++ b/tests/test_zarr.py
@@ -242,10 +242,9 @@ async def test_zarr_structure_ahead_of_data_503(prefix, path, app):
 @pytest.mark.parametrize("path", ["/pathological/structure_misaligned"])
 @pytest.mark.asyncio
 async def test_zarr_incompatible_structure_returns_422(prefix, path, app):
-    # Serve an array whose stored structure claims more frames than the
-    # underlying data has, mimicking a catalog that is ahead of the file
-    # during a streaming append. The read should surface as a retryable
-    # HTTP 503, not an opaque 500.
+    # Serve an array whose stored structure is incompatible with the data in a
+    # way that retrying cannot fix (trailing dimensions disagree). This is a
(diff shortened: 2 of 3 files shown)


canismarko:
~~This still needs some tests that exercise the zarr endpoints to get 503 and 422 errors.~~ 

Update: tests have been added.

genematx:
Thank you @canismarko, for putting this together! I am curious, which use case or workflow prompted this fix? The difference between hdf5 and zarr data sources is that the latter are written by an _external_ process and are only registered in Tiled, which may cause mismatches between the assets and the catalog. In contrast, zarr is written by Tiled itself (although, I believe you can register them too), and ideally those assets should always be in sync with the catalog.

Would you mind providing a bit more background information about the error you've seen, please? Your PR may still be necessary (to address the case when zarr is written externally), but I wonder if there's another bug (with internal writing) that we might be missing.

canismarko:
Hi, @genematx. Sure thing.

I'm actually still reading from HDF5 data sources, but accessing them using the zarr endpoints. This PR only changes the zarr router (`tiled/zarr.py`) for serving those endpoints, not the zarr adapter (`tiled/adapters/zarr.py`) that reads zarr from storage.

Here's my workflow that triggers the problem. During a scan, the ophyd-async detector (named e.g. "vortex_me4") writes frame N to an HDF5 file on a mounted NFS share from computer A, and bluesky-tiled-plugins writes the stream resource doc for frame N to Tiled, which is running on computer B. Tiled then sends a websocket message to the browser with the update to the array shape. The browser gets the websocket message and requests the new frame from Tiled at the endpoint `/zarr/v3/<run_uid>/primary/vortex_me4/c/N/0/0`. The Tiled hdf5 adapter opens the HDF5 file from computer B's mounted NFS share but since there is a delay in the NFS transfer, it sees the wrong shape and raises `DataNotReadyError`. If this were the `/api/v1/array/` endpoint, this would result in the `503` response, but since the zarr router does not handle `DataNotReadyError`, it produces a generic `500` response. This PR fixes that so that `zarr/v3/<run_uid>/primary/vortex_me4/c/N/0/0` also returns the same `503` response as `/api/v1/array/full/<run_uid>/primary/vortex_me4/` (same for `422`).

When I first noticed this problem, I was using `/api/v1/array/full`. Since I didn't really understand what was going wrong, I tried the `/zarr/v3/` endpoints to see if it fixed the problem. It didn't help, but I like the zarr endpoints because it meant I could use zarrita to handle the request and deserialization, and deserializing the JSON from `/api/v1/array/full/` was too slow.

I think the 503/422 responses make sense, and my thought with this PR was to have the zarr endpoints behave the same as the other array endpoints.

genematx:
Thank you for explaining, Mark! It all makes sense now (I must admit, I haven't carefully read the code before making my comment); the way you've fixed zarr _endpoints_ makes perfect sense.

genematx (approved):

Issue #1469, opened 2026-08-13
Enable navigation to new origin from the web UI after login


SeijDeLeon:
https://github.com/bluesky/tiled/blob/f4eb100884fc2a338d91bc8224266897b5eb0125/web-frontend/src/routes/auth-callback.tsx#L36

This line uses react router's `navigate` to go to an optional path found in the state parameter after logging in. Could we modify this to allow for a redirect to a different origin instead of just a path off the Tiled UI?

```js
if (state) {
      try {
        const redirectUrl = new URL(state);

        // Preserve callback params so a listener from that origin can see them and save them for future use
        // refresh_token, identity.*, principal, etc.
        for (const [key, value] of params.entries()) {
          // Don't forward state itself unless the receiving app needs it.
          if (key !== "state") {
            redirectUrl.searchParams.set(key, value);
          }
        }

        window.location.replace(redirectUrl.toString());
        return;
      } catch {
        // Invalid state URL; fall through to the normal Tiled UI.
      }
    }
```

Allowing arbitrary origin redirect from `state` lets another browser client who initiated the auth flow via a popup keep track of when the login is complete, at which point it can save the tokens and close the popup (we do this flow in Tiled-backed apps at the ALS). This would solve the issue described in #1280

PR #1513, opened 2026-09-10
Allow state redirect to external origin

SeijDeLeon:
Addresses #1469 

When using `redirect_on_success: http://tiled.localhost:8000/ui/auth-callback`, a state parameter gets tracked throughout the auth flow and is eventually used in the `auth-callback` route. This enhancement allows the state parameter to be an arbitrary url, so an external web app could initiate the flow via a popup and return to its origin at the end, allowing the tokens to be stored in the external web app.

### Checklist
- [ ] Add a Changelog entry


AditiChikkali (approved):

danielballan (approved):

danielballan:
CI failures are due to an unrelated and known issue.

PR #1514, opened 2026-09-11
Convert from access_blobs to access_tags

### LINKED ISSUE #1414: AuthZ rework tracking issue
We have observed performance issues when users have access to a long list of tags, such as a beamline scientist with access to a long list of proposals. To optimize those queries, we need to move the principal/tag/scope association (which we'll call the "grants" table) into the catalog database.

This is the order of operations:

1. Move access_blob to separate table
2. Convert access_blob to access_tags
3. Handle user-owned nodes as blob goes away
4. Add support for multiple providers
5. Move compiled tags into catalog database
6. #1393 

A sketch of the new tables in the `catalog` database:

```sql
CREATE TABLE access_tags (
   id PRIMAY KEY AUTO INCREMENT,
   tag VARCHAR(32)
);
```

```sql
CREATE TABLE nodes_access_tags_association (
  tag_id FOREIGN KEY tags(id),
  node_id FOREIGN KEY nodes(id),
);
```

```sql
CREATE TABLE grants (
  tag_id FOREIGN KEY tags(id),
  scope ENUM(...),
  principal_uuid UNICODE(36), --FOREIGN KEY via FDW to AuthN database?
);
```

Auto-generate user tags that look like `principal:<uuid4>` in the database but present as something more readable in the HTTP API and in the configuration file.
- Configuration can use the language of identity, rather than principal, and set the provider at the top, globally (overridable case by case). Principals will still be used in places, for SPs.
- The HTTP API can translate the tag into something with special meaning, like `me` when `principal:<uuid4>` matches the principle of the current request.

nmaytan:
This PR is stacked on top of #1423 and will close #1414. It converts the `access_blob` associations (from the prior PR) into `access_tags` associations. In total, it takes `access_blob` out of `Nodes`, moves `AccessTags` into their own tables, and is a significant change to the AuthZ mechanics.

Overall, the changes are expected to be a significant performance improvement. In our testing, some operations which previously timed out (took greater than 60 seconds) have come down to only a few seconds (for example, catalog `repr` as a user with broad access). Other operations went from seconds to 100's of milliseconds (such as `len(catalog)` for some users).

Though the vast majority of tests have shown good improvement, it is possible that there is _some_ regression in speed in select cases. One case we see is in waiting for a non-existent catalog entry to return a 404 (perhaps 2x slower). However, this new implementation clears the way to build upon and make additional improvements after monitoring query statistics.

### Summary
- The `access_blob` column has been removed from the Nodes table, and has been formalized more strictly to be replaced by `access_tags`.
- Node to Tag associations now live in a separate table, instead of as a Nodes column.
- The `AccessTagsCompiler` now points at the catalog database, instead of the sidecar SQLite database.
- The catalog will now hold the table of all access tags and the association table of all grants, as well as of tag owners.
- The compiler now create "principal tags" (or "user tags") for each Tiled user. This replaces the "user-owned access blob". Note: in order to compile these tags, the compiler needs a connection to the AuthN database.
- Entity-Tag and Link-Tag association tables also exist, as graphs use the same table of access tags for access control.

### Some additional detail
- The new server is still back-compatible with clients that speak in terms of `access_blob`. 
- The `TagBasedAccessPolicy` now gives implicit grants to users equal to their principal tag.
- Tags which are removed from the tag definitions/config, but which are associated with a node, will not be deleted by the tag compiler. Instead, these are left with all grants removed. This prevents unrecoverable erasure of access control information, especially in the case of a transient failure.
- `access_tags` are no longer stored in metadata revisions when changed
- The `AccessTagsParser`, and the `AccessTagsCompiler` now work on both SQLite and Postgres
- The `AccessTagsParser` no longer needs to be given a database URI - it finds it from the server tree
- The access policy is now allowed to have fewer scopes than catalog DB knows, by intersection. Enables, for example, switching to a read-only policy.
- Various other improvements and bug fixes

### Note about catalog database migraitons
This PR contains 3 database migrations: first moving to an intermediate step where `access_blob` was itself moved into an association table, second converting the blob to tags, and third adding a `parent_id` column to the Node-Tag association table. **It is intended that a full migration be done, not stopping at any intermediate step.**

Additionally, the old metadata index is dropped in this PR. Instead, #1521 creates an improved index. One should therefore be sure to also apply this migration.

### Checklist
- [x] Add a Changelog entry

### DIFF
--- a/tiled/access_control/protocols.py
+++ b/tiled/access_control/protocols.py
@@ -14,7 +14,7 @@ class AccessPolicy(ABC):
         authn_access_tags: Optional[AccessTags],
         authn_scopes: Scopes,
         access_blob: Optional[AccessBlob] = None,
-    ) -> Tuple[bool, Optional[AccessBlob]]:
+    ) -> Tuple[bool, AccessBlob]:
         pass
 
     async def modify_node(
@@ -24,8 +24,8 @@ class AccessPolicy(ABC):
         authn_access_tags: Optional[AccessTags],
         authn_scopes: Scopes,
         access_blob: Optional[AccessBlob],
-    ) -> Tuple[bool, Optional[AccessBlob]]:
-        return (False, access_blob)
+    ) -> Tuple[bool, AccessBlob]:
+        return (False, access_blob or AccessBlob(tags=[]))
 
     @abstractmethod
     async def allowed_scopes(
--- a/tiled/access_control/protocols.py
+++ b/tiled/access_control/protocols.py
@@ -3,7 +3,16 @@ from typing import Optional, Tuple
 
 from ..adapters.protocols import BaseAdapter
 from ..server.schemas import Principal
-from ..type_aliases import AccessBlob, AccessTags, Filters, Scopes
+from ..type_aliases import Filters, Scopes
+
+
+class AccessTags(frozenset[str]):
+    def __new__(cls, tags=()):
+        if isinstance(tags, str):
+            raise TypeError(
+                "AccessTags expects an iterable of strings, not a single string."
+            )
+        return super().__new__(cls, tags)
 
 
 class AccessPolicy(ABC):
@@ -13,8 +22,8 @@ class AccessPolicy(ABC):
         principal: Principal,
         authn_access_tags: Optional[AccessTags],
         authn_scopes: Scopes,
-        access_blob: Optional[AccessBlob] = None,
-    ) -> Tuple[bool, AccessBlob]:
+        access_tags: Optional[AccessTags] = None,
+    ) -> Tuple[bool, AccessTags]:
         pass
 
     async def modify_node(
@@ -23,9 +32,9 @@ class AccessPolicy(ABC):
         principal: Principal,
         authn_access_tags: Optional[AccessTags],
         authn_scopes: Scopes,
-        access_blob: Optional[AccessBlob],
-    ) -> Tuple[bool, AccessBlob]:
-        return (False, access_blob or AccessBlob(tags=[]))
+        access_tags: Optional[AccessTags],
+    ) -> Tuple[bool, AccessTags]:
+        return (False, access_tags or AccessTags())
 
     @abstractmethod
     async def allowed_scopes(
--- a/tiled/access_control/protocols.py
+++ b/tiled/access_control/protocols.py
@@ -1,10 +1,11 @@
 from abc import ABC, abstractmethod
-from typing import Optional, Tuple
+from typing import Iterable, Optional, Tuple
+
+from typing_extensions import Self
 
 from ..adapters.protocols import BaseAdapter
 from ..server.schemas import Principal
 from ..type_aliases import Filters, Scopes
-
 
 # Prefixes of principal tags ('user:<id>', 'service:<uuid>'), which mark
 # nodes as owned by a single principal.
@@ -12,7 +13,7 @@ PRINCIPAL_TAG_PREFIXES = ("user:", "service:")
 
 
 class AccessTags(frozenset[str]):
-    def __new__(cls, tags=()):
+    def __new__(cls, tags: Iterable[str] = ()) -> Self:
         if isinstance(tags, str):
             raise TypeError(
                 "AccessTags expects an iterable of strings, not a single string."
--- a/tiled/access_control/protocols.py
+++ b/tiled/access_control/protocols.py
@@ -6,6 +6,11 @@ from ..server.schemas import Principal
 from ..type_aliases import Filters, Scopes
 
 
+# Prefixes of principal tags ('user:<id>', 'service:<uuid>'), which mark
+# nodes as owned by a single principal.
+PRINCIPAL_TAG_PREFIXES = ("user:", "service:")
+
+
 class AccessTags(frozenset[str]):
     def __new__(cls, tags=()):
         if isinstance(tags, str):
--- a/tiled/catalog/orm.py
+++ b/tiled/catalog/orm.py
@@ -21,7 +21,7 @@ from sqlalchemy import (
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.orm import Mapped, mapped_column, relationship
-from sqlalchemy.schema import PrimaryKeyConstraint, UniqueConstraint, CheckConstraint
+from sqlalchemy.schema import CheckConstraint, PrimaryKeyConstraint, UniqueConstraint
 from sqlalchemy.sql import func
 
 from ..server.schemas import Management
--- a/tiled/catalog/orm.py
+++ b/tiled/catalog/orm.py
@@ -201,6 +201,44 @@ class NodeAccessBlob(Base):
     )
 
 
+@event.listens_for(NodeAccessBlob.__table__, "after_create")
+def create_node_access_blob_cleanup_trigger(target, connection, **kw):
+    if connection.engine.dialect.name == "sqlite":
+        connection.execute(
+            text(
+                """
+CREATE TRIGGER node_access_blobs_delete_cleanup
+AFTER DELETE ON node_access_blobs
+BEGIN
+    DELETE FROM access_blobs WHERE id = OLD.access_blob_id;
+END"""
+            )
+        )
+    elif connection.engine.dialect.name == "postgresql":
+        connection.execute(
+            text(
+                """
+CREATE OR REPLACE FUNCTION delete_node_access_blob()
+RETURNS TRIGGER AS $$
+BEGIN
+    DELETE FROM access_blobs WHERE id = OLD.access_blob_id;
+    RETURN OLD;
+END;
+$$ LANGUAGE plpgsql;
+"""
+            )
+        )
+        connection.execute(
+            text(
+                """
+CREATE TRIGGER node_access_blobs_delete_cleanup
+AFTER DELETE ON node_access_blobs
+FOR EACH ROW EXECUTE FUNCTION delete_node_access_blob();
+"""
+            )
+        )
+
+
 class DataSourceAssetAssociation(Base):
     """
     This describes which Assets are used by which DataSources, and how.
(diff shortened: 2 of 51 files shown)


tiled/stream_messages.py
```diff
@@ -65,7 +65,7 @@ class ChildCreated(Update):
     specs: list[Spec]
     metadata: dict
     data_sources: list[DataSource]
-    access_blob: dict
+    access_tags: list[str]
```
danielballan, line 68: When a new server sends a message to a new client, it should accept `access_blob` and rewrite it into `access_tags`. This backward-compatibility affordance does not need to be maintained indefinitely but I think it could save real user pain to support it for a number of months.

tiled/queries.py
```diff
@@ -512,46 +512,40 @@ def SpecQuery(spec):
     return SpecsQuery([spec])
 
 
-@register(name="access_blob_filter")
+@register(name="access_tags_filter")
 @dataclass
-class AccessBlobFilter:
+class AccessTagsFilter:
```
danielballan, line 517: In practice, I expect this is used virtually only on the server side. In principle it can be used by clients to filter search results, and that could raise back-compat issues, but I think that use case is currently obscure enough that we can ignore it.

tiled/config.py
```diff
@@ -65,7 +65,7 @@ class CatalogConfig(BaseSettings):
     readable_storage: Optional[list[str]] = None
     init_if_not_exists: bool = False
     adapters_by_mimetype: Optional[dict[str, EntryPointString]] = None
-    top_level_access_blob: Optional[dict] = None
+    top_level_access_tags: Optional[list[str]] = None
```
danielballan, line 68: It's worth considering whether it's worth adding back-compat here, for a release or two.

tiled/catalog/orm.py
```diff
@@ -137,6 +146,242 @@ class NodesClosure(Base):
...
+        Index(
+            "idx_access_tags_is_public",
+            "name",
+            postgresql_where=text("is_public"),
+            sqlite_where=text("is_public"),
+        ),
+    )
+
+
+class NodeAccessTag(Base):
```
genematx: maybe call it `NodeAccessTagAssociation` (to distinguish from `AccessTag`)?

tiled/catalog/orm.py
```diff
@@ -137,6 +146,242 @@ class NodesClosure(Base):
...
+    """
+    Association table mapping Nodes to Access Tags (many-to-many).
+    Used to perform lookups in both directions, i.e.:
+        - Which tags are on this node? (served by the primary key)
+        - Which nodes have this tag? (served by the secondary index)
+    """
+
+    __tablename__ = "node_access_tags"
```
genematx: same, `*_association` to match the existing association tables

tiled/catalog/orm.py
```diff
@@ -137,6 +146,242 @@ class NodesClosure(Base):
...
+        passive_deletes=True,
+    )
+    owned_tags: Mapped[List["AccessTagOwner"]] = relationship(
+        back_populates="principal",
+        cascade="all, delete-orphan",
+        passive_deletes=True,
+    )
+
+
+class Scope(Base):
```
danielballan: This table can be dropped. The scopes enum can be used directly.

tiled/queries.py
nmaytan: Unlike client back compat which is essentially a facade, this is a lot closer to the access control logic. I think that's a good reason to force it forward instead of maintaining version/type-based branches in the logic.

tiled/config.py
nmaytan: I'm less convinced of this. IF you update your server version, I think it's reasonable to update your server configuration. My current thinking, anyway.

tiled/catalog/orm.py
nmaytan: As above, we will make this name more clear

tiled/stream_messages.py
nmaytan: Client-side back compatibility is probably worth saving. I was hoping rambunctiously to be fully rid of the blob, but avoiding all client updates my make sense.

tiled/catalog/orm.py
nmaytan: Will convert this

nmaytan: Overall seems that we lean enum, but a concern to consider:
- What if we need to rename a scope? -> have to migrate potentially millions of grants rows

danielballan: In the table, an enum is represented as an integer. The label is stored separately as a type and can be inexpensively changed. https://www.postgresql.org/docs/current/sql-altertype.html

tiled/access_control/protocols.py
```diff
@@ -1,9 +1,24 @@
...
 from ..server.schemas import Principal
-from ..type_aliases import AccessBlob, AccessTags, Filters, Scopes
+from ..type_aliases import Filters, Scopes
+
+# Prefixes of principal tags ('user:<id>', 'service:<uuid>'), which mark
+# nodes as owned by a single principal.
+PRINCIPAL_TAG_PREFIXES = ("user:", "service:")
+
+
+class AccessTags(frozenset[str]):
```
danielballan: I've given this more attention than it deserves. I guess using a class where a built-in will do is a pet peeve for me.

Two alternatives to consider.

1. Use a function to perform the validation, and return a builtin type.

```python
def access_tags(tags: Iterable[str] = ()) -> frozenset:
    if isinstance(tags, str):
        raise TypeError(
            "AccessTags expects an iterable of strings, not a single string."
        )
    return frozenset(tags)
```

2. Lean in to this being a class and get it a distinct `__repr__`, e.g., `f"{type(self).__name__}({', '.join(self)})`. Custom classes should also have `__getstate__` and `__setstate__` so that are pickleable in case they ever interact with multi-process code in the future. (I can think of some planned features where this could arise.)

danielballan:
For Future Us or for anyone following along, this PR addresses an acute performance issue that we have been exploring since Spring 2026. Here I will summarize the problem that this PR solves.

---

Queries to `count` or list the nodes table timing out (i.e., more than 30 seconds) for some users. Investigation with `pg_stat_statements` on the prod PG instance showed the offending query shape was among the most expensive on the table, as measured by cumulative time:

```sql
SELECT count(*) FROM nodes
WHERE nodes.parent = $1
AND ((nodes.access_blob[$2] ?| $3::TEXT[]) OR (nodes.access_blob ->> $4) = $5)
```

This query shape accounted for ~919K calls at ~50ms mean (~46 billion ms / ~12,900 hours cumulative), with the metadata-filtered variants adding thousands more calls at 2-5s. Users with access to many proposals (beamline staff, tiled admins) saw much worse performance, 30s or more..

The `access_blob['tags'] ?| ARRAY[...]` filter in-lines the entire set of a principal's accessible tags into every query as a literal array. For ordinary users this list is short, but for admins the list routinely reaches high hundreds or low thousands of tags. The `?|` operator performs roughly one GIN index probe per array element and unions the resulting bitmaps, so cost scales with the number of tags. At ~1000 tags that is ~1000 probes plus a ~1000-way bitmap union, per query.

No index can fix this. I ran some experiment on a copy of prod, restored from a backup onto staging.

A full-column GIN on `access_blob` was available but _never chosen by the planner_. The B-tree index on `parent` was always more selective, so queries scanned all of a parent's children (e.g. ~100k rows) and filtered them row-by-row against the `OR` condition. Building targeted expression indexes, GIN on `access_blob['tags']` and a B-tree on `access_blob ->> 'user'`, made single-tag lookups fast. A clean `BitmapOr` of the two indexes `BitmapAnd`-ed with `ix_nodes_parent` ~9–45ms vs 355ms.

A second confounding issue was that the JSONB keys 'tags'/'user' arrived as bind parameters (`$2`, `$4`) rather than literals, defeating the expression indexes under generic plans; `plan_cache_mode = force_custom_plan` restored the fast plan.

But at scale, these improvements were overwhelmed by a larger scaling problem. The `?|` operator issues one probe per element in the array. An index can make one probe fast, but it cannot collapse a thousand probes into one operation. The optimizations tried above were all necessary but insufficient because they optimize the per-probe cost, not the probe count.

Rather than carrying a principal's accessible tags as a literal array in-lined into every query, this PR introduces an access_tag / nodes association table. The filter becomes a join against that table: the planner hashes the principal's accessible set once and performs a single indexed join against nodes' tags and, unlike the opaque JSONB array, the association table gives the planner real statistics to plan against. This is the structural change that scales to thousands of tags where the index-only approach fundamentally cannot.

(Separately, unrelated to access_blob: the nodes table also carried a broken 51GB GIN index—`top_level_metadata`, on `(parent, time_created, id, metadata, access_blob)`—that mixed btree columns into a GIN and could not be used by any query. Its removal, and the metadata/pagination index work, are tracked separately.)

tiled/catalog/orm.py
nmaytan: Addressed in incoming commits.

tiled/catalog/orm.py
nmaytan: Incoming commit(s)

tiled/catalog/orm.py
nmaytan: Incoming commit(s)

tiled/access_control/protocols.py
nmaytan: Upon reflection, I've settled on (1) + a type alias. I originally thought a class would be more graceful than something like this, but I've changed my mind for 3 reasons:
* I appreciate your instinct here
* We use type aliases for other somewhat similar cases
* Set operations effectively strip the subclass type (`AccessTags `-> `frozenset`) which I _really_ dislike.

nmaytan: Incoming commit(s)

danielballan (approved):

danielballan:
We _could_ go to some additional trouble to help users of the Python client who use the `access_blob` keyword arguments or property. I suspect usage of this is limited enough---i.e., mainly people who are following this work closely and are aware. So I'm leaning that it is not worth the effort or cruft. We could always reverse that decision later if we receive report of trouble.

nmaytan:
The client already used `access_tags` for its interface, so that should continue to work the same.

I agree about the attribute, and my default position is to push folks to use the newer `access_tags` before the older attribute becomes more ingrained. As you say, we can revisit this if it is a major problem for someone.

Issue #1205, opened 2025-10-29
Evaluate RustFS as a MinIO successor


danielballan:
As MinIO [pulls back support for container images](https://eu.36kr.com/en/p/3521684347706501), the project [RustFS](https://github.com/rustfs/rustfs) looks interesting.

We might consider testing against that in addition to or instead of MinIO.

ZohebShaikh:
Out of interest in this, What does NSLS-2 use Minio for ?
Is it used for putting the large detector data into a S3 bucket which are create by detectors ?

danielballan:
We have NetApp appliance that provides an S3 interface. This is the only S3 solution we use in production. Currently, it is used to store processed data, analyzed data, or any large arrays that are uploaded "inline" with Bluesky Event documents. It is not _yet_ used as store for detectors, but we have run some tests in that direction and hope to be doing that kind of things soon.

As for MinIO specifically, it's only used for local development and testing. For production, we're 100% on the NetApp solution.

swelborn:
i dropped this in to another project following this blog https://rustfs.com/blog/binary-replacement-a-simple-way-to-migrate-from-minio-to-rustfs/

worked just fine for local dev

PR #1518, opened 2026-09-16
ci: change minio container registry

ZohebShaikh:
### Checklist
- [ ] Add a Changelog entry

As minio is removed from docker IO, added alternative mirrors so that the CI can pass [News](https://www.reddit.com/r/minio/comments/1wfs5hc/minio_just_removed_their_dockerhub_image/)

I'm using `quay.io/minio/mc:latest` to create the bucket. But the long term solution is moving away from it https://github.com/bluesky/tiled/issues/1205


ZohebShaikh:
Screenshot from 2026-09-16 15-17-19


We have a developer group in Github, which includes the maintainer of the project. Can we have something like this in bluesky org as well. It will be good to tag a group rather than individuals for a PR.

ZohebShaikh:
closed in favour of https://github.com/bluesky/tiled/pull/1522

PR #1527, opened 2026-09-21
BUG: include /api in the root_path reported by the About endpoint

swelborn:
the current behavior yields:

`“root_path”: request.scope.get(“root_path”) or (“” + “/api”)`

rather than what we want:

`“root_path”: (request.scope.get(“root_path”) or “”) + “/api”)`

### Checklist
- [ ] Add a Changelog entry


swelborn:
Folded into #1530, which fixes the other `root_path` bugs.

